Issue #141, opened 2026-08-24
[BENCH-1] Reproducible framework benchmark suite + README performance results


leo-aa88:
## Summary

Build a **reproducible, checked-in benchmark suite** for Gombit and publish a concise, defensible performance section in the root `README.md`.

Gombit is positioned as a **Django-for-Go full-stack framework**, not merely an HTTP router. The benchmark suite must therefore answer two different questions:

1. **What does Gombit's abstraction layer cost compared with the underlying Go stack?**
2. **How does a realistic Gombit application compare with conventional full-stack frameworks under equivalent workloads?**

Do **not** optimize for a vanity “hello world requests/sec” number.

The primary story we want to quantify is:

> How much performance do Gombit's batteries — typed handlers, validation, OpenAPI, GORM, authentication, migrations, generated client conventions, and single-binary deployment — cost relative to hand-assembled Go, and what application-level performance/footprint does that produce compared with established full-stack frameworks?

The benchmark must be reproducible from the repository, document all methodology and versions, avoid cherry-picked results, and make it difficult to accidentally publish misleading numbers.

---

## Repository / governance note

`docs/GOMBIT_BUILD_PLAN.md` is the authoritative backlog and `AGENTS.md` instructs agents not to invent work outside that backlog.

This issue is explicit authorization to add a new post-v0.1 backlog item:

```text
[BENCH-1] Reproducible framework benchmark suite + README performance results
```

Add the entry to the appropriate post-v0.1 / infrastructure section of `docs/GOMBIT_BUILD_PLAN.md` before considering this work complete.

Do not alter any locked architecture decision while implementing the benchmarks.

---

# Motivation

The existing README explains Gombit's value primarily in terms of integrated batteries:

* Gin + Huma typed HTTP runtime
* validation
* OpenAPI 3.1 generated from handlers
* generated TypeScript client
* GORM
* Atlas migrations
* JWT and cookie/session authentication
* CSRF
* Django-style admin
* React frontend
* `gombit build --embed` single-binary deployment

A raw router benchmark cannot measure that proposition.

Gombit already has an M0 Huma-over-Gin spike with a Go microbenchmark. Preserve that benchmark and use it as the seed for the new suite rather than creating an unrelated benchmark implementation.

The new benchmark suite should distinguish between:

### Framework overhead

Measure the incremental cost of:

```text
net/http
   ↓
Gin
   ↓
Huma over Gin
   ↓
Gombit
```

### Application performance

[...]

PR #176, opened 2026-08-25
feat: BENCH-1 canonical schema + seed + Gin+GORM control app (Phase 3a)

leo-aa88:
## Summary

Phase 3a of the BENCH-1 plan ([docs/plans/BENCH-1-benchmark-suite.md](docs/plans/BENCH-1-benchmark-suite.md)):
the canonical schema/API/envelope every `benchmarks/apps/` implementation
converges on, and the first implementation against it.

- [benchmarks/docs/schema.md](benchmarks/docs/schema.md): the `users`/`projects`
  DDL (issue #141 §3), the five canonical routes, and the decision to use the
  D10 envelope on **both** Go apps in this comparison — deliberately
  different from `benchmarks/micro/scenario`'s per-stack-idiomatic choice,
  since this comparison holds the envelope constant to isolate framework
  cost. Verified column-by-column against a real running Postgres instance
  (`psql \d`), not assumed from the model definitions — caught and fixed a
  real gap this way (`Project.Name` and the timestamp columns were nullable;
  the schema requires `NOT NULL`).
- `benchmarks/apps/shared`: `PageMeta` (`page`/`limit`/`total` — not
  `contract.PageMeta`, which uses `per_page`; issue #141 specifies `limit`)
  and `ProjectData`, shared by every Go implementation so "same JSON shape"
  (issue #141 §15) is true by construction, not by two hand-maintained
  struct definitions staying in sync.
- `benchmarks/apps/gin-gorm`: the primary framework-tax control (issue
  "isolates the incremental cost of adopting Gombit rather than changing
  programming languages"). GORM `AutoMigrate`; `.Preload("Owner")` on the
  list query, verified against real Postgres statement logs to issue exactly
  3 SQL statements per list request (count + page + one batched owner
  `IN (...)`, not one owner query per row) before writing that as an
  automated regression guard gated behind `-tags integration` (matching
  `database/integration_test.go`'s existing convention — `go test ./...`
  needs no DSN and skips it). 20/20 connection pool limits per issue's
  "Connection pooling" requirement, overriding Gombit's own 25/5 Postgres
  default — the fairness pin is the issue's, not either framework's.
- `benchmarks/compose.yml`: pinned `postgres:16.4-alpine`, resource-limit
  block documented as best-effort outside Swarm per the issue's own
  "detect/report rather than silently pretend" requirement.

Verified end-to-end against a real dockerized Postgres instance (not just
unit-level assertions): migrate, seed (1,000 users / 100,000 projects), full
CRUD round trip, pagination/ordering across a page boundary, and the N+1
guard.

Part of #141. Not closing it — Phase 3b (the `gombit` app implementing the
same API, and the cross-implementation fairness checks that need a second
implementation to compare against) is a follow-up, along with everything
else in the plan's later phases.

## Acceptance criteria

Phase 3 as originally scoped in the plan bundles both apps into one AC; this
PR is the 3a half:

- [x] Canonical schema documented and verified against real PostgreSQL
- [x] Deterministic, reproducible seed (1,000 users / 100,000 projects)
- [x] `benchmarks/apps/gin-gorm` exposes the canonical CRUD routes with the
      D10 envelope and equivalent pagination
- [x] N+1 avoided on the list endpoint, verified against real Postgres
      statement logs and covered by an automated regression test
- [ ] `benchmarks/apps/gombit` — deferred to Phase 3b
- [ ] Cross-implementation fairness checks — deferred to Phase 3b (need a
      second implementation to compare against)
- [ ] `make benchmark-crud` / `results.json` output — deferred; also depends
      on Phase 1's still-open result-schema/summarizer work

## Scope notes

Deferred to Phase 3b and later phases (see plan doc):

- `benchmarks/apps/gombit` (Huma handlers, `framework.App`, Atlas migrations)
- Cross-implementation fairness checks (route surface, JSON shape, query
  count, ordered IDs)
- `benchmarks/compose.yml` app services (currently Postgres only)
- Django/Rails/Laravel/NestJS apps (Phase 4)
- Load harness, auth/TechEmpower/concurrency/footprint benchmarks, CI,
  README `## Performance` section (Phases 5–8)

No locked architecture decision is touched.

## Validation

- [x] `go build ./...`
- [x] `go test ./...` — green except the same two pre-existing, unrelated
      `error obtaining VCS status` failures in `cmd/gombit`/`goldentest`
      confirmed on unmodified `main` in earlier PRs against this issue
- [x] `go vet` / `go vet -tags integration` — clean
- [x] `gofmt -l` — clean
- [x] `golangci-lint run ./benchmarks/...` — 0 issues
- [x] `go test -tags integration ./benchmarks/apps/gin-gorm/...` against a
      real `docker compose -f benchmarks/compose.yml up -d postgres` —
      green (`TestCRUDRoundTrip`, `TestListPaginationAndOrdering`,
      `TestListDoesNotN1`)
- [x] Manual end-to-end verification against the running app: migrate,
      seed, every CRUD route, 404, 422 validation failure, and SQL
      statement-log inspection confirming exactly 3 queries per list request
- [ ] Project `code-review` skill run against this diff — not yet run

- [x] New behavior has tests; DB-touching change is covered by an
      integration-tagged suite run against a real PostgreSQL instance
      (SQLite/MySQL don't apply here — this benchmark specifically requires
      PostgreSQL parity per issue #141, not the v0.1 DB matrix)
- [ ] Stable features ship docs and appear in an example app — N/A, this is
      benchmark tooling under `benchmarks/`, not a user-facing framework
      feature; it has its own README instead
- [x] Extraction preserves contracts — N/A, no extraction; reuses
      `contract.Data`/`DataMeta`/error-mapping helpers as-is
- [ ] Generators are idempotent/additive, AST-safe — N/A, no generator
      changes
- [ ] API changes regenerate OpenAPI/TS client — N/A, no public Gombit API
      changes (this is a separate benchmark app, not part of the framework)
- [x] No secrets in generated frontend source — N/A, no frontend changes
- [x] Scope stays inside milestone — post-v0.1, no M6 battery pulled in
- [x] This PR links its issue and states which AC it satisfies


benchmarks/docs/schema.md
```diff
@@ -0,0 +1,116 @@
...
+  a different call than `benchmarks/micro/scenario`'s framework-tax matrix,
+  which deliberately let each stack use its own idiomatic default —
+  intentional: that matrix isolates per-layer cost; this one isolates
+  adopting-Gombit cost against a fixed contract.
+- The list endpoint (`GET /api/projects`) returns 20 projects per page by
+  default, ordered by `id DESC` (deterministic — issue §"Canonical list
+  query" requires "deterministic ordering" and "the same ordered IDs for a
+  known query"), with the owner relationship preloaded in the same query set
+  (no N+1: one query for the page of projects, one query for their distinct
+  owners, not one owner query per project).
```
leo-aa88: **BLOCKING — query-count contract is false.**

This doc is the reference other implementations must match, and it specifies list = **2** SQL statements (page + distinct owners).

`handlers.list` does `Count` + `Find` + `Preload("Owner")`, and `TestListDoesNotN1` asserts **exactly 3**. The plan text already says 3.

```text
THE DOC SAYS:  2 queries
THE RUNTIME:   3 queries (count + page + batched owners)
```

Phase 3b/4 will copy this paragraph and either fail fairness or diverge while believing they complied. Fix the contract to pin `COUNT` + page + owner `IN (...)` for a non-empty page (or explicitly allow any fixed O(1) eager-load strategy and drop `== 3`).

leo-aa88 (commented):
# REQUEST CHANGES

> Note: GitHub refuses `REQUEST_CHANGES` on the PR author's own pull request, so this is submitted as a `COMMENT` review. The merge recommendation is still **do not merge** until the blockers below are fixed.

Phase 3a is real work — schema verified against live Postgres, seed is deterministic, list N+1 guard exists — but the **authoritative list-query contract contradicts the implementation**, and the tests that prove the rest never run in CI. That is not merge-ready for a fairness baseline other implementations will copy.

## 1. Canonical schema documents 2 list queries; runtime and tests pin 3

**BLOCKING.**

`benchmarks/docs/schema.md` is explicitly "the reference every implementation's migration is checked against." It says the list endpoint is:

> one query for the page of projects, one query for their distinct owners

The handler immediately does `Count`, then `Find`+`Preload`. `TestListDoesNotN1` asserts **exactly 3**. The plan doc already says 3. The schema doc says 2.

That is not a wording nit. Phase 3b/4 implementers will read schema.md, ship `COUNT`+`JOIN` (2) or a single windowed select (1), and then either fail a fairness pin written against gin-gorm's Preload shape or silently diverge while believing they matched the contract.

Fix: make schema.md state the pinned strategy explicitly — `COUNT` + page `SELECT` + batched owner `IN (...)` = 3 for a non-empty page — and keep handlers comments aligned. If you intend to allow any O(1) eager-load strategy, say that and stop asserting `== 3`.

## 2. CI is green without executing any test in this package

**MAJOR.**

Every test file is `//go:build integration`. CI runs integration tags for `database`/`auth`/`admin`/`migrations` against Postgres. It does **not** run `./benchmarks/apps/gin-gorm/...`.

`go test ./...` on this package is `[no test files]`. The PR's entire correctness claim is manual. Green CI here is not evidence. Wire the existing postgres job (or a thin compose-based job) to run these tests, or this baseline will rot the first time someone "refactors" handlers.

## 3. Seed contract has no automated coverage

**MAJOR.**

Phase 3a's seed is half the deliverable: 1,000/100,000, `user-%04d@example.com`, round-robin owners, truncate+`RESTART IDENTITY` idempotency. Nothing in CI or unit tests asserts any of that. `seedFixture` in tests is a different, smaller path.

A typo in the email format or owner formula becomes a silent fairness landmine in 3b. Extract the pure formatting/owner math (or run a tiny DB assertion on counts/IDs after a reduced seed) and test it without requiring the full 100k insert in CI.

## 4. Invalid `owner_id` is a 500, not a client error

**MAJOR.**

`POST /api/projects` with a non-existent `owner_id` hits the FK and returns `{"error":{"code":"internal","message":"create project"}}` at 500. Issue #141 §15 lists "every implementation rejects equivalent invalid input" as a fairness example. A bad client FK is not an internal failure.

Map FK violations to 4xx (or validate owner existence and return 422/404) and add a test. Leaving this as 500 guarantees cross-stack divergence the moment fairness checks look at error paths.

## 5. "Without Huma" control still imports framework packages that pull Huma

**MAJOR.**

The control is sold as idiomatic Gin+GORM without Huma/`framework.App`. It imports `contract` and `database`. `contract` imports Huma (`ErrorEnvelope` is a `huma.StatusError`). So the dependency graph of the primary "no Huma" baseline includes Huma.

Envelope shape sharing is required by the issue — that belongs in `benchmarks/apps/shared` (or a tiny D10 stub), not via linking the framework's contract/database stack. `database.MapPersistError` is especially unjustified here: projects have no unique business key; the conflict branch is dead weight, and FK failures become "internal" through that helper (see #4).

# VERDICT

The deepest problem is **specification**: the document that future apps must converge on describes a different list-query shape than the control implements and tests. Everything else (CI theater, untested seed, FK→500, Huma in the "no Huma" binary) is the same class of defect — claims that will be treated as true by the next engineer and are not.

Do not merge until the query-count contract is corrected and the other MAJOR items are addressed or explicitly deferred with a tracked follow-up that Phase 3b cannot ignore.

benchmarks/apps/gin-gorm/handlers.go
```diff
@@ -0,0 +1,243 @@
...
+
+type updateProjectRequest struct {
+	Name        *string `json:"name" binding:"omitempty,max=255"`
+	Description *string `json:"description"`
+}
+
+// list handles GET /api/projects?page=&limit=: a single page of projects,
+// deterministically ordered (id DESC), owner preloaded in the same query
+// set (one query for the page, one for its distinct owners: GORM's
+// .Preload, not one owner query per row) so this endpoint never N+1s.
```
leo-aa88: **Same lie, local copy.**

Comment claims "one query for the page, one for its distinct owners" — then the body opens with `Count` (line 60). That is three statements on the non-empty path. Comments that disagree with the next ten lines are bugs; they will be pasted into the next implementation.

benchmarks/apps/gin-gorm/main_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+// TestListDoesNotN1s counts the SQL statements the list endpoint issues via
+// a counting gorm.Logger and asserts a small constant bound, independent of
+// page size — issue #141 §16 requires "the same fixed number of SQL
+// queries" and specifically forbids an implementation whose query count
+// scales with the number of returned rows. Verified manually against
+// Postgres statement logs (docs/plans/BENCH-1-benchmark-suite.md, Phase 3
+// notes) before writing this as an automated regression guard: one count
+// query, one page query, one batched owner IN (...) query — 3 total,
+// regardless of whether the page holds 1 row or 20.
```
leo-aa88: **MAJOR — test comment overclaims; CI never runs this.**

1. Comment says "exactly 3 … regardless of whether the page holds 1 row or 20." This test only hits `limit=20` with rows present. An empty page (`page=99`) issues **2** queries (count + page; preload skipped). Pin the invariant you actually mean: non-empty benchmark page ⇒ 3, and count must not scale with N.
2. The whole file is `//go:build integration`, and CI does not invoke this package. Green PR checks do not execute this guard. Wire it into the postgres integration job or admit the N+1 claim is manual-only.

benchmarks/apps/gin-gorm/seed.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// seedDatabase truncates and repopulates the canonical benchmark dataset:
+// deterministic content so every implementation seeds byte-identical rows,
+// never inside a timed benchmark run. Truncating first (RESTART IDENTITY)
+// makes repeated `-seed` invocations idempotent instead of accumulating
+// duplicate data, and resets the users sequence to 1 so seedProjects'
+// ownerID computation (round-robin over 1..seedUserCount) stays correct
+// without reading back generated IDs.
+func seedDatabase(ctx context.Context, db *gorm.DB) error {
```
leo-aa88, line 20: **MAJOR — claimed seed contract is untested.**

This function is half of Phase 3a (deterministic 1k/100k, `user-%04d@…`, round-robin owners, truncate+`RESTART IDENTITY`). Nothing asserts those invariants. Integration tests use `seedFixture`, a different code path.

Add coverage for the pure formatting/owner math and/or a reduced-N DB check. Otherwise 3b fairness on "byte-identical row N" is wishful.

benchmarks/apps/gin-gorm/handlers.go
```diff
@@ -0,0 +1,243 @@
...
+		return
+	}
+
+	row := Project{
+		OwnerID:     body.OwnerID,
+		Name:        body.Name,
+		Description: body.Description,
+	}
+	if err := h.DB.WithContext(c.Request.Context()).Create(&row).Error; err != nil {
+		writeMapped(c, database.MapPersistError(c.Request.Context(), err, "project already exists", "create project"))
```
leo-aa88: **MAJOR — invalid FK is advertised as an internal error.**

`MapPersistError` only special-cases unique violations. A missing `owner_id` hits `fk_projects_owner` and becomes HTTP **500** `code=internal` `message="create project"` (verified against the live compose Postgres).

Issue #141 §15 expects equivalent **rejection of invalid input**. A bad client reference is not a 500. Map FK violations to 4xx (or pre-check the owner) and add a test. Also: this is why importing `database.MapPersistError` into the "plain Gin" control is not free glue — it encodes Gombit's conflict ontology onto a table with no unique business key.

benchmarks/apps/gin-gorm/handlers.go
```diff
@@ -0,0 +1,243 @@
+package main
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+	"github.com/gombit-dev/gombit/benchmarks/apps/shared"
+	"github.com/gombit-dev/gombit/contract"
+	"github.com/gombit-dev/gombit/database"
```
leo-aa88: **MAJOR — "without Huma" is false in the dependency graph.**

`contract` pulls `huma/v2` (`ErrorEnvelope` is a `huma.StatusError`). The primary control sold as Gin+GORM without Huma still links Huma, plus `database` framework helpers.

Sharing the D10 **JSON shape** is required; linking the framework packages is not. Put envelope types in `benchmarks/apps/shared` (or a tiny stub) so the control's module graph matches the story you will publish numbers against.

benchmarks/apps/gin-gorm/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```sh
+docker compose -f benchmarks/compose.yml up -d postgres
+go test -tags integration ./benchmarks/apps/gin-gorm/... \
+  -database.dsn "postgres://gombit:gombit@127.0.0.1:55432/gombit_bench?sslmode=disable"
+```
+
+Gated behind the `integration` build tag (matching `database/integration_test.go`'s
+convention) since it needs a live PostgreSQL instance — `go test ./...`
+without the tag skips this package's tests entirely, no DSN required.
```
leo-aa88: **MAJOR — documents that default `go test ./...` skips everything, and CI agrees.**

Matching `database/integration_test.go`'s tag is fine only if CI also runs the tag. Today the Database (postgres) job runs `./database`, `./auth`, `./admin` — not this package. Until that line exists, this README is an admission that merge CI cannot defend the control app.

benchmarks/apps/gin-gorm/handlers.go
```diff
@@ -0,0 +1,295 @@
...
+	c.JSON(http.StatusOK, shared.Data[shared.ProjectData]{Data: toProjectData(row)})
+}
+
+// create handles POST /api/projects.
+func (h *Handler) create(c *gin.Context) {
+	var body createProjectRequest
+	if err := c.ShouldBindJSON(&body); err != nil {
+		writeError(c, shared.ValidationError(err.Error(), nil))
+		return
+	}
```
leo-aa88, line 124: **MAJOR — create does not share PATCH's non-blank name rule.**

After `ShouldBindJSON`, create proceeds with whatever `required` accepted. `binding:"required"` rejects `""` but accepts `"   "`.

Verified at HEAD against live Postgres:

```text
POST {"owner_id":1,"name":"   "} → 201, name stored as three spaces
```

PATCH (below) explicitly TrimSpace-rejects blank names and the comment says that matches POST's rule / "at least one non-whitespace character". POST does not enforce that.

Apply the same check (or a shared helper) on create and add a test. Right now the resource has two name contracts.

benchmarks/apps/gin-gorm/handlers.go
```diff
@@ -0,0 +1,295 @@
...
+	// binding's `omitempty` on a *string skips `max=255` when the pointed-to
+	// value is empty, not just when the pointer is nil — a client that sends
+	// {"name":""} (as opposed to omitting the field) would otherwise bypass
+	// the same non-blank-name rule POST /api/projects enforces via
+	// `required`. Checked explicitly since binding tags can't express "at
+	// least one non-whitespace character, only when provided" on a pointer.
+	if body.Name != nil && strings.TrimSpace(*body.Name) == "" {
```
leo-aa88: This check is correct for PATCH. The bug is the missing twin on create.

A client can invent a whitespace-only name via POST, then never clear it via PATCH — absurd asymmetry for a fairness baseline that later apps will copy.

leo-aa88 (commented):
# REQUEST CHANGES

> Note: GitHub refuses `REQUEST_CHANGES` on the PR author's own pull request, so this is submitted as a `COMMENT` review. Merge recommendation: **do not merge** until the remaining defect below is fixed.

Re-review at HEAD `2cfa142`.

Prior review blockers are fixed on this tip:

- `schema.md` pins 3 queries (non-empty) / 2 (empty) and documents O(1) alternatives
- CI `database-postgres` runs `./benchmarks/apps/gin-gorm/...` against `gombit_bench`
- Seed formulas unit-tested; `seedDatabaseN` idempotency covered under integration
- FK violations map to 422 `validation_error` without importing `database`
- D10 envelopes live in `benchmarks/apps/shared` — control dependency graph has no Huma/`contract`/`database`/`framework`
- PATCH blank-name check, queryCounter mutex, `Session` logger, timestamp assertion hardened

Unit + integration suites are green at this HEAD. One defect remains from the incomplete blank-name fix.

## 1. POST still accepts whitespace-only `name`; PATCH rejects it

**MAJOR.**

`update` rejects `strings.TrimSpace(*body.Name) == ""` and claims that matches POST's non-blank rule / "at least one non-whitespace character". `create` still relies only on `binding:"required,max=255"`.

Verified against live compose Postgres at this commit:

```text
POST {"owner_id":1,"name":"   "}  → 201 Created  (name stored as three spaces)
POST {"owner_id":1,"name":""}     → 422 validation_error
```

`required` rejects empty string, not whitespace. Half an invariant is not an invariant. A client can create a whitespace-only name via POST and then never clear it via PATCH.

Apply the same TrimSpace rejection on create (shared helper preferred), and add a create-side test. Until then the control teaches two name contracts on one resource.

# VERDICT

Previous specification/architecture mismatches are gone. Remaining issue is **implementation**: the blank-name contract introduced in this patch series is only enforced on PATCH. Fix that asymmetry, then Phase 3a is mergeable.

PR #213, opened 2026-08-28
fix: admin create no longer masks type-coercion errors as "is required"

### LINKED ISSUE #198: [bug] admin create reports a provided-but-invalid required field as "is required"
## What happened

When creating an admin resource (`POST /api/v1/admin/resources/{slug}`), if a **required** field is *present in the body but fails type coercion* (e.g. a required `integer` field sent as `"not-a-number"`), the D10 validation error for that field is `"is required"` instead of the real coercion error (`"must be an integer"`).

The client submitted the field, so `"is required"` is misleading — it tells the user the field is missing when in fact its value is the wrong type.

### Root cause

`admin/resources.go` → `applyWrite`. On the create path the second loop overwrites any error already recorded for a required field:

```go
if err := f.set(inst, raw); err != nil {
    fields[name] = []string{err.Error()}   // e.g. "must be an integer"
    continue                                // seen[name] is NOT set
}
...
if creating {
    for i := range m.fields {
        f := &m.fields[i]
        if !f.Required || f.ReadOnly { continue }
        if _, ok := seen[f.Name]; !ok {
            fields[f.Name] = []string{"is required"} // overwrites the coercion error
        }
    }
}
```

Because a failed `f.set` leaves the field out of `seen`, the `creating` loop unconditionally replaces the specific message with `"is required"`. The fix is to only set `"is required"` when the field has no error recorded yet (i.e. it was truly absent from the body): `if _, ok := fields[f.Name]; ok { continue }`.

## Steps to reproduce

```shell
# Register an admin model with a required integer field, then:
curl -X POST /api/v1/admin/resources/<slug> \
  -H 'Content-Type: application/json' \
  --cookie 'gombit_access=...; gombit_csrf=...' \
  -H 'X-CSRF-Token: ...' \
  -d '{"count": "not-a-number"}'
# error.fields.count == ["is required"]  (expected: ["must be an integer"])
```

Confirmed with a focused unit test against `applyWrite` (creating=true, required integer field, value `"not-a-number"`): `fields["count"]` comes back as `["is required"]`.

## gombit version

dev (built from `main` / `bench-1-snapshot`)

## Environment

dev checkout, module `github.com/gombit-dev/gombit`

## Database

- not database related

## Auth mode

- cookie (session + CSRF)

## Area

- admin

## Logs or output

```
applyWrite(creating=true, {"count":"not-a-number"}) -> error.fields.count = ["is required"]
expected                                            -> error.fields.count = ["must be an integer"]
```

## Before submitting
- [x] This reproduces on `main`

Brunotlps:
## Summary

- `admin/resources.go`'s `applyWrite` unconditionally overwrote a required
  field's real coercion error with the generic `"is required"` message on
  create, whenever that field wasn't in `seen` — including when it failed
  type coercion rather than being absent.
- Fix: skip the "is required" pass when the field already carries an error
  in `fields`, not only when it was successfully set (`seen`).

Closes #198

## Acceptance criteria

(the issue is a bug report without a formal AC checklist; derived from its
"What happened" / "Root cause" sections)

- [x] A required field present in the body but failing type coercion
      reports its real coercion error (e.g. `"must be an integer"`),
      not `"is required"`.
- [x] A required field genuinely absent from the body still reports
      `"is required"` (regression-checked in combination with the above,
      in the same request).

## Scope notes

Nothing deferred — this is a single logic fix contained to `applyWrite`.

## Validation

- [x] `go build ./...`
- [x] `go test ./...`
- [x] Project `code-review` skill run against this diff (APPROVE)
- [x] Confirmed both new tests fail against the pre-fix code (reproducing
      the exact bug) and pass with the fix

- [x] New behavior has tests. DB-touching matrix N/A: `applyWrite` is pure
      in-memory validation that runs before any persistence call, with no
      driver-specific behavior.
- [ ] Docs + example — N/A, bugfix to existing validated behavior, not a
      new stable feature.
- [ ] Extraction preserves contracts — N/A, not an extraction.
- [ ] Generators — N/A, no generator touched.
- [ ] API changes regenerate OpenAPI + TS client — N/A, the D10
      `error.fields` shape is unchanged; only the runtime message content
      for one already-dynamic field is corrected.
- [ ] No secrets in generated frontend — N/A, no frontend touched.
- [x] Scope stays inside this issue's milestone — single logic fix, no M6
      battery creep.
- [x] This PR links its issue and states which acceptance criteria it
      satisfies.

### DIFF
--- a/admin/resources.go
+++ b/admin/resources.go
@@ -286,9 +286,13 @@ func applyWrite(ctx context.Context, m *registered, inst any, body map[string]an
 			if !f.Required || f.ReadOnly {
 				continue
 			}
-			if _, ok := seen[f.Name]; !ok {
-				fields[f.Name] = []string{"is required"}
+			if _, ok := seen[f.Name]; ok {
+				continue
 			}
+			if _, ok := fields[f.Name]; ok {
+				continue
+			}
+			fields[f.Name] = []string{"is required"}
 		}
 	}
 	if len(fields) > 0 {
--- a/admin/resources_test.go
+++ b/admin/resources_test.go
@@ -422,6 +422,56 @@ func TestResourceValidation(t *testing.T) {
 	assertError(t, unknown, http.StatusUnprocessableEntity, contract.CodeValidationError)
 }
 
+func TestResourceCreateInvalidRequiredFieldReportsCoercionError(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	app := newCookieApp(t)
+	registerWidgets(t, app, func(o *admin.Options) {
+		for i := range o.Fields {
+			if o.Fields[i].Name == "price" {
+				o.Fields[i].Required = true
+			}
+		}
+	})
+	jar := loginSuperuser(t, app)
+
+	rec := doRequest(app, jar, http.MethodPost, "/api/v1/admin/resources/widgets",
+		`{"name":"X","price":"not-a-number"}`)
+	assertError(t, rec, http.StatusUnprocessableEntity, contract.CodeValidationError)
+
+	env := decodeError(t, rec)
+	got := env.Fields["price"]
+	if len(got) != 1 || got[0] != "must be an integer" {
+		t.Fatalf("fields.price = %#v, want [\"must be an integer\"]", got)
+	}
+}
+
+func TestResourceCreateMixedMissingAndInvalidRequiredFields(t *testing.T) {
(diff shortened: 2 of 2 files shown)


leo-aa88:
Thanks for your contribution. I'll review it.

leo-aa88 (approved):
LGTM

PR #216, opened 2026-08-28
fix: admin create/update/delete no longer report FK/NOT NULL violations as 500

### LINKED ISSUE #202: [bug] admin data plane returns 500 for client-caused constraint violations (FK / NOT NULL)
## What happened

Creating or updating an admin resource with a bad `belongs_to` foreign key (a parent row that does not exist), or omitting a column the DB enforces `NOT NULL` on, makes the data-plane handler return **500 Internal Server Error** with `error.code = "internal"`. These are client-caused input errors and should be a 4xx (validation/conflict), not a server error.

A 500 both misleads the API client and trips production error alerting for what is ordinary bad input.

### Root cause

`admin/resources.go` create/update/delete call `database.MapPersistError(...)`, which only classifies **unique/duplicate** as a client error:

```go
func MapPersistError(ctx context.Context, err error, conflict, internal string) error {
    if err == nil { return nil }
    if IsUniqueViolation(err) {   // matches only "unique" / "duplicate" / ErrDuplicatedKey
        return contract.WithContext(ctx, contract.Conflict(conflict))
    }
    return contract.WithContext(ctx, contract.Internal(internal))  // FK, NOT NULL, CHECK -> 500
}
```

`IsUniqueViolation` looks for the substrings `unique`/`duplicate`; a foreign-key or not-null violation contains neither, so every such error falls through to `contract.Internal` (500).

## Steps to reproduce

Register a model with a `belongs_to` relation (or a `NOT NULL` column not marked required in the admin `Fields`), then:

```shell
curl -X POST /api/v1/admin/resources/<slug> \
  -H 'Content-Type: application/json' \
  --cookie 'gombit_access=...; gombit_csrf=...' -H 'X-CSRF-Token: ...' \
  -d '{"parent_id": 999999}'   # no parent row 999999
# -> HTTP 500 {"error":{"code":"internal", ...}}   (expected: a 4xx)
```

## Confirmed

A focused unit test against `MapPersistError` with the real driver error strings maps all of them to 500:

| driver error string | status |
| --- | --- |
| `FOREIGN KEY constraint failed` (sqlite) | 500 |
| `ERROR: insert or update on table violates foreign key constraint` (postgres) | 500 |
| `Cannot add or update a child row: a foreign key constraint fails` (mysql) | 500 |
| `NOT NULL constraint failed: widgets.name` (sqlite) | 500 |

## gombit version

dev (built from `main` / `bench-1-snapshot`)

## Environment

dev checkout, module `github.com/gombit-dev/gombit`

## Database

- sqlite
- postgres
- mysql

## Auth mode

- cookie (session + CSRF)

## Area

- admin
- database / GORM

## Logs or output

```
MapPersistError("FOREIGN KEY constraint failed") -> status=500 code=internal
```

Fix direction: add cross-driver foreign-key / not-null detection (mirroring `IsUniqueViolation`, ideally via `gorm.ErrForeignKeyViolated` where available) and map those to a 4xx category. Suggested: FK violation → `conflict`/`validation`, not `internal`.

## Before submitting
- [x] This reproduces on `main`

Brunotlps:
## Summary

- `database.MapPersistError` only classified unique-key violations; a
  foreign-key violation (bad `belongs_to` reference) or a NOT NULL
  violation fell through to a generic internal/500 error.
- Enabled `gorm.Config.TranslateError` so unique/foreign-key detection is
  sentinel-based (portable across SQLite/Postgres/MySQL error codes)
  instead of driver-message-only. NOT NULL has no `gorm` sentinel on any
  of the three dialectors, so it stays message-based by necessity.
- Added `IsForeignKeyViolation` / `IsNotNullViolation`, mirroring
  `IsUniqueViolation`'s shape. `MapPersistError` now maps both to a D10
  validation error on create/update.
- Added `MapDeleteError`, used only by the admin delete path: a
  foreign-key violation there means another row still references the one
  being deleted — the opposite meaning from create/update — so it maps to
  conflict, not validation. Found and fixed during this PR's own
  `code-review` pass (first pass was REQUEST CHANGES over exactly this).

Closes #202

## Acceptance criteria

(the issue is a bug report without a formal AC checklist; derived from its
"Fix direction" section, plus one criterion this PR added during review)

- [x] A foreign-key violation on create/update reports a 4xx validation
      error, not 500.
- [x] A NOT NULL violation on create/update reports a 4xx validation
      error, not 500.
- [x] (found during review, not in the original report) A foreign-key
      violation on **delete** reports 409 conflict — not the create/update
      validation message, which would be factually wrong there, and not
      500.

## Scope notes

The delete-path distinction (`MapDeleteError`) wasn't part of the original
report (scoped to create/update), but was necessary to avoid shipping a
new, factually incorrect message on delete as a side effect of fixing
create/update. Covered by tests in this PR rather than deferred.

## Validation

- [x] `go build ./...`
- [x] `go test ./...` (SQLite)
- [x] `go test -tags integration ./database -database.postgres-dsn ...`
      (real Postgres: FK-on-create, NOT NULL-on-create, FK-on-delete)
- [x] `go test -tags integration ./database -database.mysql-dsn ...`
      (real MySQL: same three scenarios)
- [x] `go test -tags integration ./auth -auth.postgres-dsn ...` (confirms
      `TranslateError` doesn't regress the duplicate-email registration path)
- [x] `go test -tags integration ./admin -admin.postgres-dsn ... -run TestResourcePostgres`
      (admin CRUD smoke test still green)
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run` (0 issues)
- [x] Project `code-review` skill run twice: first pass REQUEST CHANGES
      (delete-path message defect), second pass APPROVE after the fix

- [x] New behavior has tests; DB-touching changes pass the SQLite +
      PostgreSQL + MySQL matrix — verified against real containers, not
      fabricated error strings, since the driver message text differs per
      database (confirmed empirically rather than assumed).
- [ ] Docs + example — N/A, bugfix to existing validated behavior, not a
      new stable feature.
- [ ] Extraction preserves contracts — N/A, not an extraction.
- [ ] Generators — N/A, no generator touched; `MapPersistError`'s call-site
      signature is unchanged, so no generated app or golden template needs
      regeneration.
- [ ] API changes regenerate OpenAPI + TS client — N/A, the D10 error
      envelope shape is unchanged; only which category a given driver
      error maps to is corrected.
- [ ] No secrets in generated frontend — N/A, no frontend touched.
- [x] Scope stays inside this issue's milestone — contained to
      database/admin error classification, no M6 battery creep.
- [x] This PR links its issue and states which acceptance criteria it
      satisfies.

### DIFF
--- a/admin/resources.go
+++ b/admin/resources.go
@@ -206,7 +206,7 @@ func (h *handlers) deleteResource(ctx context.Context, input *itemInput) (*delet
 		return nil, contract.WithContext(ctx, contract.Internal("admin database is not attached"))
 	}
 	if err := db.WithContext(ctx).Delete(inst).Error; err != nil {
-		return nil, database.MapPersistError(ctx, err, "resource already exists", "persist resource")
+		return nil, database.MapDeleteError(ctx, err, "resource is still referenced by other records", "delete resource")
 	}
 	return &deleteOutput{Body: contract.Data[deleteResult]{Data: deleteResult{OK: true}}}, nil
 }
--- a/admin/resources_test.go
+++ b/admin/resources_test.go
@@ -596,6 +596,111 @@ func TestResourceBelongsToStoresFK(t *testing.T) {
 	}
 }
 
+func TestResourceCreateForeignKeyViolationIsValidationError(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	type FKCategory struct {
+		ID uint `gorm:"primaryKey" json:"id"`
+	}
+	type FKItem struct {
+		ID         uint       `gorm:"primaryKey" json:"id"`
+		Name       string     `json:"name"`
+		CategoryID uint       `json:"category_id"`
+		Category   FKCategory `gorm:"constraint:OnUpdate:CASCADE,OnDelete:RESTRICT;" json:"-"`
+	}
+	app := newCookieApp(t)
+	if err := app.DB().AutoMigrate(&FKCategory{}, &FKItem{}); err != nil {
+		t.Fatalf("AutoMigrate: %v", err)
+	}
+	if err := admin.Register(app, FKItem{}, admin.Options{
+		Slug: "fk-items",
+		Fields: []admin.Field{
+			{Name: "id", Type: admin.TypeInteger, ReadOnly: true},
+			{Name: "name", Type: admin.TypeString},
+			{
+				Name:    "category_id",
+				Type:    admin.TypeRelation,
+				Related: &admin.Relation{Slug: "fk-categories", Kind: admin.RelBelongsTo, LabelField: "id"},
+			},
+		},
+	}); err != nil {
+		t.Fatalf("Register: %v", err)
+	}
+	jar := loginSuperuser(t, app)
+
+	// Unlike TestResourceBelongsToStoresFK, FKItem.Category is a real GORM
+	// association with a DB-level constraint, so a nonexistent category_id
+	// reaches the database as an actual foreign-key violation, not just
+	// stored as an opaque integer.
+	rec := doRequest(app, jar, http.MethodPost, "/api/v1/admin/resources/fk-items", `{"name":"Nail","category_id":999}`)
+	assertError(t, rec, http.StatusUnprocessableEntity, contract.CodeValidationError)
+}
+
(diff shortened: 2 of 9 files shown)


leo-aa88 (approved):
LGTM

Issue #1, opened 2026-08-15
[M0-1] Create repo, module path, CI skeleton


leo-aa88:
## Scope
new repo, `go.mod` with `github.com/LAA-Software-Engineering/gombit`, MIT license, golangci-lint, GitHub Actions running `go test ./...` + lint.

## Acceptance criteria
- [ ] green CI on an empty package
- [ ] branch protection on `main`

## Size
S

## Dependencies
D1

## Definition of done
This issue also follows the Agent Working Agreement in docs/GOMBIT_BUILD_PLAN.md §5.

leo-aa88:
Implementation plan for M0-1:

1. Create a focused branch from `main`.
2. Add the root Go module `github.com/LAA-Software-Engineering/gombit` with a minimal package and smoke test so `go test ./...` has a stable target.
3. Add golangci-lint configuration and a GitHub Actions workflow that runs `go test ./...` and lint.
4. Add contributor/working-agreement docs that point back to the build plan.
5. Run the checks locally, open a draft PR linked to this issue, then verify or configure `main` branch protection.

leo-aa88:
PR opened for M0-1: https://github.com/LAA-Software-Engineering/gombit/pull/39

Local validation and GitHub Actions are green. Branch protection verification is blocked by GitHub API HTTP 403 because branch protection is unavailable unless the repository is public or the account has GitHub Pro for private repos.

leo-aa88:
PR #39 for the M0-1 code/CI skeleton has merged: https://github.com/LAA-Software-Engineering/gombit/pull/39

Current status:

- `go.mod`, minimal package/test, golangci-lint config, GitHub Actions test/lint jobs, and contributor docs are merged.
- Latest PR checks were green before merge: `Test` passed and `Lint` passed.
- Remaining acceptance criterion: branch protection on `main`.

Branch protection is still blocked by repository settings. The repo is currently private, and GitHub returns HTTP 403 for branch protection reads with: "Upgrade to GitHub Pro or make this repository public to enable this feature." I cannot complete or verify this AC without either making the repository public or enabling private-repo branch protection through the account/plan settings.

PR #238, opened 2026-08-29
[feature] admin relations: Postgres/MySQL matrix + example app (#223)

### LINKED ISSUE #223: [feature] Admin support for relationships: belongs-to/has-many pickers and many-to-many
### Problem

The runtime admin cannot represent relationships. With derived fields
(`admin.FieldsFrom`), a belongs-to FK shows as a bare `integer` text input (no
picker, no target model), and a many-to-many field is dropped entirely (skipped
when its DB column name is empty). `admin.TypeRelation` exists with
`belongs_to`/`has_many`, but `options.go` states "Many-to-many is deferred", and
`inferFieldType` collapses any struct/slice to `TypeJSON`. See bug #221.

For a domain with shared inventory (engine <-> warehouse m2m) or ownership
history, the admin simply cannot edit the relationships, which undercuts the
"your registry is the admin" promise for anything past a flat table.

### Proposed surface

- A relation **widget** for `belongs_to`/`has_many`: render as a select backed by
  the related model's admin list endpoint, showing a label field, not a raw id.
- **Many-to-many** support end to end: registry metadata (`admin.Field` with a
  m2m `Relation`), a data-plane read that returns related ids/labels, and a write
  path that syncs the join table.
- Auto-derivation should emit `TypeRelation` (with the target slug) for
  association fields instead of flattening to integer/JSON.

### Django (or other framework) equivalent

Django admin: `ForeignKey` renders as a select (`raw_id_fields`/autocomplete for
large sets), `ManyToManyField` renders as a multi-select / `filter_horizontal`.

### Where do you think this belongs?

admin

### Alternatives and workarounds

Manage relationships by typing raw integer ids into the FK column (no validation
that the target exists for no-FK columns), and manage m2m rows directly in the
database. There is no admin path for m2m at all today.

leo-aa88:
## Summary

Closes out the two remaining **#223** leftovers I flagged after the admin relationship widgets shipped (#231/#233/#235/#236):

1. **Postgres/MySQL integration matrix for relations.** `admin/integration_test.go` adds `TestRelationsPostgres` / `TestRelationsMySQL`, running the belongs_to / has_many / many_to_many admin round-trips against real Postgres and MySQL — the DB-touching paths (join-table sync, transactional rollback on a bad related id, belongs_to FK write, has_many preload) that emit engine-specific SQL and can't be covered by SQLite alone. They run under the **existing** `go test -tags integration ./admin -admin.{postgres,mysql}-dsn` CI jobs — no workflow change.

2. **Example app.** `examples/admin` gains `warehouse.Warehouse` (belongs_to + m2m target) and `part.Part` (has_many child). `Widget` now carries all three relation kinds and registers with **empty `Fields`**, so `admin.FieldsFrom` derives them — the headline #223 behavior — documented in the README.

3. **Auto-derivation fix (found by building the example).** `many_to_many` omitted `LabelField` while belongs_to/has_many set it, so the multi-select rendered raw ids. `FieldsFrom` now defaults the m2m label the same way.

Closes #223.

## Acceptance criteria (issue #223)

- [x] belongs_to/has_many render as pickers backed by the related list, labelled not raw ids — **now consistent across all three kinds** (m2m label fixed here; belongs_to/has_many from #233/#236)
- [x] many_to_many end to end: registry metadata, data-plane read of ids/labels, write that syncs the join table — **now matrix-tested on PG + MySQL**
- [x] Auto-derivation emits `TypeRelation` with the target slug for association fields — exercised by the example's empty-`Fields` registration

## Validation

- [x] `go test ./...` (SQLite) + `golangci-lint run ./admin/... ./examples/...` — 0 issues
- [x] **Real DB matrix**, verified locally against `postgres:16-alpine` and `mysql:8.4`:
  ```
  TestResourcePostgres PASS   TestResourceMySQL PASS
  TestRelationsPostgres PASS  TestRelationsMySQL PASS
  ```
  (also with CI's exact MySQL DSN, no `multiStatements`)
- [x] Example runs: logged in, created warehouses, created a widget with `warehouse_id` (belongs_to) + `warehouses:[1,2]` (m2m) — both round-trip; meta advertises `belongs_to` / `many_to_many` / read-only `has_many`, m2m `label_field` now `name`.

## Working agreement

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — **this PR is that matrix coverage**
- [x] Stable feature appears in an example app (examples/admin) with docs (README + links to docs/admin.md, docs/cli.md)
- [x] Extraction preserves contracts — reuses existing helpers; no rewrite of passing tests
- [x] No API contract change → no OpenAPI/TS regen needed (admin data plane is generic; N/A)
- [x] No secrets in frontend source
- [x] Scope stays in M6 admin (#223); no new batteries
- [x] Links its issue and states which acceptance criteria it satisfies

### DIFF
--- a/admin/fields.go
+++ b/admin/fields.go
@@ -88,8 +88,9 @@ func FieldsFrom(model any) ([]Field, error) {
 			Name: relationFieldName(rel.Field),
 			Type: TypeRelation,
 			Related: &Relation{
-				Kind: RelManyToMany,
-				Slug: rel.FieldSchema.Table,
+				Kind:       RelManyToMany,
+				Slug:       rel.FieldSchema.Table,
+				LabelField: labelFieldFor(rel.FieldSchema),
 			},
 		})
 	}
--- a/examples/admin/internal/widget/widget.go
+++ b/examples/admin/internal/widget/widget.go
@@ -16,8 +16,12 @@ type Widget struct {
 	Name string `json:"name"`
 	SKU  string `json:"sku"`
 
-	// belongs_to: the widget's primary warehouse (FK + association).
-	WarehouseID uint                `gorm:"index" json:"warehouse_id"`
+	// belongs_to: the widget's primary warehouse (FK + association). The FK is a
+	// *uint so it is genuinely optional: a widget with no warehouse stores NULL
+	// rather than 0, which a non-nullable uint could not represent under foreign
+	// key enforcement (the admin advertises belongs_to as optional, so the create
+	// must actually accept "none").
+	WarehouseID *uint               `gorm:"index" json:"warehouse_id"`
 	Warehouse   warehouse.Warehouse `json:"-"`
 
 	// many_to_many: every warehouse this widget is stocked at (join table). The
--- a/examples/admin/internal/widget/widget.go
+++ b/examples/admin/internal/widget/widget.go
@@ -1,12 +1,32 @@
 package widget
 
-import "time"
+import (
+	"time"
 
-// Widget is the small model registered with the admin in this example.
+	"github.com/gombit-dev/gombit/examples/admin/internal/part"
+	"github.com/gombit-dev/gombit/examples/admin/internal/warehouse"
+)
+
+// Widget is the model registered with the admin in this example. It carries all
+// three relation kinds so the admin renders their widgets from the registry
+// alone (ADR-013 / #223): a belongs_to picker, a many_to_many multi-select, and
+// a read-only has_many view.
 type Widget struct {
-	ID        uint      `gorm:"primaryKey" json:"id"`
-	Name      string    `json:"name"`
-	SKU       string    `json:"sku"`
+	ID   uint   `gorm:"primaryKey" json:"id"`
+	Name string `json:"name"`
+	SKU  string `json:"sku"`
+
+	// belongs_to: the widget's primary warehouse (FK + association).
+	WarehouseID uint                `gorm:"index" json:"warehouse_id"`
+	Warehouse   warehouse.Warehouse `json:"-"`
+
+	// many_to_many: every warehouse this widget is stocked at (join table). The
+	// JSON name is the admin field name the multi-select binds to.
+	Warehouses []warehouse.Warehouse `gorm:"many2many:widget_warehouses;" json:"warehouses"`
+
+	// has_many: the widget's parts (read-only in the admin; Part carries the FK).
+	Parts []part.Part `json:"parts"`
+
 	CreatedAt time.Time `json:"created_at"`
 	UpdatedAt time.Time `json:"updated_at"`
 }
--- a/examples/admin/main.go
+++ b/examples/admin/main.go
@@ -36,8 +36,9 @@ func main() {
 		_ = db.Close()
 		log.Fatal(err)
 	}
-	// Register relation targets before Widget so its belongs_to / many_to_many
-	// pickers resolve against registered slugs.
+	// Register every model. Order is not significant: the belongs_to /
+	// many_to_many pickers resolve against the target slug at request time (the
+	// SPA calls the target's list endpoint), not at Register time.
 	for _, register := range []func(*framework.App) error{
 		warehouse.RegisterAdmin,
 		part.RegisterAdmin,
--- a/examples/admin/main.go
+++ b/examples/admin/main.go
@@ -8,6 +8,8 @@ import (
 	"github.com/gombit-dev/gombit/auth"
 	"github.com/gombit-dev/gombit/config"
 	"github.com/gombit-dev/gombit/database"
+	"github.com/gombit-dev/gombit/examples/admin/internal/part"
+	"github.com/gombit-dev/gombit/examples/admin/internal/warehouse"
 	"github.com/gombit-dev/gombit/examples/admin/internal/widget"
 	"github.com/gombit-dev/gombit/framework"
 )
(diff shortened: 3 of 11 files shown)


examples/admin/internal/widget/widget.go
```diff
@@ -1,12 +1,32 @@
...
 type Widget struct {
-	ID        uint      `gorm:"primaryKey" json:"id"`
-	Name      string    `json:"name"`
-	SKU       string    `json:"sku"`
+	ID   uint   `gorm:"primaryKey" json:"id"`
+	Name string `json:"name"`
+	SKU  string `json:"sku"`
+
+	// belongs_to: the widget's primary warehouse (FK + association).
+	WarehouseID uint                `gorm:"index" json:"warehouse_id"`
```
leo-aa88: **BLOCKING.** `WarehouseID uint` is non-nullable and the example runs SQLite with `_fk=1`, but auto-derivation advertises `warehouse_id` as `required:false`. A create with no warehouse (or `warehouse_id:0`) 422s — verified against this branch. So the belongs_to picker is optional in the meta and required at runtime. The README's E2E create `{"name":"Wrench","sku":"w-1"}` now fails. Either make this `*uint` (genuinely optional, if the admin handles nullable FK writes — the #237-deferred work) or make it honestly required and seed a warehouse + update the README.

examples/admin/main.go
```diff
@@ -34,16 +36,24 @@ func main() {
 		_ = db.Close()
 		log.Fatal(err)
 	}
-	if err := widget.RegisterAdmin(app); err != nil {
-		_ = db.Close()
-		log.Fatal(err)
+	// Register relation targets before Widget so its belongs_to / many_to_many
```
leo-aa88: **MINOR.** "so its belongs_to / many_to_many pickers resolve against registered slugs" — the pickers resolve at request time via the target's list endpoint, and `admin.Register` doesn't require the target registered first. The ordering is tidy, not required; the comment implies an invariant that isn't enforced. Soften it, or enforce the dependency in Register.

admin/fields.go
```diff
@@ -88,8 +88,9 @@ func FieldsFrom(model any) ([]Field, error) {
 			Name: relationFieldName(rel.Field),
 			Type: TypeRelation,
 			Related: &Relation{
-				Kind: RelManyToMany,
-				Slug: rel.FieldSchema.Table,
+				Kind:       RelManyToMany,
+				Slug:       rel.FieldSchema.Table,
+				LabelField: labelFieldFor(rel.FieldSchema),
```
leo-aa88, line 93: Correct fix, and the right one — m2m now labels consistently with belongs_to/has_many instead of rendering raw ids. Note there's still no coverage for a target model with no string field (`labelFieldFor` -> ""), but that gap predates this PR and is shared by all three kinds.

leo-aa88 (commented):
# REQUEST CHANGES

(Posting as COMMENT — GitHub won't let the author submit a formal REQUEST_CHANGES on their own PR. Treat this as a blocking review.)

The matrix tests are the right call and genuinely earn their keep: `assertManyToManyRoundTrip` exercises the join-table sync AND the bad-id 422 that must roll the scalar back inside one transaction — that is the exact behavior SQLite can hide and PG/MySQL enforce. The m2m `LabelField` fix is correct and now consistent with belongs_to/has_many. Verified against postgres:16 and mysql:8.4.

But the headline deliverable — "the stable feature appears in a **working** example" (working agreement §5.2) — is broken by this same PR, and its own README proves it.

## 1. The example's belongs_to is advertised optional but is de-facto required; the documented create now 422s

**BLOCKING (for this PR's own contract).**

`Widget.WarehouseID` is a non-nullable `uint`, and the example runs SQLite with `_fk=1`. Auto-derivation emits `warehouse_id` as a belongs_to with `required:false` (confirmed in `GET /api/v1/admin/meta/widgets`). So:

- THE META SAYS: `warehouse_id` is optional.
- THE RUNTIME DOES: reject any create without an existing warehouse.

Reproduced against this branch's binary:

```
POST /api/v1/admin/resources/widgets {"name":"Loose","sku":"l-1"}
-> 422 {"error":{"code":"validation_error","message":"The request references a resource that does not exist."}}
POST ... {"name":"Zero","sku":"z-1","warehouse_id":0}  -> 422 (same)
```

The README's own E2E create (`examples/admin/README.md`, unchanged in this PR) is `-d '{"name":"Wrench","sku":"w-1"}'` — no `warehouse_id`. On `main` that returns 200; on this branch it returns 422. The PR breaks the very commands it ships as the example's documented happy path. `( ͡° ͜ʖ ͡°)`

Root cause is the #237-deferred one, now made observable: an optional belongs_to backed by a non-nullable scalar FK cannot represent "no parent" under FK enforcement. Two honest fixes:

- (a) Make it genuinely optional: `WarehouseID *uint` so unset -> NULL. This depends on the generic admin actually handling a nullable/pointer FK on write — the deferred nullable-FK work. If the admin can't yet, (a) is not available, which is itself worth stating in the example.
- (b) Make it honestly required: seed a warehouse in `OnStart`, mark the field required, and update the README E2E to pass `warehouse_id`. Less impressive, but true.

What's not acceptable is shipping an example whose documented first create fails and whose meta lies about optionality.

## 2. Registration-order comment overstates a request-time fact

**MINOR.** `main.go` says targets are registered before Widget "so its belongs_to / many_to_many pickers resolve against registered slugs." The pickers resolve at **request** time (the SPA calls the target's list endpoint); `admin.Register` does not require the target to be registered first. The ordering is tidy, not load-bearing — the comment implies an invariant that doesn't exist. Either enforce it at Register (and keep the comment) or soften the comment.

# VERDICT

The tests and the derivation fix are good engineering. The defect is in the deliverable that the working agreement actually requires: the example does not work as documented, and its belongs_to picker's advertised contract (optional) contradicts runtime (required). This is an abstraction/contract problem surfaced by the example, not a test problem — CI is green because no test creates a widget without a warehouse. Fix #1 before merge; #2 is cleanup.

CI is green. This is not a failing-test problem. Add a test that creates the example's headline model with no belongs_to selected, and it will fail today.

leo-aa88:
Addressed the review in 45f57f9:

**Finding #1 (BLOCKING) — optional belongs_to rejected "none".** `Widget.WarehouseID` is now `*uint`, so an unset warehouse stores NULL and the create succeeds. The generic admin handles the nullable FK on write with no admin-side change — verified the README's E2E create `{"name":"Wrench","sku":"w-1"}` now returns 200 with `warehouse_id:null`. Regression tests added: `TestBelongsToOptionalNullable` (SQLite) and the matrix `assertBelongsToRoundTrip` now asserts the omitted-FK create succeeds/stores null on Postgres + MySQL (where FK enforcement bites). The nullable model is now the one the matrix registers.

**Finding #2 (MINOR) — misleading registration-order comment.** Corrected: the pickers resolve at request time against the target slug, so registration order isn't significant. README updated to document the `*uint`/optional behavior.

Validation: `go test ./...`, golangci-lint (default + `-tags integration`) clean, and the full PG/MySQL admin matrix green against postgres:16 / mysql:8.4.

Note: the "Cursor Automation: Code Reviewer" check is NEUTRAL because that automation is out of budget ("Background Agent requires at least $2 remaining until your hard limit"), not because of a review finding.

leo-aa88 (commented):
LGTM

Issue #243, opened 2026-08-30
[PERF-5] Benchmark suite has no parallel/contention row, so scaling regressions are invisible


leo-aa88:
## Summary

The BENCH-1 microbench suite (`benchmarks/micro`) runs each scenario single-goroutine, so it cannot surface contention/scaling bugs. The metrics-mutex scaling problem (PERF-1) was invisible to the existing framework-tax numbers and only appeared under a parallel driver. The suite should include a parallel/contention benchmark so scaling regressions are caught.

## Motivation

- PERF-1 measured only ~1.26x throughput from 1→16 cores; the existing `BenchmarkFrameworkTax` (single-goroutine) reports none of that.
- A contention benchmark is the acceptance-test vehicle for PERF-1 and PERF-2 (both are lock-contention findings).

## Proposed change

- Add a `RunParallel`-based benchmark to the shared `benchmarks/micro/scenario` harness (e.g. `RunParallelBenchmark`) so every row (nethttp/gin/huma/gombit) reports per-core scaling for at least the plaintext and valid-POST scenarios.
- Record 1-core vs N-core ns/op (or an explicit scaling ratio) in `benchmarks/results` and the schema (`benchmarks/docs/schema.md`).
- Optionally wire a scaling-ratio floor into the existing `benchmark-report-drift`/`benchmark-smoke` CI gates so a future regression in horizontal scaling fails CI.

## Acceptance criteria

- [ ] Parallel benchmark exists in the shared harness and runs for all four rows.
- [ ] Results/schema updated to carry the parallel numbers.
- [ ] Methodology doc (`benchmarks/docs/methodology.md`) explains what the parallel row measures and how to read it.

## Scope notes

Post-v0.1, benchmark-infra only (labels align with #141: infra/tests/ci). No runtime change. Enables PERF-1/PERF-2 verification.

Issue #240, opened 2026-08-30
[PERF-2] Per-request request/trace IDs use crypto/rand (getrandom syscall + global lock) unnecessarily


leo-aa88:
## Summary

Every request generates a request ID and a trace ID, and both use `crypto/rand.Read` (a `getrandom` syscall guarded by a global reader lock). Correlation IDs are not secrets, so paying cryptographic-RNG cost — plus a global lock — on every request is unnecessary overhead on the fixed per-request path.

- `newRequestID()` — [framework/request_id.go:92](../blob/main/framework/request_id.go#L92)
- `newTraceID()` — [framework/http_middleware.go:266](../blob/main/framework/http_middleware.go#L266)

Both run in `requestContextMiddleware` on every request.

## Evidence (measured)

Plaintext scenario, BENCH-1 harness, Go 1.25.7:

- CPU profile (single-threaded): `crypto/rand.Read` → `crypto/internal/sysrand.Read` → `GetRandom` syscall accounts for **~14% of total CPU** (two reads per request).
- Mutex profile (16 cores): `crypto/rand`'s global reader lock is ~12% of contention, second only to the metrics mutex (PERF-1).

## Proposed fix

Use a non-cryptographic, contention-free source for correlation IDs:
- `math/rand/v2` (its top-level functions are per-P and lock-free), or a pooled/per-P generator, to fill the 16 random bytes.
- Keep the existing output formats (UUIDv4-shaped request ID; 32-hex trace ID) so headers and downstream parsing are unchanged.

These IDs are opaque correlation tokens, not security tokens; predictability is not part of their contract. (If any consumer *does* depend on unpredictability, document that and scope this to trace IDs only.)

## Acceptance criteria

- [ ] Neither request-ID nor trace-ID generation calls `crypto/rand` on the request path.
- [ ] Output formats/lengths unchanged; existing request_id/trace tests pass.
- [ ] Benchmark shows the `getrandom` frames gone from the plaintext CPU profile and reduced ns/op.

## Scope notes

Post-v0.1 optimization, runtime-only, no contract change. Build plan §5 applies.

PR #244, opened 2026-08-30
perf: make metrics middleware hot path lock-free (#239)

### LINKED ISSUE #239: [PERF-1] Metrics middleware global mutex caps throughput at ~1 core (no horizontal scaling)
## Summary

`framework.metricsMiddleware` serializes every request on a single process-global `sync.Mutex`, so Gombit does not scale across CPU cores. Under concurrent load, throughput is capped near single-core regardless of how many cores are available.

The middleware takes the lock **three times per request**: `metrics.addActive(1)`, `defer metrics.addActive(-1)`, and `metrics.observe(...)` — all on the same `httpMetrics.mu` — [framework/http_middleware.go:160](../blob/main/framework/http_middleware.go#L160), backed by `addActive`/`observe` at L180-191.

## Evidence (measured)

Profiled on this repo at `main`, Go 1.25.7, 12th-gen i7 (16 logical CPUs), plaintext scenario via the BENCH-1 harness with a parallel `RunParallel` driver:

| GOMAXPROCS | ns/op |
|---|---|
| 1  | 4355 |
| 16 | 3450 |

**16 cores yield only ~1.26x the throughput of one core** — near-zero horizontal scaling. A `-mutexprofile` of the 16-core run attributes ~87% of contention to `sync.(*Mutex).Unlock` + `runtime.unlock` (the rest is `crypto/rand`'s global reader lock, tracked separately in PERF-2). A correctly parallel handler should scale close to linearly.

## Proposed fix

Remove the mutex from the hot path:
- `active` → `atomic.Int64` (increment/decrement, no lock).
- Per-key request counts and latency sums → either `atomic.Int64` values behind a `sync.Map`/sharded map keyed by `metricsKey`, or a sharded counter array. The render path can snapshot atomically.

The `/metrics` render path may keep a lock or snapshot-copy; only the per-request write path must be lock-free.

## Acceptance criteria

- [ ] Per-request metrics recording takes no shared lock.
- [ ] A parallel/contention benchmark (see PERF-5) shows near-linear scaling from 1→N cores on the plaintext path.
- [ ] `/metrics` output is unchanged (same series, same values under a single-threaded load).
- [ ] Existing metrics tests pass; add a concurrency test that asserts counts are exact under N goroutines.

## Scope notes

Post-v0.1 optimization; no API/contract change, no new battery. Runtime-only. Definition of done (build plan §5) applies: new behavior covered by tests; benchmark evidence in the PR.

leo-aa88:
## Summary

The metrics middleware acquired a single process-global `sync.Mutex` three times per request (`addActive` twice, `observe` once), serializing every request and capping throughput near a single core.

This makes the hot path lock-free: `active` is an `atomic.Int64`, and per-series counters live in a `sync.Map` keyed by `metricsKey`, each holding two `atomic.Int64` accumulators (request count + latency nanoseconds). The label set (method × route × status) is bounded, so after warmup the hot path is a lock-free `sync.Map.Load`; only the first request for a new series takes the `LoadOrStore` slow path. `render()` snapshots via `Range` + atomic `Load` and its output is unchanged.

**Measured** (parallel plaintext benchmark, BENCH-1 harness, Go 1.25.7, 16 logical CPUs):
- Before: 1→16 cores scaled only **1.26x**; mutex profile attributed ~87% of contention to this lock.
- After: scales **~1.95x**, and the application `sync.Mutex` is gone from the contention profile. The residual ceiling is now `crypto/rand`'s global lock, tracked in #240.

Closes #239

## Acceptance criteria

- [x] Per-request metrics recording takes no shared lock.
- [ ] A parallel/contention benchmark shows near-linear scaling from 1→N cores — **deferred to #243 (PERF-5)**, which adds the parallel row to the committed suite. Verified here with a throwaway benchmark (not committed); scaling improved 1.26x→1.95x, with the remaining ceiling being #240, not this lock.
- [x] `/metrics` output is unchanged (existing metrics tests pass byte-for-byte).
- [x] Existing metrics tests pass; added a concurrency test asserting exact counts under 32 goroutines × 500 requests and that the active gauge drains to zero (passes under `-race`).

## Scope notes

Runtime-only, post-v0.1 optimization; no API/contract change. The committed parallel benchmark is #243's scope, so the near-linear-scaling AC is verified here but its permanent regression guard lands with PERF-5. Full linear scaling also depends on #240 (crypto/rand per request).

## Validation

- [x] `go test ./framework/ -race` — passes (incl. new concurrency test)

- [ ] Stable features ship docs and appear in an example app — N/A, internal runtime optimization, no user-facing surface
- [x] Extraction preserves contracts — refactor and move, don't rewrite code that already passes its tests — `/metrics` output preserved

### DIFF
--- a/framework/http_middleware.go
+++ b/framework/http_middleware.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gin-gonic/gin"
@@ -106,11 +107,25 @@ func securityHeadersMiddleware(includeHSTS bool) gin.HandlerFunc {
 	}
 }
 
+// httpMetrics accumulates request metrics without locking the request path.
+// The per-request writers (addActive, observe) use atomics only, so the
+// metrics middleware no longer serializes concurrent requests on a single
+// process-global mutex (issue #239). The label set — method × route × status —
+// is bounded (normalizeMetricsMethod caps method cardinality, FullPath caps
+// route), so the sync.Map holding the per-series counters reaches a steady
+// state after warmup and the hot path settles into a lock-free Load.
 type httpMetrics struct {
-	mu       sync.Mutex
-	active   int64
-	requests map[metricsKey]int64
-	latency  map[metricsKey]time.Duration
+	active   atomic.Int64
+	counters sync.Map // metricsKey -> *metricCounter
+}
+
+// metricCounter holds the two accumulators for a single metrics series. Both
+// are updated with atomic adds and take no lock. latency is stored as int64
+// nanoseconds (time.Duration's underlying unit) so render can convert it back
+// with time.Duration(ns).Seconds() exactly as before.
+type metricCounter struct {
+	requests atomic.Int64
+	latency  atomic.Int64
 }
 
 type metricsKey struct {
@@ -151,10 +166,10 @@ func normalizeMetricsMethod(method string) string {
 }
 
 func newHTTPMetrics() *httpMetrics {
-	return &httpMetrics{
-		requests: make(map[metricsKey]int64),
-		latency:  make(map[metricsKey]time.Duration),
-	}
+	// The zero value is ready: atomic.Int64 starts at 0 and sync.Map needs no
+	// initialization. Counters are created lazily by observe on first sight of
+	// a series.
+	return &httpMetrics{}
 }
 
 func metricsMiddleware(metrics *httpMetrics) gin.HandlerFunc {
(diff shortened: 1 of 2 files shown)


leo-aa88 (commented):
LGTM.

Traced the hot path end-to-end: `addActive` → `atomic.Int64.Add`; `observe` → `counterFor` (lock-free `Load`, `LoadOrStore` only on first sight of a bounded series) → two atomic adds; `render` snapshots via `Range` + atomic `Load`. `metricsKey` is a comparable value, valid as a `sync.Map` key. Label set is bounded (`normalizeMetricsMethod` caps method, `FullPath` caps route), so the map reaches steady state and stays a lock-free `Load` — the correct use case for `sync.Map`. Concurrency test asserts exact counts under 32×500 and drains the active gauge to zero; passes under `-race` locally.

One non-blocking note, not a defect: `render` no longer takes a consistent snapshot the way the mutex version did — a concurrent `observe` can be seen between its `requests.Add` and `latency.Add`, skewing one series average by a single sample transiently. That is the standard trade for lock-free metrics and is fine for a scrape endpoint; the key sets stay matched because both accumulators live on one `metricCounter`.

PR #245, opened 2026-08-30
perf: generate correlation IDs without crypto/rand (#240)

### LINKED ISSUE #240: [PERF-2] Per-request request/trace IDs use crypto/rand (getrandom syscall + global lock) unnecessarily

leo-aa88:
## Summary

`newRequestID` and `newTraceID` each called `crypto/rand.Read` on every request — a `getrandom` syscall behind a globally locked reader. That cost ~14% of single-threaded plaintext CPU and was the top remaining cross-core contention point after the metrics mutex was removed (#239).

Request and trace IDs are opaque correlation tokens, not secrets, so they don't need a CSPRNG. This fills the 16 ID bytes from `math/rand/v2`, whose top-level generator is per-P and lock-free. Output formats are unchanged (UUIDv4-shaped request ID; 32-hex trace ID), and because `math/rand/v2` cannot fail, the dead error paths are removed.

**Measured**: after the change `crypto/rand` no longer appears in the plaintext CPU profile.

Closes #240

## Acceptance criteria

- [x] Neither request-ID nor trace-ID generation calls `crypto/rand` on the request path.
- [x] Output formats/lengths unchanged; existing request_id/trace tests pass. Added dedicated UUIDv4 + 32-hex format and uniqueness tests (10k draws each) and a traceparent round-trip test.
- [x] Benchmark shows the `getrandom` frames gone from the plaintext CPU profile.

## Scope notes

Runtime-only, post-v0.1 optimization; no contract change. IDs remain non-guessable-enough for correlation but are explicitly **not** security tokens — documented in the `randomBytes16` helper. If any consumer ever needs unpredictable IDs, that would be a separate, opt-in change.

## Validation

- [x] `go test ./framework/ -race` — passes (incl. new format/uniqueness tests)

- [ ] Stable features ship docs and appear in an example app — N/A, internal runtime optimization, no user-facing surface change
- [x] Extraction preserves contracts — ID formats preserved; existing trace/request-id tests unchanged

### DIFF
--- a/framework/correlation_id_test.go
+++ b/framework/correlation_id_test.go
@@ -1,0 +1,58 @@
+package framework
+
+import (
+	"regexp"
+	"testing"
+)
+
+// uuidV4Pattern matches the canonical UUIDv4 textual form: version nibble 4 and
+// variant nibble in [89ab]. newRequestID must keep emitting this shape after
+// switching its randomness source from crypto/rand to math/rand/v2 (issue #240).
+var uuidV4Pattern = regexp.MustCompile(`^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$`)
+
+// traceIDPattern is a 32-char lowercase hex string, the W3C trace-id form.
+var traceIDPattern = regexp.MustCompile(`^[0-9a-f]{32}$`)
+
+func TestNewRequestIDFormatAndUniqueness(t *testing.T) {
+	const n = 10000
+	seen := make(map[string]struct{}, n)
+	for range n {
+		id := newRequestID()
+		if !uuidV4Pattern.MatchString(id) {
+			t.Fatalf("newRequestID() = %q, want UUIDv4 form", id)
+		}
+		if _, dup := seen[id]; dup {
+			t.Fatalf("newRequestID() produced a duplicate %q within %d draws", id, n)
+		}
+		seen[id] = struct{}{}
+	}
+}
+
+func TestNewTraceIDFormatAndUniqueness(t *testing.T) {
+	const n = 10000
+	seen := make(map[string]struct{}, n)
+	for range n {
+		id := newTraceID()
+		if !traceIDPattern.MatchString(id) {
+			t.Fatalf("newTraceID() = %q, want 32-char lowercase hex", id)
+		}
+		if id == "00000000000000000000000000000000" {
+			t.Fatalf("newTraceID() returned the all-zero trace ID, which is invalid")
+		}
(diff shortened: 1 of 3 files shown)


leo-aa88 (commented):
LGTM.

Contract check: request/trace IDs are correlation tokens, not secrets — `requestContextMiddleware` even honors a client-supplied `X-Request-Id`, so unpredictability was never part of the contract. Dropping `crypto/rand` for `math/rand/v2` is justified; its top-level generator is ChaCha8-backed and concurrency-safe, so no lock/syscall on the request path. Formats preserved (UUIDv4 nibbles set in `newRequestID`; `newTraceID` keeps the all-zero guard). Removing the dead error paths is correct since `math/rand/v2` cannot fail; the residual `== "" -> "unknown"` fallbacks in the middleware are now unreachable but harmless. Format + 10k-uniqueness + traceparent round-trip tests pass under `-race`.

Issue #201, opened 2026-08-27
[bug] XSS middleware silently discards all text after an unclosed <script>/<style>/<textarea> tag (data loss)


leo-aa88:
### What happened

The XSS sanitizer (`stripHTML` in `framework/xss.go`, wired into the default
runtime middleware stack for every app) discards **all** text that follows an
**unclosed** "skip element" start tag — `<script>`, `<style>`, `<textarea>`,
`<iframe>`, `<object>`, `<embed>`, `<noscript>`. Because the tokenizer enters
skip mode on the start tag and never leaves it without a matching end tag,
every subsequent text token is dropped and the truncated value is what reaches
(and is persisted by) the handler.

This is distinct from the already-fixed #118 (incomplete `a<b` tags): here the
input contains a *complete, well-formed* start tag, but no closing tag, so it
matches `completeHTMLTag` and triggers the skip path.

Realistic trigger — a plain text/description field whose legitimate content
merely mentions one of these element names:

```
input:  "To embed a video use the <iframe> element with a src attribute"
stored: "To embed a video use the "
```

```
input:  "Wrap long input in a <textarea> so it scrolls"
stored: "Wrap long input in a "
```

A properly *closed* `<script>...</script>foo` keeps `foo`, so the data loss is
specific to unbalanced markup, which is easy to produce in prose, docs, code
snippets, or CMS-style content.

### Steps to reproduce

```
1. gombit new demo   (any auth mode)
2. Add a resource with a text field, e.g. gombit make resource note --fields "body:text"
3. POST /api/v1/notes  {"body":"Use the <iframe> tag to embed"}
4. GET the row back: body is stored as "Use the " — everything after the
   unclosed skip tag was silently discarded.
```

### Impact

Silent data loss / corruption of user-submitted content in any string field,
for all apps (the XSS middleware is on by default). The value is truncated
without any error, so neither the client nor the server logs indicate the loss.

### Suggested fix

For an unclosed skip element, strip the tag itself but keep the following text
as plain text (the sanitizer already removes tags, so trailing text carries no
executable markup), rather than dropping the remainder of the string. At
minimum, treat an unterminated skip element as "strip the tag, keep the text"
instead of fail-closed silent truncation.

Reproduces on `main`.

Issue #118, opened 2026-08-24
[bug] XSS middleware silently truncates JSON strings that look like HTML tags (e.g. a<b)


leo-aa88:
### What happened

`framework.xssMiddleware` rewrites JSON string values (and GET query values) through `stripHTML` whenever the string contains `<` or `>`. The HTML tokenizer treats `<` followed by a letter as a start tag, so **non-HTML text is silently shortened** before the handler sees it.

Reproduced with the same tokenizer `stripHTML` uses:

```
"a<b"            -> "a"
"foo <bar> baz"  -> "foo  baz"
"a < b"          -> "a < b"          (spaces: preserved)
```

A `POST /api/v1/products` body `{"name":"a<b","price":1}` is stored as `name=a`. The client sent a real name; the database gets a different value with no validation error.

`framework/xss.go` `stripHTML` only short-circuits when the string has no `<>` at all. `sanitizeJSONBody` then `json.Marshal`s the mutated payload back onto `Request.Body`.

Expected: comparison text and names that are not actual HTML should be unchanged (or rejected as invalid input), not silently truncated. Tag-like substrings without a real document (`a<b`, `foo <bar> baz`) are the failing cases.

### Steps to reproduce

```sh
gombit new demo --database sqlite
cd demo
# migrate + start the API
curl -s -H 'Content-Type: application/json' \
  -d '{"name":"a<b","price":1}' \
  http://127.0.0.1:8080/api/v1/products
curl -s http://127.0.0.1:8080/api/v1/products
```

The created row’s `name` is `a`, not `a<b`. Same for any Huma JSON string field (admin create/update included).

### gombit version

dev

### Environment

```
gombit:   dev
commit:   0f43797
built:    unknown
go:       go1.25.7
platform: linux/amd64
```

Checked on `main` at 0f43797.

### Database

sqlite

### Auth mode

not auth related

### Area

runtime (framework.App, routing, lifecycle)

### Logs or output

Request:

```json
{"name":"a<b","price":1}
```

Stored / listed:

```json
{"data":[{"id":1,"name":"a","price":1}],"meta":{...}}
```

### Before submitting

- [x] I searched existing issues for a duplicate
- [x] This reproduces on the latest release or on `main`

PR #246, opened 2026-08-30
perf: skip XSS JSON decode/re-encode when body has no markup (#241)

### LINKED ISSUE #241: [PERF-3] XSS middleware decodes+re-encodes every JSON body, redundant with Huma's typed decode
## Summary

The XSS input sanitizer fully decodes and re-encodes every JSON request body, redundantly with Huma's own typed decode that runs immediately afterward. `sanitizeJSONBody` does: `io.ReadAll` → `json.NewDecoder(...).Decode(&any)` into `map[string]any` → recursive `sanitizeJSONValue` → `json.Marshal` → new body reader — [framework/xss.go:119](../blob/main/framework/xss.go#L119).

So a POST body is parsed twice (once here into `any`, once by Huma into the typed struct) and re-serialized once, on every mutating request.

## Evidence (measured)

valid-POST scenario, BENCH-1 harness, Go 1.25.7:

- CPU: `sanitizeJSONBody` is **15.4% of total CPU** on the POST path.
- Allocations (`-memprofile`, alloc_objects), the redundant work alone:
  - `decoder.Decode(&payload)` into `any`: ~5.9M objects
  - `json.Marshal(payload)`: ~4.7M objects
  - `NewDecoder`/`bytes.NewReader`: ~1.6M
  - `io.ReadAll` + `LimitReader` + `NopCloser` + `sanitizeJSONValue`: several M more
  - Together ≈ **~28% of all POST allocations**.

This is distinct from #137 (size cap) and #232 (webhook HMAC correctness) — both already handled. This issue is purely the cost of the decode/re-encode round-trip.

## Proposed options

Reduce or eliminate the redundant round-trip. Candidates (pick per correctness review):
1. **Skip the re-encode when nothing changed** — sanitize in place and only re-marshal if `sanitizeJSONValue` actually mutated a value (track a `changed` flag, as `sanitizeQuery` already does at L106). Most bodies contain no HTML and would then avoid the `json.Marshal` entirely.
2. **Move sanitization to the typed layer** — sanitize decoded string fields after Huma binds the struct (e.g. a Huma transformer / resolver), so the body is parsed once.
3. **Streaming token rewrite** — strip HTML while scanning tokens without building a full `map[string]any`.

Option 1 is the smallest, lowest-risk win and keeps current semantics.

## Acceptance criteria

- [ ] Bodies with no sanitizable content incur no `json.Marshal` re-encode.
- [ ] Existing XSS tests (incl. #118, #201 truncation cases) still pass; sanitized bytes handlers receive are unchanged for inputs that *do* contain HTML.
- [ ] Benchmark shows reduced POST ns/op and allocs; `sanitizeJSONBody` CPU share drops.

## Scope notes

Post-v0.1 optimization, runtime-only. Must not change sanitization behavior — coordinate with the XSS correctness tests. Build plan §5 applies.

leo-aa88:
## Summary

The XSS input sanitizer decoded every JSON request body into `map[string]any` and re-marshaled it, redundantly with Huma's own typed decode that runs immediately after. On the valid-POST microbench that round trip was **~15% of CPU** and **~28% of allocations**.

Two changes:
- **Fast path** — an HTML tag can only reach a decoded string as a literal `<`/`>` or a `\uXXXX` escape, and every JSON escape carries a backslash. A body with none of `<`, `>`, or `\` cannot contain a tag in any string value or key, so sanitization is a guaranteed no-op: skip the decode + re-encode entirely and pass the original bytes through. This is the common case for API traffic.
- **Changed-flag** — when a body *is* decoded, only `json.Marshal` it if `sanitizeJSONValue` actually changed a value; otherwise the original bytes pass through untouched.

Bodies that contain markup are sanitized exactly as before.

**Measured** (valid-POST microbench): `83 → 60 allocs/op` (~28% fewer), `~8340 → ~6937 ns/op` (~17% faster); invalid-POST `93 → 73 allocs/op`.

Closes #241

## Security note

The backslash guard in the fast path is load-bearing: an escaped tag like `{"c":"<script>…"}` has no literal `<`/`>` but MUST still be sanitized. It carries a backslash, so it correctly bypasses the fast path and is decoded + stripped. `TestSanitizeJSONBodyFastPathDoesNotBypassEscapedTags` fails loudly if that guard is ever dropped.

## Acceptance criteria

- [x] Bodies with no sanitizable content incur no `json.Marshal` re-encode (and, via the fast path, no decode either).
- [x] Existing XSS tests still pass (incl. #118 `a<b`, #201 truncation, raw-body exemption); sanitized bytes are unchanged for inputs that contain HTML.
- [x] Benchmark shows reduced POST ns/op and allocs; `sanitizeJSONBody`'s decode/marshal is skipped for clean bodies.

## Scope notes

Runtime-only, post-v0.1 optimization; sanitization behavior is unchanged for any body that contains markup. A side effect (improvement): clean bodies now reach the handler byte-for-byte instead of being JSON-normalized/`<`-escaped by the re-marshal — locked by `TestSanitizeJSONBodyCleanBodyPassesThroughVerbatim`. Moving sanitization to the typed layer (parse once) remains a possible larger follow-up but is out of scope here.

## Validation

- [x] `go test ./framework/ -race` — passes (incl. new bypass-safety + verbatim tests)

- [ ] Stable features ship docs and appear in an example app — N/A, internal runtime optimization; behavior unchanged
- [x] Extraction preserves contracts — sanitization semantics preserved for markup-bearing bodies; existing XSS tests unchanged

### DIFF
--- a/framework/xss.go
+++ b/framework/xss.go
@@ -149,9 +149,10 @@ func sanitizeJSONBody(c *gin.Context) {
 	// carry a tag in any string value or key, so sanitization is a guaranteed
 	// no-op. Skip the decode + re-encode round trip entirely and hand the
 	// original bytes to the handler unchanged. This is the common case for API
-	// traffic. The escaped-bypass reasoning is locked by
+	// traffic. ContainsAny scans the (up to 8 MiB) body once; the escaped-bypass
+	// reasoning — why '\' must be in the set — is locked by
 	// TestSanitizeJSONBodyFastPathDoesNotBypassEscapedTags.
-	if bytes.IndexByte(raw, '<') < 0 && bytes.IndexByte(raw, '>') < 0 && bytes.IndexByte(raw, '\\') < 0 {
+	if !bytes.ContainsAny(raw, "<>\\") {
 		c.Request.Body = io.NopCloser(bytes.NewReader(raw))
 		return
 	}
--- a/framework/xss.go
+++ b/framework/xss.go
@@ -143,6 +143,19 @@ func sanitizeJSONBody(c *gin.Context) {
 		return
 	}
 
+	// Fast path (issue #241): an HTML tag can only reach a decoded JSON string
+	// as a literal '<'/'>' or as a "\uXXXX" escape, and every JSON escape needs
+	// a backslash. A body containing none of '<', '>', or '\' therefore cannot
+	// carry a tag in any string value or key, so sanitization is a guaranteed
+	// no-op. Skip the decode + re-encode round trip entirely and hand the
+	// original bytes to the handler unchanged. This is the common case for API
+	// traffic. The escaped-bypass reasoning is locked by
+	// TestSanitizeJSONBodyFastPathDoesNotBypassEscapedTags.
+	if bytes.IndexByte(raw, '<') < 0 && bytes.IndexByte(raw, '>') < 0 && bytes.IndexByte(raw, '\\') < 0 {
+		c.Request.Body = io.NopCloser(bytes.NewReader(raw))
+		return
+	}
+
 	decoder := json.NewDecoder(bytes.NewReader(raw))
 	decoder.UseNumber()
 	var payload any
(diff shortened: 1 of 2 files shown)


leo-aa88 (commented):
LGTM.

The fast-path guard is sound and the security reasoning holds: a `<`/`>` can only reach a decoded string as a literal byte or via a `\uXXXX` escape (which carries a backslash), and `0x3C`/`0x3E` never appear inside a multibyte UTF-8 sequence, so `IndexByte` cannot miss one — a body free of `<`, `>`, and `\` cannot yield a tag, and `stripHTML` is a guaranteed no-op. `TestSanitizeJSONBodyFastPathDoesNotBypassEscapedTags` locks the load-bearing backslash check. The changed-flag path is also a strict improvement: clean bodies now reach the handler verbatim instead of being HTML-escaped by `json.Marshal` re-encoding (`a & b` → `a \u0026 b`), and since the decoded value is identical and raw-body signature consumers are already exempt via `WithRawBodyPaths`, no contract breaks. Existing XSS tests + new verbatim/bypass tests pass under `-race`.

PR #247, opened 2026-08-30
perf: skip request-timeout re-wrap when a tighter deadline exists (#242)

### LINKED ISSUE #242: [PERF-4] Request-timeout middleware allocates a timer+context per request on the fast path
## Summary

`requestTimeoutMiddleware` calls `context.WithTimeout` on every request, which allocates a `time.Timer` plus a derived context — even for handlers that complete in well under a microsecond. On the fixed per-request path this is pure overhead for the common fast case.

- [framework/http_middleware.go:29](../blob/main/framework/http_middleware.go#L29)

## Evidence (measured)

Plaintext scenario, BENCH-1 harness, alloc profile: `context.WithDeadlineCause` (~11%) + `time.newTimer` (~3%) ≈ **~15% of plaintext allocations** come from the per-request timeout context.

## Proposed fix

This is the lowest-priority of the PERF set because context deadlines are inherently allocating; treat it as investigate-and-measure rather than a guaranteed win:
- Consider whether the request timeout can share the server's `ReadTimeout`/`WriteTimeout` for the common path and only install a per-request `context.WithTimeout` where a distinct, shorter deadline is actually configured.
- Or gate the middleware so a zero/disabled timeout is a true no-op (already partly done at L30) and document the trade-off of the timer allocation when enabled.

Do not regress the cancellation semantics that downstream handlers/DB calls rely on.

## Acceptance criteria

- [ ] A measured reduction in per-request allocations on the plaintext path, **or** a written conclusion that the timer allocation is unavoidable given the cancellation contract (close as won't-fix with the rationale).
- [ ] Timeout behavior unchanged: requests exceeding the configured timeout still cancel.

## Scope notes

Post-v0.1 optimization, runtime-only. Lowest priority in the PERF batch. Build plan §5 applies.

leo-aa88:
## Summary

`requestTimeoutMiddleware` calls `context.WithTimeout` on every request, allocating a `time.Timer` plus a shallow `Request` copy — ~15% of plaintext-path allocations.

**Honest framing (this is the P2 issue):** the timer is *intrinsic* to context-based cancellation. It is what propagates the deadline to handlers and DB calls when the request carries no deadline of its own — the common case — so it cannot be removed there without weakening cancellation semantics. What this PR removes is the genuinely redundant case: a request that **already** has a deadline at or before our timeout. Re-wrapping it only allocates a second timer that can never fire first (context always honors the earliest deadline), so the guard skips the wrap and hands the handler the original context unchanged.

Closes #242

## Acceptance criteria

The issue allows either a measured reduction on the plaintext path **or** a written conclusion that the timer is unavoidable there. This PR does the latter, plus a real (narrow) reduction:

- [x] Written conclusion: on the plaintext/no-inbound-deadline path the `context.WithTimeout` timer is unavoidable given the cancellation contract — documented inline in the middleware.
- [x] Real reduction where it *is* avoidable: requests that arrive with a tighter-or-equal deadline no longer allocate a redundant second timeout context.
- [x] Timeout behavior unchanged: requests with no deadline still get one (`TestRequestTimeoutMiddlewareImposesDeadlineWhenNonePresent`, existing `TestDefaultRouterRequestTimeoutSetsDeadline`); a disabled timeout stays a no-op.

## Scope notes

Runtime-only, post-v0.1, lowest-priority of the PERF batch. No plaintext alloc change (the timer is intrinsic there). A `BenchmarkRequestTimeoutMiddleware` is added to guard the enabled path against regressions. Deeper avoidance (a pooled custom deadline context) was considered and rejected as not worth the cancel/reset-race complexity for a P2.

## Validation

- [x] `go test ./framework/ -race` — passes (incl. new guard/identity tests)

- [ ] Stable features ship docs and appear in an example app — N/A, internal runtime optimization; behavior unchanged
- [x] Extraction preserves contracts — timeout/cancellation semantics preserved; existing timeout test unchanged

### DIFF
--- a/framework/http_middleware.go
+++ b/framework/http_middleware.go
@@ -34,7 +34,21 @@ func requestTimeoutMiddleware(timeout time.Duration) gin.HandlerFunc {
 	}
 
 	return func(c *gin.Context) {
-		ctx, cancel := context.WithTimeout(c.Request.Context(), timeout)
+		ctx := c.Request.Context()
+		// If the request already carries a deadline at or before the one we
+		// would impose, wrapping it again only allocates a second timer and a
+		// shallow Request copy that can never take effect first — context
+		// always honors the earliest deadline, which is the existing one. Skip
+		// the wrap in that case (issue #242). When no deadline (or a later one)
+		// is present — the common case — the context.WithTimeout timer is
+		// intrinsic to propagating cancellation to handlers and DB calls, so it
+		// stays.
+		if deadline, ok := ctx.Deadline(); ok && !deadline.After(time.Now().Add(timeout)) {
+			c.Next()
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(ctx, timeout)
 		defer cancel()
 		c.Request = c.Request.WithContext(ctx)
 		c.Next()
--- a/framework/request_timeout_test.go
+++ b/framework/request_timeout_test.go
@@ -1,0 +1,105 @@
+package framework
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/gin-gonic/gin"
+)
+
+// TestRequestTimeoutMiddlewareSkipsWrapWhenTighterDeadlineExists locks the
+// issue #242 guard: when the incoming request already carries a deadline at or
+// before the middleware's timeout, the middleware must not allocate a second
+// timeout context — the handler must observe the exact same context instance it
+// came in with. Context identity is the discriminator: a re-wrap would hand the
+// handler a different (child) context even though the effective deadline is
+// unchanged.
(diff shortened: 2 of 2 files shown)


leo-aa88 (commented):
LGTM.

Guard is correct: when the inbound deadline is at or before `now+timeout`, `context.WithTimeout` would only produce a child whose effective deadline is the pre-existing (tighter) one, so skipping the wrap is behavior-preserving — and correct on ownership grounds too, since we do not own the caller's context and should not attach our own cancel to it. Already-expired inbound deadlines fall into the skip branch and hand the handler an expired context, which matches what re-wrapping would have produced. The honest framing is right: the timer is intrinsic on the no-deadline path and stays. Identity/deadline/no-op tests pass under `-race`.

PR #248, opened 2026-08-30
test(bench): add cross-core scaling row to the micro harness (#243)

### LINKED ISSUE #243: [PERF-5] Benchmark suite has no parallel/contention row, so scaling regressions are invisible

leo-aa88:
## Summary

The BENCH-1 micro suite ran every scenario single-goroutine, so it could not surface per-request serialization. The metrics-mutex scaling problem (#239) was invisible to `BenchmarkFrameworkTax` and only appeared under a parallel driver.

This adds `scenario.RunParallelBenchmark` (plaintext + valid-POST via `b.RunParallel`) and a `BenchmarkFrameworkTaxParallel` in all four rows (nethttp/gin/huma/gombit). Reading a row's per-op time across `-cpu` values shows whether it scales (ns/op falls with cores) or is contention-bound (ns/op flattens):

```
go test ./benchmarks/micro/... -bench=BenchmarkFrameworkTaxParallel -benchmem -cpu=1,2,4,8,16
```

Demonstration on this branch (Gombit still mutex-bound here, i.e. #239 not yet merged), plaintext ns/op:

| stack | -cpu=1 | -cpu=4 | -cpu=8 |
| --- | --- | --- | --- |
| gombit | 4401 | 2470 | 2825 (plateaus/regresses — contention-bound) |
| net-http (control) | 637 | — | 419 (keeps improving — no shared lock) |

Closes #243

## Acceptance criteria

- [x] Parallel benchmark exists in the shared harness and runs for all four rows.
- [x] Methodology doc (`benchmarks/docs/methodology.md`) explains what the parallel row measures and how to read it.
- [ ] Results/schema updated to carry the parallel numbers — **deliberately deferred** (see scope): the canonical snapshot must be regenerated on the dedicated benchmark host named in `metadata.json`, not a dev machine, or it would poison the README table with host-noisy numbers.

## Scope notes

Benchmark-infra only (labels align with #141: infra/tests/ci); no runtime change. The parallel row is a **diagnostic**, host- and scheduler-sensitive, intentionally kept out of the published README snapshot. Wiring the numbers into the generated results pipeline (`run-suite.sh`, `schema.md`) and a CI scaling-floor gate is a follow-up that belongs on the canonical host. This harness is what verifies #239/#240's scaling claims.

## Validation

- [x] `go test ./benchmarks/micro/... -run TestScenarios` — passes (all rows still correct)
- [x] `go vet ./benchmarks/micro/...` — clean
- [x] `BenchmarkFrameworkTaxParallel` runs for all four rows

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — N/A, no DB change (benchmark harness)
- [x] Stable features ship docs and appear in an example app — methodology doc updated; the "example" for a benchmark is the harness itself
- [x] Extraction preserves contracts — existing `BenchmarkFrameworkTax` and `TestScenarios` untouched

### DIFF
--- a/benchmarks/micro/scenario/parallel.go
+++ b/benchmarks/micro/scenario/parallel.go
@@ -1,0 +1,63 @@
+package scenario
+
+import (
+	"bytes"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// RunParallelBenchmark reports how a stack scales across CPU cores by driving
+// the plaintext and valid-POST scenarios concurrently with b.RunParallel, as
+// b.Run sub-benchmarks. It is the contention counterpart to RunBenchmark, whose
+// single-goroutine numbers hide any per-request serialization: a stack that
+// funnels every request through a shared lock barely improves here as
+// GOMAXPROCS rises, while a lock-free stack approaches linear scaling. That gap
+// is exactly what a single-goroutine microbench cannot surface (issue #243,
+// motivated by the metrics-mutex finding in #239).
+//
+// Read a row by comparing its per-op time across -cpu values, e.g.:
+//
+//	go test ./benchmarks/micro/... -bench=BenchmarkFrameworkTaxParallel -benchmem -cpu=1,2,4,8,16
+//
+// A row whose ns/op falls roughly in proportion to the core count scales;
+// a row whose ns/op flattens is contention-bound. The shared plaintext GET
+// request is safe to reuse across goroutines: none of the four stacks mutate
+// the inbound *http.Request in place (context changes go through
+// Request.WithContext, which returns a copy). POST cannot share a request — a
+// drained body can't be re-read — so each iteration builds its own.
+func RunParallelBenchmark(b *testing.B, stack Stack) {
+	b.Helper()
+
+	b.Run("plaintext", func(b *testing.B) {
+		request := httptest.NewRequest(http.MethodGet, "/plaintext", nil)
+		b.ReportAllocs()
+		b.ResetTimer()
+		b.RunParallel(func(pb *testing.PB) {
+			for pb.Next() {
+				response := httptest.NewRecorder()
+				stack.Handler.ServeHTTP(response, request)
+				if response.Code != http.StatusOK {
+					b.Fatalf("plaintext status = %d, want %d; body: %s", response.Code, http.StatusOK, response.Body.String())
+				}
+			}
+		})
+	})
+
+	b.Run("valid-post", func(b *testing.B) {
+		payload := []byte(ValidCreateUserBody)
+		b.ReportAllocs()
+		b.ResetTimer()
+		b.RunParallel(func(pb *testing.PB) {
+			for pb.Next() {
+				request := httptest.NewRequest(http.MethodPost, "/users", bytes.NewReader(payload))
+				request.Header.Set("Content-Type", "application/json")
+				response := httptest.NewRecorder()
+				stack.Handler.ServeHTTP(response, request)
+				if !statusOKOrCreated(response.Code) {
+					b.Fatalf("valid-post status = %d; body: %s", response.Code, response.Body.String())
+				}
+			}
+		})
+	})
+}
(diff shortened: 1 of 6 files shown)


benchmarks/micro/scenario/parallel.go
```diff
@@ -0,0 +1,63 @@
...
+	b.Run("plaintext", func(b *testing.B) {
+		request := httptest.NewRequest(http.MethodGet, "/plaintext", nil)
+		b.ReportAllocs()
+		b.ResetTimer()
+		b.RunParallel(func(pb *testing.PB) {
+			for pb.Next() {
+				response := httptest.NewRecorder()
+				stack.Handler.ServeHTTP(response, request)
+				if response.Code != http.StatusOK {
+					b.Fatalf("plaintext status = %d, want %d; body: %s", response.Code, http.StatusOK, response.Body.String())
```
leo-aa88: MINOR (non-blocking). `b.Fatalf` here — and at the valid-post site below — runs inside a goroutine spawned by `b.RunParallel`, not the goroutine running the benchmark function. `testing` documents that `FailNow`/`Fatal` must be called from the benchmark goroutine; from a `RunParallel` worker `Fatalf` calls `runtime.Goexit()` on *that* worker only, so the failure can race the other workers and the benchmark may not stop cleanly.

It only fires when a stack returns a non-2xx during a benchmark (a broken row), so the happy path is unaffected — hence non-blocking. But since these asserts exist precisely to catch a broken stack, prefer `b.Errorf(...)` + `return` (or record the bad status into a var and fail once after `RunParallel` returns) so the failure is reported deterministically.

leo-aa88 (commented):
COMMENT — one non-blocking note left inline.

Harness is correct and useful: verified the shared plaintext GET request is race-free to reuse — the gombit chain only mutates via `Request.WithContext` (a shallow copy) on that path; `c.Header(...)` writes the response and the XSS `Content-Length` write is POST-only, and POST builds a fresh request per iteration. `go vet` is clean and `BenchmarkFrameworkTaxParallel` runs for all four rows locally. Deferring the canonical snapshot to the dedicated host is the right call.

One minor inline finding on the `b.Fatalf` calls inside `RunParallel` workers (see comment) — non-blocking; it only affects the should-never-fire error path.

leo-aa88 (commented):
Re-reviewed against HEAD a6f2ea6. **LGTM** — the sole prior finding is resolved.

The new commit switches both `RunParallel` workers from `b.Fatalf` to `b.Errorf(...)` + `return`, with a comment naming the reason (`Fatal`/`FailNow` are only safe from the benchmark goroutine; `Errorf` marks the failure concurrency-safely, `return` stops the worker). That is exactly the fix, applied at both the plaintext and valid-post sites. `go vet ./benchmarks/micro/...` clean and `TestScenarios` passes for all four rows at HEAD.

Everything else is unchanged from the prior pass and stands: shared-GET reuse is race-free (mutation only via `Request.WithContext`), the methodology doc is accurate, and deferring the canonical snapshot to the dedicated host is correct. Nothing else to flag.

PR #252, opened 2026-08-31
fix: admin search is case-insensitive across SQLite/Postgres/MySQL

### LINKED ISSUE #200: [bug] Admin search uses plain LIKE, so results are case-sensitive on Postgres but not SQLite/MySQL
### What happened

The generic admin list search (`applySearch` in `admin/resources.go`) builds
its predicate with a plain SQL `LIKE`:

```go
ors = append(ors, clause.Expr{
    SQL:  quoteIdent(q, col) + " LIKE ? ESCAPE ?",
    Vars: []any{pattern, `\`},
})
```

`LIKE` case sensitivity is **driver-dependent**, so the same `?search=` query
returns different results depending on which database the app runs against:

- **PostgreSQL** — `LIKE` is case-**sensitive**. Searching `ada` does not match
  a row whose value is `Ada`.
- **SQLite** — `LIKE` is case-insensitive for ASCII only.
- **MySQL** — case sensitivity follows the column collation (default
  `utf8mb4_0900_ai_ci` / `..._general_ci` is case-**insensitive**).

So an admin search that "works" on a SQLite/MySQL dev box silently returns
fewer (or zero) results in a PostgreSQL deployment. This also contradicts the
Django-admin model Gombit is emulating, where `search_fields` uses
case-insensitive `icontains`.

### Steps to reproduce

```
1. gombit new demo --auth cookie
2. Register a model with Options.Search = []string{"title"} and seed a row
   with title "Ada".
3. GET /api/v1/admin/resources/<slug>?search=ada
4. On postgres: 0 results. On sqlite/mysql-ci-collation: 1 result.
```

### Impact

Admin search results differ across the supported SQLite + PostgreSQL + MySQL
matrix for identical data and query — the exact kind of cross-driver
divergence the working agreement (build plan §5) calls out. On PostgreSQL the
admin search box is effectively broken for any non-exact-case term.

### Suggested fix

Normalize case in the predicate rather than relying on `LIKE`: use `ILIKE` on
PostgreSQL, or `LOWER(col) LIKE LOWER(?)` uniformly (with the escape clause) so
all three drivers behave the same and match Django's case-insensitive search.

Reproduces on `main`.

Brunotlps:
## Summary

`applySearch` built its predicate with a plain SQL `LIKE`, whose case
sensitivity is driver-dependent: case-sensitive on PostgreSQL, case-
insensitive on SQLite/MySQL by default collation. The same `?search=`
query silently returned fewer/zero results on Postgres than on
SQLite/MySQL for the same data. Fixed by wrapping both sides of the
comparison in SQL `LOWER()` uniformly (no per-driver branching),
matching the cross-driver convention already used in
`database/errors.go`.

Closes #200

## Acceptance criteria

- [x] The generic admin list search (`Options.Search`) returns the same
      result set for the same query and data on SQLite, PostgreSQL, and
      MySQL — verified by running the fix against live Postgres 16 and
      MySQL 8.4 containers, not just the SQLite default test path.
- [x] No driver-specific branching introduced (`ILIKE`, `Dialector.Name()`,
      etc.) — the fix is a single uniform SQL predicate.

## Scope notes

Case-insensitivity here is ASCII-only, matching what SQLite's built-in
`LOWER()` can do without the ICU extension (not loaded by this
codebase). Postgres/MySQL fold full Unicode via their locale/charset
`LOWER()`, so accented/non-Latin search terms are not guaranteed
identical results across all three drivers — that gap is documented in
`docs/admin.md` and in the two regression-test comments rather than
silently implied. Closing it fully would mean either a per-driver
collation/extension story or reaching into Unicode-aware text handling,
which borders the M6 i18n battery and is out of scope for this issue
(the reported symptom, and the issue's own suggested fix, are ASCII
case only).

## Validation

- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go test -tags integration ./admin -admin.postgres-dsn '...'` (Postgres 16 container)
- [x] `go test -tags integration ./admin -admin.mysql-dsn '...'` (MySQL 8.4 container)
- [x] Project `code-review` skill run against this diff (one MAJOR finding — overclaiming "case-insensitive on every driver" without the ASCII qualifier — addressed by tightening the docs/comments wording; no code change needed)

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix
- [ ] Stable features ship docs and appear in an example app — N/A, this is a bugfix to existing documented behavior, not a new stable feature; `docs/admin.md`'s search description is updated in place instead
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR — N/A, no request/response shape change, internal query-building only
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep (jobs, events, scheduler, mail, storage, gRPC, multi-tenancy, i18n) — explicitly declined to chase full Unicode parity for this reason, see Scope notes

https://claude.ai/code/session_019TdyJhLKLe6k26JEi9i6VZ

### DIFF
--- a/admin/integration_test.go
+++ b/admin/integration_test.go
@@ -60,6 +60,23 @@ func runResourceDriver(t *testing.T, driver config.DatabaseDriver, dsn string) {
 	if list.Code != http.StatusOK {
 		t.Fatalf("list status = %d; body: %s", list.Code, list.Body.String())
 	}
+
+	// Search must be ASCII case-insensitive on every driver, matching
+	// SQLite's default behavior — Postgres's plain LIKE is case-sensitive,
+	// which was issue #200. (ASCII only: SQLite's LOWER() doesn't fold
+	// non-ASCII case, so this doesn't prove Unicode parity with Postgres/
+	// MySQL's locale-aware LOWER().)
+	caseInsensitive := doRequest(app, jar, http.MethodGet, "/api/v1/admin/resources/widgets?search=bolt", "")
+	if caseInsensitive.Code != http.StatusOK {
+		t.Fatalf("case-insensitive search status = %d; body: %s", caseInsensitive.Code, caseInsensitive.Body.String())
+	}
+	var foundLower listEnvelope
+	if err := json.Unmarshal(caseInsensitive.Body.Bytes(), &foundLower); err != nil {
+		t.Fatalf("decode case-insensitive search: %v", err)
+	}
+	if foundLower.Meta == nil || foundLower.Meta.Total != 1 || len(foundLower.Data) != 1 || foundLower.Data[0]["name"] != "Bolt" {
+		t.Fatalf("case-insensitive search = %+v", foundLower)
+	}
 	del := doRequest(app, jar, http.MethodDelete, "/api/v1/admin/resources/widgets/"+id, "")
 	if del.Code != http.StatusOK {
 		t.Fatalf("delete status = %d; body: %s", del.Code, del.Body.String())
--- a/admin/resources.go
+++ b/admin/resources.go
@@ -485,7 +485,7 @@ func applySearch(q *gorm.DB, m *registered, term string) (*gorm.DB, error) {
 			continue
 		}
 		ors = append(ors, clause.Expr{
-			SQL:  quoteIdent(q, col) + " LIKE ? ESCAPE ?",
+			SQL:  "LOWER(" + quoteIdent(q, col) + ") LIKE LOWER(?) ESCAPE ?",
 			Vars: []any{pattern, `\`},
 		})
 	}
--- a/admin/resources_test.go
+++ b/admin/resources_test.go
@@ -515,6 +515,19 @@ func TestResourceListPaginationSearchOrderFilter(t *testing.T) {
 		t.Fatalf("search = %+v", found)
 	}
 
+	// Search is ASCII case-insensitive regardless of driver collation — see
+	// issue #200. (Not a full Django icontains equivalent: SQLite's LOWER()
+	// only folds ASCII, so Unicode case-folding parity across drivers isn't
+	// guaranteed for accented/non-Latin terms.)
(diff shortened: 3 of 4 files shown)


leo-aa88 (approved):
LGTM.

Traced end-to-end and the fix is correct, minimal, and honestly scoped.

- `LOWER(col) LIKE LOWER(?) ESCAPE ?` — `LOWER()` on the pattern only folds letters, so `escapeLike`'s escaped `%`/`_`/`\` and the `ESCAPE '\'` char all survive; wildcards and escaping still behave.
- No index regression: the leading `%` already precluded a b-tree index, so wrapping the column in `LOWER()` costs nothing that plain `LIKE '%term%'` didn't already cost. `LOWER(NULL)`→NULL→no match, same as before.
- AC2 satisfied: single uniform predicate, no `ILIKE`/`Dialector.Name()` branching.
- AC1 genuinely guarded: the `?search=bolt`→`Bolt` assertion in `runResourceDriver` runs under `go test -tags integration ./admin` against live Postgres 16 + MySQL 8.4 in `ci.yml` per-PR. On old code Postgres's case-sensitive `LIKE` fails that assertion; the fix makes it pass. That is the real regression guard.

Two minor, non-blocking notes:

1. `admin/resources_test.go` — the added SQLite unit assertion (`?search=beta`→`Beta`) does not actually guard this fix. SQLite's `LIKE` is ASCII-case-insensitive by default, so it passes with *or* without the `LOWER()` change. The comment already concedes this ("matching SQLite's default behavior"), and the integration test is the true guard — so this is fine as documentation of intent, just be aware `go test ./...` alone would stay green on a revert.

2. PR description nit (no code impact): "matching the cross-driver convention already used in `database/errors.go`" is a stretch — that file uses Go `strings.ToLower` on driver *error strings*, not SQL `LOWER()`. Different mechanism. Doesn't affect the diff; just don't lean on it as precedent.

Docs updated with an accurate ASCII-only caveat, scope correctly declines the Unicode/i18n rabbit hole. Ship it.

leo-aa88 (approved):

PR #253, opened 2026-08-31
fix: memory cache reclaims expired entries never re-read via Get

### LINKED ISSUE #199: [bug] In-memory cache never evicts expired entries (unbounded growth for high-cardinality keys)
### What happened

The in-process `cache.Memory` driver (`cache/memory.go`) never removes expired
entries except when the **same key** is read again via `Get`. There is no
background janitor and no sweep on `Set`/`Increment`/`Delete` of other keys.

For any workload that writes many short-TTL keys that are rarely (or never)
read back with the same key — rate limiting (`key = "rate:" + ip`), one-time
tokens, per-session nonces, idempotency keys — the map grows without bound.
Each `Set` inserts a new `memoryItem` and the expired ones are only deleted on
a subsequent `Get(sameKey)` that happens to land after expiry. Keys that are
never fetched again (the common case for the workloads above) stay resident
forever, so RSS grows for the life of the process. Redis (`cache/redis.go`)
expires keys server-side, so the two drivers behave very differently for the
same code — a memory-driver dev/test env looks fine while the leak only shows
up under sustained traffic.

```go
// cache/memory.go — Set never evicts other expired keys
m.mu.Lock()
m.items[key] = memoryItem{payload: payload, expiresAt: expiresAt}
m.mu.Unlock()
// Get is the only place a delete can happen, and only for the key being read.
```

### Steps to reproduce

```
1. cache := cache.NewMemory()
2. for i := 0; i < 1_000_000; i++ {
      _ = cache.Set(ctx, fmt.Sprintf("rate:%d", i), 1, time.Second)
   }
3. time.Sleep(2 * time.Second) // everything is now logically expired
4. // len(cache.items) is still 1_000_000; nothing was reclaimed.
   // runtime.ReadMemStats shows the map still pinned in the heap.
```

### Impact

Unbounded memory growth for high-cardinality / short-TTL keys when the
default `memory` cache driver is used (the default in `config.DefaultFor`).
This is a slow-burn production leak that will not surface in tests.

### Suggested fix

Add a background reaper (opt-in `time.Ticker` goroutine started by
`NewMemory`, stopped on a `Close`), or sweep a bounded number of expired keys
on write. Either way the memory driver needs a lifecycle hook so the goroutine
can be shut down cleanly (and so `cache.Store.Close` can stop it).

Reproduces on `main`.

Brunotlps:
## Summary

`cache.Memory` only ever removed an expired entry when that same key
was read again via `Get`. Write-heavy, high-cardinality, rarely-re-read
keys (rate limit counters, one-time tokens, per-session nonces) grew
the map without bound for the life of the process; `cache.Redis`
doesn't have this problem since Redis expires keys server-side.

Added an opt-in background janitor (`WithJanitor`) that periodically
sweeps expired entries, wired only through `cache.Open`'s memory driver
case so `NewMemory()` with no arguments keeps its current
zero-goroutine behavior for every existing direct caller. `cache.Open`
reuses the existing `Store.close` lifecycle hook (already wired for
Redis), so `framework.App` closes it on shutdown with no framework
changes beyond documenting that a cache attached via `framework.WithCache`
is not closed by the App.

The sweep snapshots keys under a read lock, then deletes expired ones
in bounded batches under short write-lock acquisitions, so a large map
is never held under a single lock for the length of a full scan — the
same reason Redis's own active-expire cycle samples in batches rather
than scanning its keyspace in one shot.

Closes #199

## Acceptance criteria

- [x] A key that is set and never read again is still reclaimed once
      expired (not just on a same-key `Get`) — proven deterministically
      with a fake clock, no reliance on timing.
- [x] `Set`/`Increment` stay O(1); no full-map scan was added to the
      write hot path.
- [x] The background sweep never holds a single lock for the whole map
      — proven by a batch-count test on a map larger than the batch size.
- [x] `NewMemory()` with no arguments has no behavior change (no
      goroutine) for existing callers.

## Scope notes

The janitor interval and sweep batch size are fixed constants
(`memoryJanitorInterval`, `memorySweepBatchSize`), not new
`config.CacheConfig` fields — this is a correctness fix, not a new
tunable. `cache.Redis` and `cache.Noop` are unaffected; they don't have
this bug.

## Validation

- [x] `go build ./...`
- [x] `go test ./cache/... -race -v`
- [x] `go test ./... -race` (full repo)
- [x] Project `code-review` skill run twice against this diff: first
      round flagged two MAJOR findings (a full-map scan holding the
      write lock for the whole sweep, and `framework.WithCache`
      silently not closing an injected cache with a running janitor);
      both addressed (bounded-batch sweep; documented the ownership
      contract), second round approved

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — N/A, no database touched
- [ ] Stable features ship docs and appear in an example app — N/A, this is a bugfix to existing driver behavior, not a new feature
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR — N/A, no request/response shape change
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep (jobs, events, scheduler, mail, storage, gRPC, multi-tenancy, i18n) — fixed janitor interval instead of a new config knob, per Scope notes

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -53,7 +53,9 @@ func Open(cfg config.CacheConfig) (*Store, error) {
 
 	switch driver := Driver(cfg.Driver); driver {
 	case DriverMemory:
-		impl = NewMemory()
+		mem := NewMemory(WithJanitor(memoryJanitorInterval))
+		impl = mem
+		closeFn = mem.Close
 	case DriverRedis:
 		var err error
 		client, err = NewRedisClient(cfg.Redis)
--- a/cache/cache_test.go
+++ b/cache/cache_test.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


leo-aa88 (approved):
LGTM.

Traced the full path — sweep snapshots keys under `RLock`, deletes in bounded batches under short `Lock` acquisitions, and re-checks `expired(now)` against the live map inside `sweepBatch` so a key `Set` again after the snapshot is never wrongly evicted. `Close` is `sync.Once`-guarded and no-ops when no janitor was started, so the ~10 bare `NewMemory()` call sites keep their zero-goroutine behavior. The `cache.Open` memory case wires `closeFn = mem.Close` and `closeOwnedCache` only closes app-owned stores, and you documented on `WithCache` that an injected cache (janitor and all) stays the caller's to close — the one real ownership footgun this introduces, correctly called out rather than hidden.

Tests earn their place: `sweepExpired` proven deterministically with a fake clock, the batch-count test (25+1 keys, batch 10 → 3 batches) proves no single lock hold scales with map size, and the janitor test exercises the real-timing wiring with a bounded poll. Ran `go test ./cache/ -race` locally — green.

The `RLock` snapshot is still O(n) allocation + iteration under the read lock once per interval, and the fixed 1-minute period means a burst of short-TTL keys can accumulate for up to a minute between sweeps — but that's bounded growth replacing unbounded growth, which is exactly the contract #199 asks for, and the tradeoff is documented in the sweep comment. No change requested.

PR #254, opened 2026-08-31
fix: embedded frontend serves HEAD like GET on SPA routes and index.html

### LINKED ISSUE #251: [bug] Embedded frontend returns 404 for HEAD / while GET / serves the SPA (health checks with HEAD fail)
### What happened

With an embedded frontend (`gombit build --embed`), `HEAD /` returns **404**
while `GET /` returns **200** with `index.html`. The same 404 hits `HEAD` on
any SPA client-side route (e.g. `HEAD /products`). RFC 9110 requires HEAD to
behave like GET minus the body, and this breaks anything that probes with
HEAD: load-balancer health checks, uptime monitors, link checkers, `curl -I`.

Cause: `embeddedFrontendHandler` (`framework/embed.go:48`) admits GET **and**
HEAD, serves real files for both, but restricts the SPA `index.html` fallback
to GET:

```go
name := strings.TrimPrefix(urlPath, "/")      // "" for "/"
if name == "index.html" { serveIndexHTML(...); return }
if name != "" && name != "." && fs.ValidPath(name) {
    if serveEmbeddedFile(c, fsys, name) { return }
}
if c.Request.Method != http.MethodGet {       // HEAD lands here
    c.AbortWithStatus(http.StatusNotFound)
    return
}
serveIndexHTML(c, fsys, apiPrefix)
```

`/` produces `name == ""`, skips the file branch, and hits the `!= GET` 404 —
so the root document itself is HEAD-unreachable. Oddly, `HEAD /index.html`
works (200), and `HEAD /assets/app.js` works, so the behavior is internally
inconsistent too.

The framework's own admin SPA already gets this right: `adminSPAHandler`
(`framework/adminui.go`) serves `HEAD /admin` and `HEAD /admin/<route>` with
200 + correct headers via `writeBytes`, and `internal adminui tests assert it`
(`framework/adminui_test.go:119`). The embedded-frontend handler should match.

Expected: `HEAD /` (and HEAD on SPA fallback routes) returns the same status
and headers as GET, with no body.

### Steps to reproduce

```shell
1. gombit new demo && cd demo
2. gombit build --embed
3. ./bin/server &
4. curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080/        # 200
5. curl -s -o /dev/null -w '%{http_code}\n' -I http://127.0.0.1:8080/     # 404  <- bug
6. curl -s -o /dev/null -w '%{http_code}\n' -I http://127.0.0.1:8080/index.html  # 200 (inconsistent)
```

### gombit version

### Environment

```text
go 1.x, linux
app built with gombit build --embed (framework.WithEmbeddedFrontend)
```

### Database

not database related

### Auth mode

not auth related

### Area

runtime (framework.App, routing, lifecycle)

### Logs or output

```text
$ curl -I http://127.0.0.1:8080/
HTTP/1.1 404 Not Found

$ curl http://127.0.0.1:8080/ -s -o /dev/null -w '%{http_code}\n'
200
```

### Suggested fix

Allow HEAD through the SPA fallback in `embeddedFrontendHandler` (change the
late `!= GET` guard to also accept HEAD, as the first guard already does), and
serve `index.html` the way `adminui.go`'s `writeBytes` does so HEAD gets
status + Content-Type/Content-Length without a body. The net/http server
already suppresses the body for HEAD, so the minimal fix is dropping the
GET-only guard.

### Before submitting

Brunotlps:
## Summary

`embeddedFrontendHandler` blocked HEAD with a redundant guard that ran
after the file-serving attempts, 404ing `HEAD /` and any unmatched
client-side route while `GET` served the SPA fine. The guard was dead
weight: given the handler's own entry guard, only GET or HEAD can ever
reach that point.

Removing it exposed a second, deeper problem the issue itself didn't
catch: `serveIndexHTML` and `serveEmbeddedFile`'s non-seekable fallback
wrote the response body via `c.Data` even for HEAD, so `/index.html`
"working" (200 status) still leaked a body that `curl -I` can't reveal
(libcurl discards a HEAD body silently). Both now route through the
`writeBytes` helper `adminui.go` already uses for the same purpose,
which explicitly skips the body (and sets `Content-Length` correctly)
on HEAD.

Closes #251

## Acceptance criteria

- [x] `HEAD /` (and any unmatched SPA route) returns the same status
      and headers as `GET`, with no body — verified against `/`,
      `/login`, `/products/new`.
- [x] `HEAD /index.html` returns the same, closing a body-suppression
      gap the issue's own `curl -I` repro couldn't detect.
- [x] `HEAD` on a real embedded asset (`/assets/app.js`) keeps working
      as before.
- [x] `Content-Length` on a HEAD response matches what the equivalent
      GET would send (RFC 9110), asserted against the live GET response
      rather than a hardcoded value.

## Scope notes

None — this is a self-contained runtime fix, no generator/contract/DB
surface touched.

## Validation

- [x] `go build ./...`
- [x] `go test ./framework/... -race -v`
- [x] `go test ./...` (full repo, matches CI baseline)
- [x] Project `code-review` skill run against this diff: one finding
      (the regression test didn't assert `Content-Length` on HEAD,
      the most specific RFC 9110 guarantee the issue names) — addressed
      by comparing HEAD's `Content-Length` against the equivalent GET
      response's actual body length

Note: `go test ./... -race` surfaces a pre-existing, unrelated data race
in `dev/supervisor.go` (`TestRunChildEnvReplacesParentHTTPAddr`,
concurrent stdout/stderr pipe copying between two spawned child
processes). Confirmed present on `main` before this branch — out of
scope here, flagging separately.

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — N/A, no database touched
- [ ] Stable features ship docs and appear in an example app — N/A, bugfix restoring documented HEAD/GET parity, not a new feature
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR — N/A, no request/response shape change; these routes stay out of OpenAPI (raw Gin escape hatch, unchanged)
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep (jobs, events, scheduler, mail, storage, gRPC, multi-tenancy, i18n)

### DIFF
--- a/framework/embed.go
+++ b/framework/embed.go
@@ -69,10 +69,9 @@ func embeddedFrontendHandler(fsys fs.FS, apiPrefix string) gin.HandlerFunc {
 			}
 		}
 
-		if c.Request.Method != http.MethodGet {
-			c.AbortWithStatus(http.StatusNotFound)
-			return
-		}
+		// SPA fallback: only GET and HEAD reach this handler at all (the
+		// guard above), and serveIndexHTML writes via writeBytes, which
+		// already skips the body for HEAD — so no method check belongs here.
 		serveIndexHTML(c, fsys, apiPrefix)
 	}
 }
@@ -139,7 +138,7 @@ func serveEmbeddedFile(c *gin.Context, fsys fs.FS, name string) bool {
 	if err != nil {
 		return false
 	}
-	c.Data(http.StatusOK, contentTypeFor(name), data)
+	writeBytes(c, contentTypeFor(name), data)
 	return true
 }
 
@@ -151,7 +150,7 @@ func serveIndexHTML(c *gin.Context, fsys fs.FS, apiPrefix string) {
 	}
 	data = injectAPIPrefixHTML(data, apiPrefix)
 	applySPAContentSecurityPolicy(c)
-	c.Data(http.StatusOK, "text/html; charset=utf-8", data)
+	writeBytes(c, "text/html; charset=utf-8", data)
 }
 
 func contentTypeFor(name string) string {
--- a/framework/embed_test.go
+++ b/framework/embed_test.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
 	"strings"
 	"testing"
 	"testing/fstest"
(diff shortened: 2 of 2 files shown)


framework/embed.go
```diff
@@ -139,7 +138,7 @@ func serveEmbeddedFile(c *gin.Context, fsys fs.FS, name string) bool {
 	if err != nil {
 		return false
 	}
-	c.Data(http.StatusOK, contentTypeFor(name), data)
+	writeBytes(c, contentTypeFor(name), data)
```
leo-aa88, line 141: MINOR / non-blocking. This `writeBytes` swap on the non-seekable branch is a correct defensive improvement, but note it's effectively unreachable in both production and the tests: `embed.FS` files implement `io.ReadSeeker`, so real embedded assets always take the `http.ServeContent` path above; `fstest.MapFS` (the test fixture) is seekable too, so `TestEmbeddedFrontendHeadMatchesGet`'s `/assets/app.js` case exercises `http.ServeContent`, not this line. So the HEAD-body suppression you're adding here has no test that actually drives it, and won't get one from any FS the framework hands this function today. Fine to keep for correctness, but worth a comment noting the branch is only hit for a non-seekable `fs.FS` an app might supply directly — otherwise the next reader will assume it's covered.

leo-aa88 (approved):
LGTM.

The core fix is correct: the removed guard was dead weight that 404'd HEAD on every SPA-fallback route, and the entry guard at the top already restricts the handler to GET/HEAD, so routing HEAD through `serveIndexHTML` is right. Switching `serveIndexHTML` and the non-seekable `serveEmbeddedFile` branch to `writeBytes` also fixes the real RFC 9110 gap — `c.Data` on a HEAD request over a real `net/http.Server` yields 200 with no `Content-Length` (the server discards the body it would have measured), whereas `writeBytes` sets `Content-Length` explicitly and omits the body. The seekable path stays on `http.ServeContent`, which already handles HEAD.

Good call asserting HEAD's `Content-Length` against the live GET body length rather than a constant, and being explicit that `serveEmbed` drives the handler against a `ResponseRecorder` (which doesn't strip HEAD bodies) — that's what lets the empty-body assertion actually mean something. Ran the regression test locally — green.

One genuinely minor, non-blocking note inline.

PR #256, opened 2026-09-01
fix: eliminate two data races in dev package process-supervision tests

### LINKED ISSUE #255: [bug] Data race in dev package tests: two mocked child processes share one bytes.Buffer as Stdout
## What happened

`go test ./... -race` fails on `TestRunChildEnvReplacesParentHTTPAddr`
(`dev/run_test.go`) with a real, reproducible data race — not flaky, it
reproduced on every run I tried, including on a clean `main` checkout
before any of my changes.

The test does:

```go
// dev/run_test.go:497
stdout := new(bytes.Buffer)
```

and passes that single, unsynchronized `*bytes.Buffer` as `Options.Stdout`
for both of the two mocked child processes the test spawns (standing in
for the Go backend + Vite frontend `gombit dev` supervises). Neither
`ProcSpec` sets its own `cmd.Stdout`, so `runProcesses`'s fallback
(`dev/supervisor.go`: `if cmd.Stdout == nil { cmd.Stdout = stdout }`)
assigns the *same* buffer instance to both `exec.Cmd`s.

When `cmd.Stdout` is a plain `io.Writer` (not an `*os.File`), Go's
`os/exec` starts an internal goroutine per process that does
`io.Copy(cmd.Stdout, pipe)` for the life of that process. With two child
processes sharing one buffer, two such goroutines end up calling
`bytes.Buffer.Write`/`grow` concurrently on the exact same buffer, which
isn't safe for concurrent use — hence the race.

## Does this affect the real `gombit dev` binary?

No, as far as I can trace. `cmd/gombit/main.go` passes `os.Stdout` /
`os.Stderr` (real `*os.File`s) all the way down through
`cli.Execute` → `newDevCommand` → `dev.Options.Stdout/Stderr` →
`runProcesses`. When `cmd.Stdout` is literally an `*os.File`, `os/exec`
skips the copying-goroutine path entirely and wires the child's stdout
file descriptor directly to that file (dup2), so there is no in-process
buffer for two goroutines to race on. Two real child processes writing to
the same real `os.Stdout` file descriptor is the ordinary, safe way every
terminal multiplexes concurrent output.

So this looks confined to this test's setup (`new(bytes.Buffer)` shared,
unguarded, across two concurrently-running mocked processes) rather than
a defect reachable through the shipped `gombit dev` command. Flagging it
as-is rather than guessing at scope — happy to be corrected if I'm
missing a path where a non-file Stdout is actually shared across
processes in production.

## Steps to reproduce

```shell
git clone https://github.com/gombit-dev/gombit.git
cd gombit
go test ./dev/... -race -run TestRunChildEnvReplacesParentHTTPAddr -count=3
# -> WARNING: DATA RACE ... bytes.(*Buffer).grow() / ReadFrom(), two
#    goroutines both created at os/exec.(*Cmd).Start() from
#    dev.runProcesses (dev/supervisor.go:55), called from
#    dev.TestRunChildEnvReplacesParentHTTPAddr (dev/run_test.go:535)
# --- FAIL: TestRunChildEnvReplacesParentHTTPAddr
```

Reproduced identically on a clean `main` checkout with no other changes
applied (confirmed via `git stash` before running), so this isn't
something introduced by an in-flight branch.

## gombit version

dev (built from a `main` checkout, 2346841)

## Environment

```text
gombit:   dev
commit:   2346841
go:       go1.25.7
platform: linux/amd64
```

## Database

- not database related

## Auth mode

- not auth related

## Area

- CLI / generators

## Logs or output

```text
==================
WARNING: DATA RACE
Read at 0x00c0005825b8 by goroutine 49:
  bytes.(*Buffer).Len()
  bytes.(*Buffer).grow()
  bytes.(*Buffer).ReadFrom()
  io.copyBuffer()
  io.Copy()
  os.genericWriteTo()
  os.(*File).WriteTo()
  io.copyBuffer()
  io.Copy()
  os/exec.(*Cmd).writerDescriptor.func1()
  os/exec.(*Cmd).Start.func2()
  os/exec.(*Cmd).Start.gowrap1()

Previous write at 0x00c0005825b8 by goroutine 46:
  bytes.(*Buffer).grow()
  bytes.(*Buffer).ReadFrom()
  io.copyBuffer()
  io.Copy()
  os.genericWriteTo()
  os.(*File).WriteTo()
  io.copyBuffer()
  io.Copy()
  os/exec.(*Cmd).writerDescriptor.func1()
  os/exec.(*Cmd).Start.func2()
  os/exec.(*Cmd).Start.gowrap1()

Goroutine 49 (running) created at:
  os/exec.(*Cmd).Start()
  github.com/gombit-dev/gombit/dev.runProcesses()
      dev/supervisor.go:55
  github.com/gombit-dev/gombit/dev.Run()
      dev/run.go:117
  github.com/gombit-dev/gombit/dev.TestRunChildEnvReplacesParentHTTPAddr.func4()
      dev/run_test.go:535

Goroutine 46 (running) created at:
  os/exec.(*Cmd).Start()
  github.com/gombit-dev/gombit/dev.runProcesses()
      dev/supervisor.go:55
  github.com/gombit-dev/gombit/dev.Run()
      dev/run.go:117
  github.com/gombit-dev/gombit/dev.TestRunChildEnvReplacesParentHTTPAddr.func4()
      dev/run_test.go:535
==================
--- FAIL: TestRunChildEnvReplacesParentHTTPAddr (0.00s)
    testing.go:1712: race detected during execution of test
FAIL
FAIL	github.com/gombit-dev/gombit/dev	0.126s
```

## Suggested fix

Give each mocked `ProcSpec` in the test its own buffer (or wrap the
shared one in a small `io.Writer` that serializes access with a mutex),
the same way real usage never shares a non-file `io.Writer` across two
concurrently-running child processes. Since this looks test-only, it's
also possible the maintainer would rather just confirm the scope and
decide whether `-race` should be part of the standard CI gate for the
`dev` package at all — noting it here rather than assuming.

## Before submitting

Brunotlps:
## Summary

`go test ./dev/... -race` failed on `TestRunChildEnvReplacesParentHTTPAddr`
(reliably, every run) because the test shared one `*bytes.Buffer` as
`Options.Stdout` across two concurrently-started mocked child processes.
`runProcesses` (`dev/supervisor.go`) fans a single writer out to every
`ProcSpec` whose own `cmd.Stdout` is nil; when that writer isn't an
`*os.File`, `os/exec` spins up a copy goroutine per child, and two such
goroutines writing to the same unsynchronized buffer is exactly what `-race`
flagged. The real `gombit dev` binary is unaffected: `cmd/gombit/main.go`
always passes `os.Stdout`/`os.Stderr` (real files) all the way down, and
`os/exec` wires those directly via dup2 with no copy goroutine.

`TestRunPrintsServiceTableAndShutsDown` had the identical construction, plus
a second hazard: it read `stdout.String()`/`stderr.String()` while those
copy goroutines could still be writing.

Fixing the buffer race exposed a **second, previously-masked race** in
`TestRunChildEnvReplacesParentHTTPAddr`: `runProcesses` assigns `cmd.Env` on
its own goroutine with no synchronization, and the existing
`waitChildEnvAssigned` test helper polled `cmd.Env` from the test goroutine
with no memory barrier connecting the two — a genuine unsynchronized access
regardless of whether the polled value was ever actually wrong. This
reproduces on unmodified `main` too (hit in 1 of 15 `-race` runs), just
normally masked by the louder, near-always-reproducible buffer race
finishing first.

## Fix

- `TestRunChildEnvReplacesParentHTTPAddr`: the shared buffer was never read
  (only ever assigned, never inspected) — dropped it for `ioDiscard{}`.
- `TestRunPrintsServiceTableAndShutsDown`: content *is* asserted, so its
  shared writers became a small mutex-guarded `syncBuffer` instead.
- The Env race: added an unexported `onCmdReady func(*exec.Cmd)` test seam to
  `runProcesses`/`Options`, invoked right after `cmd.Env` is assigned and
  before `Start`. The test now blocks on a channel receive instead of
  polling the field, which gives a real happens-before edge under the Go
  memory model. `onCmdReady` is `nil` (a no-op) on every real `gombit dev`
  invocation — `dev/run.go` is the only caller of `runProcesses` outside
  tests, and `Options` is built in a different package (`cli`) that can't
  set an unexported field.

No behavior change to the shipped `gombit dev` command.

Closes #255

## Acceptance criteria

The linked issue is a bug report without a formal AC checklist; treating its
"Suggested fix" / reproduction steps as the bar:

- [x] `go test ./dev/... -race -run TestRunChildEnvReplacesParentHTTPAddr -count=3` passes (was reliably failing before)
- [x] `go test ./dev/... -race -count=30` (targeted at all three touched tests) passes with zero races
- [x] Confirmed via code trace that `cmd/gombit/main.go` → `cli.NewRoot` → `newDevCommand` → `dev.Options` always passes a real `*os.File`, so this was never reachable through the shipped binary
- [x] The second (Env-field) race, not identified in the original issue, is fixed with real synchronization rather than left as residual flakiness

## Scope notes

Test-infrastructure fix plus one small internal (unexported) seam in
`dev/supervisor.go`/`dev/options.go`. No public API, HTTP contract,
generator, database, or auth surface touched.

## Validation

- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go test ./dev/... -race -count=30` (targeted stress run on the three previously-racy tests)
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run` (0 issues)
- [x] Project `code-review` skill run against this diff — APPROVE, two non-blocking MINOR notes (doc-comment placement on `runProcesses`; a pre-existing hardcoded channel-capacity convention in the test file, unchanged by this PR)

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — N/A, no DB touched; the "new behavior" here is the race-free test synchronization itself, exercised by a 30x `-race` stress run
- [ ] Stable features ship docs and appear in an example app — N/A, internal test-infra fix, not a user-facing feature
- [x] Extraction preserves contracts — refactor and move, don't rewrite code that already passes its tests — assertions in both touched tests are unchanged; only the synchronization mechanism was replaced
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep — confirmed, test-infra only

### DIFF
--- a/dev/options.go
+++ b/dev/options.go
@@ -60,6 +60,9 @@ type Options struct {
 	// AdminURL is printed in the service table when non-empty (cookie-mode
 	// apps that mount the framework admin SPA). JWT apps leave this empty.
 	AdminURL string
+	// onCmdReady is an in-package test seam; see runProcesses. Always nil
+	// outside dev's own tests.
+	onCmdReady func(*exec.Cmd)
 }
 
 func (opts *Options) normalize() error {
--- a/dev/run.go
+++ b/dev/run.go
@@ -114,7 +114,7 @@ func Run(ctx context.Context, opts Options) error {
 		_ = watchOpenAPI(runCtx, opts, services.Backend+"/openapi.json")
 	}()
 
-	err = runProcesses(runCtx, procs, opts.Stdout, opts.Stderr, opts.Command, opts.ShutdownWait)
+	err = runProcesses(runCtx, procs, opts.Stdout, opts.Stderr, opts.Command, opts.ShutdownWait, opts.onCmdReady)
 	cancel()
 	<-watchDone
 	if err != nil {
--- a/dev/run_test.go
+++ b/dev/run_test.go
@@ -54,8 +54,8 @@ func TestRunPrintsServiceTableAndShutsDown(t *testing.T) {
 	}
 
 	workDir := writeDevApp(t)
-	stdout := new(bytes.Buffer)
-	stderr := new(bytes.Buffer)
+	stdout := &syncBuffer{}
+	stderr := &syncBuffer{}
 	var started atomic.Int32
 	ctx, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)
@@ -150,7 +150,7 @@ func TestRunProcessesShutdownOnCancel(t *testing.T) {
 
 	errCh := make(chan error, 1)
 	go func() {
-		errCh <- runProcesses(ctx, specs, ioDiscard{}, ioDiscard{}, command, 2*time.Second)
+		errCh <- runProcesses(ctx, specs, ioDiscard{}, ioDiscard{}, command, 2*time.Second, nil)
 	}()
 
 	waitStarts(t, startedCh, 2)
@@ -172,6 +172,29 @@ type ioDiscard struct{}
 
 func (ioDiscard) Write(p []byte) (int, error) {
 	return len(p), nil
+}
+
+// syncBuffer is a bytes.Buffer safe for concurrent Write/String. Real
+// `gombit dev` always shares its Stdout/Stderr *os.File across child
+// processes (os/exec wires those directly, no in-process copy goroutine),
+// but a mocked test process makes os/exec spin up a copy goroutine per
+// child, so an in-memory sink shared across children (and read from the
+// test goroutine while they run) needs its own locking.
+type syncBuffer struct {
+	mu  sync.Mutex
+	buf bytes.Buffer
+}
+
+func (b *syncBuffer) Write(p []byte) (int, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.Write(p)
+}
+
+func (b *syncBuffer) String() string {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.String()
 }
 
 func writeDevApp(t *testing.T) string {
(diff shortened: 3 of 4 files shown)


leo-aa88 (approved):
LGTM

PR #257, opened 2026-09-01
fix: gombit db rollback runs down files statement-by-statement

### LINKED ISSUE #249: [bug] gombit db rollback runs each down file as one Exec — multi-statement down files fail on MySQL
### What happened

`gombit db rollback` executes each companion down file as **one**
`db.Exec(wholeFile)` call (`migrations/rollback.go:103`), while `gombit db seed`
carefully splits files into individual statements first
(`splitSQLStatements`, `migrations/seed.go`). On MySQL that difference is
fatal: `go-sql-driver/mysql` rejects multi-statement SQL unless the DSN opts in
with `multiStatements=true`, and no gombit-generated or documented MySQL DSN
does (`scaffold/generate.go` and the docs use
`USER:PASSWORD@tcp(127.0.0.1:3306)/app?parseTime=true`).

So on MySQL, any down file containing more than one statement — the normal
case for a migration that created a table plus an index, or two tables —
fails with a MySQL 1064 syntax error at the first `;`, and the batch cannot be
rolled back. Because rollback on MySQL runs without a transaction, a down file
whose *first* statement succeeds before a later file fails can also leave the
schema partially rolled back (the documented MySQL caveat), but here even a
fully valid down script fails on its own.

SQLite (mattn driver executes statements sequentially) and PostgreSQL (pgx
falls back to the simple protocol for no-arg Exec, which allows multiple
commands) are unaffected — which is why the multi-DB integration tests pass:
`migrations/integration_test.go` only ever uses a **single-statement** down
file (`DROP TABLE IF EXISTS widgets;`).

Expected: rollback executes down files statement-by-statement the same way
seed does (`splitSQLStatements`), so a documented multi-statement down file
works on all three supported drivers.

### Steps to reproduce

```shell
1. gombit new demo --database mysql   # DSN: USER:PASSWORD@tcp(...)/demo?parseTime=true (no multiStatements)
2. cd demo && edit .env with real MySQL credentials
3. gombit db makemigrations create_widgets --model internal/widget.Widget   # up creates widgets table + an index
4. gombit db migrate
5. Write the companion down file database/migrations/downs/<version>_create_widgets.down.sql:
     DROP INDEX idx_widgets_name ON widgets;
     DROP TABLE widgets;
6. gombit db rollback
   # => migrations: execute <version>_create_widgets.down.sql: Error 1064 ... near 'DROP TABLE widgets'
   # framework_migrations is unchanged, but on MySQL the first statement may already have run.
```

### gombit version

### Environment

```text
go 1.x, linux
MySQL 8 via docker, DSN gombit:gombit@tcp(127.0.0.1:3306)/gombit?parseTime=true
```

### Database

mysql

### Auth mode

not auth related

### Area

migrations (Atlas), CLI / generators

### Logs or output

```text
migrations: execute 20260101000000_create_widgets.down.sql: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DROP TABLE widgets' at line 2; framework_migrations unchanged
```

### Suggested fix

Run each down file through the same `splitSQLStatements` used by
`migrations/seed.go` and execute the statements in order (inside the existing
transaction on SQLite/Postgres, sequentially on MySQL). That also improves the
MySQL partial-failure error message, which can then name the exact statement
that failed rather than the whole file.

### Before submitting

Brunotlps:
## Summary

- `migrations.Rollback` ran each companion down file as a single `Exec(wholeFile)` call. On MySQL, `go-sql-driver/mysql` rejects a multi-statement string in one `Exec` unless the DSN opts in with `multiStatements=true` — which no gombit-generated or documented MySQL DSN does — so any down file with more than one statement failed with a 1064 syntax error. SQLite and PostgreSQL masked this because their drivers tolerate a multi-statement `Exec`.
- Split each down file with the same `splitSQLStatements` helper `Seed` already uses, and execute one statement per `Exec`, in order (inside the existing transaction on SQLite/Postgres, sequentially on MySQL). The failure message now names the failing statement index and file, as suggested in the issue.

Closes #249

## Acceptance criteria

The issue is a bug report without an explicit AC checklist; its "Expected" / "Suggested fix" sections are the contract this PR satisfies:

- [x] Rollback executes down files statement-by-statement the same way `Seed` does (`splitSQLStatements`), so a documented multi-statement down file works on all three supported drivers.
- [x] The MySQL partial-failure error message names the exact statement that failed rather than the whole file.

## Scope notes

None — this is scoped to `migrations.Rollback`'s down-file execution, matching the issue exactly. No M6 battery, no API surface change.

## Validation

- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go test -tags integration ./migrations -migrations.postgres-dsn ... -migrations.mysql-dsn ...` against real Postgres 16 and MySQL 8.4 containers — confirmed the MySQL test reproduces the original 1064 error against the pre-fix code and passes after the fix
- [x] Project `code-review` skill run against this diff (two rounds: first round found a non-discriminating SQLite unit test and an under-specified doc comment, both fixed and re-verified in the second round — APPROVE)

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix
- [ ] Stable features ship docs and appear in an example app — N/A, this is a bug fix to existing documented behavior, not a new feature
- [x] Extraction preserves contracts — refactor and move, don't rewrite code that already passes its tests — reuses `seed.go`'s existing `splitSQLStatements`, no new parsing logic
- [ ] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files (if this PR touches generators) — N/A, no generator touched

### DIFF
--- a/migrations/apply_test.go
+++ b/migrations/apply_test.go
@@ -250,6 +250,59 @@ func TestMigrateRecordsOnlyVersionsPresentInAtlas(t *testing.T) {
 	}
 }
 
+// Note: a happy-path multi-statement SQLite test is intentionally not here.
+// mattn/go-sqlite3 already executes a multi-statement string sequentially in
+// one Exec, and Rollback runs inside a transaction on SQLite, so such a test
+// would pass identically before and after the statement-splitting fix for
+// #249 and prove nothing. The real regression coverage for that fix is
+// TestMigrateRollbackStatusMySQL (migrations/integration_test.go, requires a
+// live MySQL DSN) and the statement-numbering check below.
+
+func TestRollbackMidStatementFailureNamesStatementNumber(t *testing.T) {
+	workDir := t.TempDir()
+	migrationDir := filepath.Join(workDir, "database", "migrations")
+	if err := os.MkdirAll(migrationDir, 0o750); err != nil {
+		t.Fatalf("mkdir: %v", err)
+	}
+	writeFile(t, filepath.Join(migrationDir, "20260101000000_create_widgets.sql"),
+		"CREATE TABLE widgets (id INTEGER PRIMARY KEY);")
+	writeFile(t, filepath.Join(migrationDir, "downs", "20260101000000_create_widgets.down.sql"),
+		"DROP TABLE widgets;\nTHIS IS NOT VALID SQL;")
+
+	dsn := "file:" + filepath.Join(workDir, "app.db") + "?cache=shared&_fk=1"
+	cfg := config.DatabaseConfig{Driver: config.DatabaseDriverSQLite, DSN: dsn}
+	opts := ApplyOptions{
+		WorkDir:      workDir,
+		MigrationDir: migrationDir,
+		Database:     cfg,
+		Stdout:       io.Discard,
+		runner:       &applyFakeAtlas{t: t},
+	}
+	if err := Migrate(context.Background(), opts); err != nil {
+		t.Fatalf("Migrate() error = %v", err)
+	}
+
+	err := Rollback(context.Background(), opts)
+	if err == nil {
+		t.Fatal("Rollback() error = nil, want second-statement failure")
+	}
+	if !strings.Contains(err.Error(), "execute statement 2 of") {
+		t.Fatalf("Rollback() error = %q, want it to name the failing statement", err)
+	}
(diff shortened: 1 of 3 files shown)


leo-aa88 (approved):
LGTM

Issue #232, opened 2026-08-29
[bug] XSS middleware re-encodes JSON bodies, so webhook HMAC verification always fails


leo-aa88:
## Summary

`framework.xssMiddleware` → `sanitizeJSONBody` **unconditionally re-marshals** every `application/json` request body ([framework/xss.go](../framework/xss.go), the `json.Marshal(payload)` + `c.Request.Body = ...cleaned` at the end of `sanitizeJSONBody`) and replaces the request body with the result — even when nothing was stripped. `json.Marshal` re-escapes `<`/`>`/`&` to `\uXXXX`, reorders map keys, and normalizes whitespace, so the bytes a handler reads via `RawBody` differ from the bytes the client sent.

That makes **HMAC signature verification over the raw body impossible**. A GitHub (or Stripe, etc.) webhook signs the *original* body and sends `Content-Type: application/json`; the handler receives the *re-encoded* body, so the recomputed HMAC never matches and every real delivery is rejected. The code comment already acknowledges the hazard:

> Re-encoding JSON may change key order and whitespace; callers that hash the raw body must hash the sanitized bytes handlers actually receive.

— but a webhook caller *cannot* do that: it does not control GitHub's signing, which is over the pre-sanitized bytes.

This is the body-mutation sibling of #226 (CSRF blocked webhooks): `WithCSRFExemptPaths` lets the request reach the handler, but the body is still rewritten, so signature verification fails anyway.

## Reproduce

1. Cookie-mode app with a webhook that verifies `X-Hub-Signature-256` over the raw body, on a `WithCSRFExemptPaths` path.
2. `curl -X POST .../webhooks/github -H 'Content-Type: application/json' -H 'X-Hub-Signature-256: sha256=<hmac of the exact body>' --data-raw '<json with markdown/URLs/unicode>'`.
3. 401 invalid signature. Drop the `Content-Type: application/json` header (so XSS skips) and the same request verifies — proving the body was mutated.

## Proposed fix

Add a raw-body exemption so a webhook path's request body reaches the handler unmodified. Since such an endpoint also cannot do the CSRF double-submit, one option should cover both:

- `framework.WithRawBodyPaths(paths ...string)` — the XSS middleware skips sanitization for these exact paths, and they are also CSRF-exempt (union with `WithCSRFExemptPaths`).
- Webhooks then declare `WithRawBodyPaths("/api/v1/webhooks/github")` and both handle their own auth (HMAC).

## Context

[...]

Issue #137, opened 2026-08-24
[bug] XSS middleware ReadAlls JSON bodies before the request timeout with no size cap


leo-aa88:
### What happened

`GOMBIT_HTTP_REQUEST_TIMEOUT` is documented as the per-request deadline. `framework.Run` sets `WriteTimeout` and `IdleTimeout` to that value and `ReadHeaderTimeout` to 5s. There is **no `ReadTimeout`**.

Middleware order is XSS **then** `request_timeout`. XSS `sanitizeJSONBody` does unbounded `io.ReadAll` on any `Content-Type: application/json` body **before** the request context deadline exists:

```go
raw, err := io.ReadAll(c.Request.Body)
```

A client that finishes headers (within 5s) and then trickles bytes, or sends a very large JSON body, can hold the goroutine and grow process memory. Handlers never see a timeout. `docs/router.md` notes body-size limits as deferred work that would sit *before* XSS; until then this path is live on every JSON POST/PUT/PATCH.

Expected: bound body reads (max bytes and/or `ReadTimeout`), and apply the request timeout before XSS reads the body.

### Steps to reproduce

```sh
gombit new demo --database sqlite
cd demo
# start the API (default RequestTimeout 60s)
```

Send a JSON POST whose headers complete, then keep the body open (slow trickle of `{` + spaces, or a large Content-Length). Example sketch:

```sh
# headers complete; body never finishes
printf 'POST /api/v1/products HTTP/1.1\r\nHost: 127.0.0.1:8080\r\nContent-Type: application/json\r\nContent-Length: 104857600\r\n\r\n{' \
  | nc 127.0.0.1 8080
```

Watch the server goroutine/RSS: XSS is blocked in `ReadAll` well past 60s; `request_timeout` middleware has not run yet.

A unit test can wrap `xssMiddleware` with a `io.Reader` that never EOFs and assert the handler is not reached within `RequestTimeout`.

### gombit version

dev

### Environment

```
gombit:   dev
commit:   0f43797
built:    unknown
go:       go1.25.7
platform: linux/amd64
```

Checked on `main` at 0f43797 (`framework/xss.go` `sanitizeJSONBody`, `framework/app.go` `http.Server` timeouts and `runtimeMiddlewareStack`).

### Database

not database related

### Auth mode

not auth related

### Area

runtime (framework.App, routing, lifecycle)

### Logs or output

`http.Server`:

```
ReadHeaderTimeout: 5s
[...]

PR #261, opened 2026-09-02
fix: stripHTML preserves text after an unclosed XSS skip tag

### LINKED ISSUE #201: [bug] XSS middleware silently discards all text after an unclosed <script>/<style>/<textarea> tag (data loss)

Brunotlps:
## Summary

- `stripHTML` (`framework/xss.go`) discarded all text following an unclosed
  `<script>`/`<style>`/`<textarea>`/`<iframe>`/`<object>`/`<embed>`/
  `<noscript>` start tag, causing silent data loss for any legitimate field
  that merely mentioned one of these tag names in prose.
- Text inside a skip element is now buffered via a stack of per-element
  checkpoints instead of dropped immediately; a checkpoint is popped and the
  buffer truncated back to it on that element's own matching close, so only
  content actually contained by a *closed* dangerous element is discarded.
  Whatever remains unresolved at EOF is re-tokenized (bounded to 4 rounds)
  before reaching the output, instead of being trusted or emitted verbatim.
- `completeHTMLTag` guards the `stripHTML` entry point only. It is the #118
  gate for user-submitted comparison text, not a stripper, so it is never
  applied to recovered skip content — anything still holding `<` or `>` on
  the flush path is re-tokenized unconditionally.

Closes #201

## Acceptance criteria

The issue is a bug report without an explicit AC checklist; its "Expected" /
"Suggested fix" sections are the contract this PR satisfies:

- [x] An unclosed skip element strips the tag itself but keeps the
      following text as plain text, instead of discarding the remainder of
      the string.
- [x] Properly closed dangerous elements still have their content discarded
      (fail-closed preserved for the case it actually protects).
- [x] Text recovered from an unclosed skip element carries no executable
      markup, which is the premise the issue's suggested fix rests on.

## Scope notes

Three security-relevant edge cases were found and closed in this PR rather
than filed separately, since all three are direct consequences of the
`stripHTML` change it makes — not new scope:

- a quoted-attribute tag-stripper bypass on the flush path;
- nested `object`/`embed` content leaking past a shared depth counter;
- `completeHTMLTag` used as an admission gate on recovered skip content
  (@leo-aa88's review), which emitted live markup the regexp does not
  recognize.

**Flagged, not fixed here:** the same regexp gate at the `stripHTML` entry
point lets `<script/src=x>body` and `y` through
unchanged when they arrive at top level rather than nested inside an unclosed
wrapper. Verified on `main`, so it predates this branch, and #118/#137/#232/
#241 (closed) and #201/#271 (open) do not cover it. Fixing it here would
widen the #118 gate on a data-loss bugfix, so it is raised in review for a
decision instead — and it may be moot under #271.

The recovered-text path deliberately does not inherit the #118 protection:
`<script>if (a<b && c>d) return` yields `if (ad) return`. Documented in
`docs/router.md` and covered by tests.

## Validation

- [x] `go build ./...`
- [x] `go test ./...` — full suite, no failures
- [x] `go test ./framework/... -run TestStripHTML` — 36 cases covering the
      unclosed/nested/quote-breakout/slash-separated-tag/recursion-budget
      boundaries. The regression cases for the gate bug were written first and
      confirmed failing against the previous commit before the fix landed.
- [x] Behavior compared against `main` for every case this PR changes, to
      separate regressions introduced here from pre-existing holes.
- [x] Re-tokenizing measured on 1 MiB inputs (plain / incomplete tags / live
      tags / deep unclosed chain): 6.8–26.6 ms, linear, budget-bounded.
- [x] Project `code-review` skill run against this diff — verdict COMMENT; the
      one finding (the docs illustrated the recovered-text tradeoff with its
      mildest example) is fixed in this push.

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — N/A, no database touched; tests added for the new stripHTML behavior
- [ ] Stable features ship docs and appear in an example app — N/A, this is a bug fix to existing documented behavior (`docs/router.md` updated to match), not a new feature
- [ ] Extraction preserves contracts — N/A, no template extraction in this change

### DIFF
--- a/framework/xss.go
+++ b/framework/xss.go
@@ -241,65 +241,70 @@ func sanitizeJSONValue(value any, fieldName string) bool {
 	return changed
 }
 
-// stripHTML removes HTML tags and discards content inside dangerous elements.
-// Plain strings without "<" or ">" are returned unchanged. Strings that
-// contain "<" / ">" but no complete tag (no closing ">", e.g. "a<b") are
-// also returned unchanged so comparison text is not truncated.
+// stripHTML removes HTML tags and discards the content of dangerous elements
+// (xssSkipElementContent). Strings holding no "<" or ">" are returned
+// unchanged, and so are strings that hold them without forming a complete tag
+// ("a<b"): golang.org/x/net/html reads `"<" + letter` as a start tag and would
+// silently eat the rest of the string (issue #118).
 //
-// An unclosed skip element (for example `<script src=x>...`) strips the tag
-// itself but keeps the following text (issue #201). Text seen while
-// skipDepth > 0 is buffered into skipBuf rather than dropped. skipStarts is
-// a stack of checkpoints into skipBuf, one per currently-open skip element:
-// pushed with the length of skipBuf at the moment that element opened,
-// popped and truncated back to on its matching close. That discards exactly
-// the content opened-and-closed by that one element, however deep it is
-// nested inside other still-open skip elements — a single shared skipDepth
-// counter is not enough, since script/style/textarea/noscript/iframe are
-// HTML "raw text" elements (see below) and can never really nest, but
-// object/embed are ordinary elements and can: for `<object><object>x
-// </object>y`, the inner </object> must discard "x" on its own, without
-// waiting for (or depending on) the outer <object> ever closing too.
-//
-// Whatever remains in skipBuf when the tokenizer runs out of input
-// (skipDepth still > 0 at ErrorToken, nothing left open ever closed) is
-// flushed to the output — re-tokenized from scratch, see below. Self-closing
-// skip tags do not enter skip mode.
-//
-// Re-tokenizing applies the same skip logic again, so a nested dangerous
-// element that finds its own closing tag inside the buffer is still
-// discarded — only text that is not inside any closed dangerous element,
-// anywhere in the chain of unclosed wrappers, survives to the output.
-//
-// The flush re-tokenizes rather than emitting the buffer verbatim, and it
-// must do so with the real tokenizer, not a regexp: script/style/textarea/
-// noscript/iframe are HTML "raw text" elements, so golang.org/x/net/html
-// scans them for a literal closing tag and, finding none, hands back
-// everything up to EOF as one opaque, never-tokenized TextToken — it can
-// still contain complete, attribute-quoted tags that were never parsed. A
-// naive regexp strip (completeHTMLTag) does not track quoting, so a `>`
-// inside a quoted attribute value (`onerror="if(1>0){...}"`) ends the match
-// early and leaks the rest of the tag as text — a classic tag-stripper
-// bypass. Re-tokenizing lets the real HTML tokenizer parse the attribute
-// correctly instead.
-//
-// maxUnclosedSkipRecursion bounds that re-tokenizing to a fixed number of
-// rounds so a pathological chain of unclosed raw-text tags
-// (`<script><script>...`) cannot force unbounded recursive re-tokenizing of
-// a string shrinking by one tag per round — that would be quadratic in
-// input size. Past the cap, the remaining tail reverts to the pre-#201
-// fail-closed default (discarded) instead of being trusted as plain text;
-// legitimate content practically never nests this deep.
+// An unclosed dangerous element strips the tag but keeps the text that follows
+// (issue #201); stripHTMLUnclosed does that work.
 func stripHTML(s string) string {
+	if !strings.ContainsAny(s, "<>") {
+		return s
+	}
+	// The #118 gate covers the string the user actually submitted, and only
+	// that one — stripHTMLUnclosed explains why recovered skip content must
+	// never be admitted by it.
+	if !completeHTMLTag.MatchString(s) {
+		return s
+	}
 	return stripHTMLUnclosed(s, maxUnclosedSkipRecursion)
 }
 
 const maxUnclosedSkipRecursion = 4
 
+// stripHTMLUnclosed tokenizes s, buffering the text of open skip elements so
+// that one which never closes can flush that text instead of dropping it
+// (issue #201), and re-tokenizes what it flushes, at most budget times.
+//
+// Text seen while skipDepth > 0 goes to skipBuf. skipStarts is a stack of
+// offsets into skipBuf, one per open skip element, each recording skipBuf's
+// length when that element opened; the element's matching close pops the
+// offset and truncates back to it, discarding exactly that element's content
+// however deep it sits inside other still-open skip elements. A single shared
+// depth counter is not enough: script/style/textarea/noscript/iframe are HTML
+// "raw text" elements and can never really nest, but object and embed are
+// ordinary elements and can — in `<object><object>x</object>y`, the inner
+// close must discard "x" whether or not the outer one ever closes.
+//
+// Whatever is still buffered at EOF goes back through this function rather
+// than being written out verbatim. It has to: a raw-text element missing its
+// closing tag makes the tokenizer hand back everything up to EOF as one
+// opaque, never-parsed TextToken, so the buffer can still hold complete tags.
+// Re-tokenizing applies the same skip logic to them, so a dangerous element
+// that does find its close inside the buffer is still discarded.
+//
+// That second pass must use the tokenizer, never completeHTMLTag. The regexp
+// is the #118 gate for user-typed comparison text, not a stripper; as an
+// admission gate it would emit live every tag it fails to recognize, and it
+// misses two classes the tokenizer handles — an attribute not preceded by
+// whitespace (`<script/src=x>`, where HTML5 reconsumes the "/" and reads
+// `<script src=x>`), and a ">" inside a quoted attribute value
+// (`onerror="if(1>0){...}"`), which ends the match early and leaks the rest of
+// the tag as text. Anything still holding "<" or ">" is therefore
+// re-tokenized unconditionally, which costs the #118 protection inside
+// recovered text: every `"<" + letter` through the next ">" goes, so
+// `<script>if (a<b && c>d) return` yields "if (ad) return". That is markup
+// smuggled through a raw-text element, not the comparison text #118 covers.
+//
+// budget bounds the re-tokenizing: each unclosed raw-text tag buys one more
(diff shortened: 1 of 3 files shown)


framework/xss.go
```diff
@@ -246,10 +246,56 @@ func sanitizeJSONValue(value any, fieldName string) bool {
...
+// (skipDepth still > 0 at ErrorToken, nothing left open ever closed) is
+// flushed to the output — re-tokenized from scratch, see below. Self-closing
 // skip tags do not enter skip mode.
+//
+// Re-tokenizing applies the same skip logic again, so a nested dangerous
+// element that finds its own closing tag inside the buffer is still
+// discarded — only text that is not inside any closed dangerous element,
+// anywhere in the chain of unclosed wrappers, survives to the output.
+//
+// The flush re-tokenizes rather than emitting the buffer verbatim, and it
```
leo-aa88: **[bug]** This comment is the contract for the flush path: skipBuf is re-tokenized, not emitted verbatim, and a regexp must not be the stripper because it misses markup the real tokenizer understands.

`stripHTMLUnclosed` then does this:

```go
if !completeHTMLTag.MatchString(s) {
    return s
}
```

That is the same regexp, used as an admission gate. When it misses, the buffer is returned unchanged — the thing this paragraph says must not happen.

`completeHTMLTag` requires whitespace before attributes (`\s[^>]*`). `<script/src=x>` is a well-known HTML5 parse (`/` is a parse error that reconsumes as before-attribute-name, i.e. `<script src=x>`). The regexp does not match, so the string is emitted as-is.

I ran this against this commit:

- `<textarea><script/src=x>body` → `<script/src=x>body`
- `<script>y` → `y`
- `<b><script/src=x>body` → `body`  (tokenizer actually runs, tags stripped)

Pre-#201 the first two were discarded. The issue’s suggested fix was “trailing text carries no executable markup.” This is executable markup. The #118 gate belongs on the original user string only, not on recovered skip content. On the recursive call, if the buffer still contains `<`/`>`, always tokenize.

framework/xss.go
```diff
@@ -259,27 +305,38 @@ func stripHTML(s string) string {
...
 	b.Grow(len(s))
+	var skipBuf []byte
+	var skipStarts []int
 	tokenizer := html.NewTokenizer(strings.NewReader(s))
 	skipDepth := 0
 	for {
 		switch tokenizer.Next() {
 		case html.ErrorToken:
+			if skipDepth > 0 && budget > 0 {
+				b.WriteString(stripHTMLUnclosed(string(skipBuf), budget-1))
```
leo-aa88, line 321: **[bug]** This is the flush. `budget-1` re-enters `stripHTMLUnclosed`, which applies `completeHTMLTag` before the tokenizer. Skip content from a raw-text element is opaque until here; if the regexp misses, this `WriteString` copies live tags into the output.

Keep `completeHTMLTag` on the top-level `stripHTML` entry so `a<b` still survives. Do not apply it to skipBuf. A regression test with a slash-separated start tag inside an unclosed wrapper (`<textarea><script/src=x>body` → `body`, not the tag) would have caught this; every new case in `xss_test.go` is a string the regexp already accepts.

framework/xss_test.go
```diff
@@ -27,7 +27,60 @@ func TestStripHTML(t *testing.T) {
...
+		// (not even a partial "<script>") leaking out around it.
+		{name: "doubly unclosed wrapper still recovers the innermost plain text", in: `<textarea><script>alert(1)`, want: "alert(1)"},
+		{name: "unclosed wrapper cannot smuggle a live img onerror tag", in: `<noscript>x<img src=x onerror=alert(1)>y`, want: "xy"},
+		// A quoted attribute value may itself contain a literal ">". A regexp
+		// strip that isn't quote-aware ends its match at that inner ">" and
+		// leaks the rest of the tag (including the closing ">") as text — the
+		// classic tag-stripper bypass. stripHTML must re-tokenize with the
+		// real (quote-aware) HTML tokenizer, not a regexp, so the whole <img>
+		// start tag is recognized and dropped instead of partially surviving.
+		{name: "unclosed wrapper: quoted attribute containing > does not defeat the strip", in: `<noscript><img src=x onerror="if(1>0){alert(document.cookie)}">`, want: ""},
```
leo-aa88, line 58: **[bug]** This test is the regexp-as-stripper case (`>` inside a quoted attribute). The flush does not use the regexp to strip; it uses it to decide whether to tokenize at all. Add a case the gate rejects and the tokenizer would strip, nested in an unclosed skip element. Without that, this suite cannot fail the implementation that agrees with itself.

leo-aa88 (changes requested):
## Summary

The #201 data-loss fix is in the right place: buffer skip-element text, discard only what a matching close actually contained, re-tokenize the leftover. The checkpoint stack for nestable `object`/`embed` is the correct model. The quoted-attribute test is aimed at a real tag-stripper class.

It does not do what its own comment promises. The flush is documented as “re-tokenize with the real HTML tokenizer, never emit skipBuf verbatim, regexp is the wrong tool.” `stripHTMLUnclosed` then early-returns skipBuf unchanged whenever `completeHTMLTag` does not match. That is the same regexp, used as an admission gate rather than a stripper. Markup the regexp misses and the tokenizer understands — e.g. `<script/src=x>`, which HTML5 treats as `script` plus a `src` attribute — used to be discarded on the unclosed-skip path and is now emitted as live tags.

CI is green. This is not a failing-test problem. The new cases all exercise inputs `completeHTMLTag` already recognizes.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

Brunotlps:
Fixed in 6162b93 (code + tests) and 1c25f50 (docs). You were right on all three
counts, including that CI could not have caught it.

**The gate moved rather than got patched.** `completeHTMLTag` now runs only at
the `stripHTML` entry point, guarding what it was always for — the string the
user actually submitted, so `a<b` still survives (#118). `stripHTMLUnclosed`
tokenizes whenever the input still holds `<` or `>`; there is no regexp
anywhere on the flush path now.

Your three cases against the new commit:

```
<textarea><script/src=x>body   ->  body
<script><img/src=x>y           ->  y
<b><script/src=x>body          ->  body   (unchanged)
```

**On the tests.** I wrote the regression cases first and confirmed five of them
failing against c52166f before touching `xss.go`, since the point of your third
comment was that the suite could not reject an implementation agreeing with
itself. Added: a slash-separated start tag inside an unclosed wrapper (skip and
non-skip element), one carrying an event handler, one carrying a quoted `>`,
and one that also closes (`<textarea><script/src=x>a</script>b` → `b`).

**The tradeoff is now explicit and tested.** Dropping the gate on the flush
path costs the #118 protection inside recovered text, and it costs more than
the `a<b` example suggests:

```
<script>if (a<b && c>d) return   ->  "if (ad) return"
```

That is in `docs/router.md` and locked by tests, alongside the shapes that do
survive (`score <10`, `price < 10` — only `"<" + letter` opens a tag). My first
draft of those docs used `a<b` as the illustration; running the project
`code-review` skill flagged it as the same species of defect you blocked on — a
comment describing something milder than the runtime — so it was corrected
before pushing.

Also measured, since the flush path lost its cheap regexp bail-out: 1 MiB
inputs run 6.8–26.6 ms across plain text, all-incomplete-tags, all-live-tags,
and a deep unclosed chain. Linear, still budget-bounded.

---

**One finding I would like your call on, outside this PR's scope.**

The same regexp at the entry point has the same blind spot, and there nothing
downstream compensates for it. On `main`, unmodified:

```
<script/src=x>body              ->  <script/src=x>body
<img/src=x onerror=alert(1)>y   ->  <img/src=x onerror=alert(1)>y
```

So after this PR the sanitizer strips `<script/src=x>` when it is nested inside
an unclosed `<textarea>` and passes the identical tag through when it arrives
on its own. It predates this branch, and I could not find it in the backlog —
#118, #137, #232 and #241 are closed, and neither #201 nor #271 covers it.

I did not touch it here: widening the #118 gate inside a data-loss bugfix is
scope creep, and AGENTS.md says to flag rather than add scope. It may also be
moot under #271 — if ingress sanitization leaves the default pipeline, this
becomes an opt-in-path bug rather than a default-stack one. Your call on
whether it warrants an issue.

leo-aa88 (approved):
LGTM

PR #263, opened 2026-09-03
feat: declared server-side list filtering/sort/search (#260)

### LINKED ISSUE #260: [feature] Generic server-side list filtering by declared/filterable fields and FKs in the generated list handler
## Summary

The generated resource **list** handler currently supports only pagination (`page` / `per_page`) and `Order("id")`. It has no way to **filter** the collection by a field value or a foreign key. This blocks two capabilities in the Gombit **Forge** page builder that should share one contract:

- **Table search / filter / sort** (Forge issue #51) — a resource table can't filter/search/sort server-side.
- **Detail-page related records** (Forge issue #53) — a detail page can't list a record's has_many children (e.g. a customer's invoices) because it can't request `GET /api/v1/invoices?customer_id=<id>`.

Forge deliberately does **not** work around this (no client-side FK filtering, no presenting an unfiltered list as "related records") — filtering is a framework primitive that belongs in Gombit (Forge P4 boundary).

## Proposed contract

Add opt-in, declared list-query support to the generated list handler:

- **Filter** by a declared set of filterable fields and belongs_to FKs, e.g. `GET /collection?<field>=<value>` (exact match to start; ranges/operators later).
- **Sort** by a declared set of sortable fields, e.g. `?sort=<field>&order=asc|desc` (replacing the fixed `Order("id")`).
- **Search** across a declared set of searchable text fields, e.g. `?q=<term>`.

"Declared" = the resource opts specific fields in (so this stays safe and indexable), mirroring how list columns are already a declared subset. The exact query-param spelling is Gombit's to define; Forge will consume whatever contract lands.

## Why one contract

Forge's `ProjectSpec` already carries per-resource `searchable_fields` / `sortable_fields` / `filterable_fields` and per-table config. Once Gombit exposes a matching list-query contract, Forge can drive both the table (#51) and detail related-record lists (#53) from it with no client-side hacks.

## Consumers (Forge)

- gombit-dev/gombit-forge #51 (resource table: search/filter/sort at runtime)
- gombit-dev/gombit-forge #53 (detail page: embedded related-record lists)

Both are intentionally kept open on the Forge side pending this feature.

leo-aa88:
## Summary

Adds opt-in, **declared** list-query support to the generated resource list handler — the shared contract the Forge page builder consumes for table search/filter/sort (gombit-dev/gombit-forge#51) and detail-page related-record lists (gombit-dev/gombit-forge#53).

Field-grammar modifiers opt specific fields into a safe, indexable query surface:

- **filter** — `?<field>=<value>` (exact match): `string, int, int64, uint, bool, enum`
- **sort** — `?ordering=<field>` (`-<field>` for DESC): any scalar + belongs_to FK
- **search** — `?search=<term>` (case-insensitive LIKE): `string, text, enum`

A `belongs_to` foreign key is **filterable by default**, so a detail page lists a record's `has_many` children with `GET /api/v1/invoices?customer_id=<id>` and no extra declaration. Filters + search apply before the `COUNT`, so `meta.total` reflects the filtered set; `?ordering=` replaces the fixed `id` order (`id` stays the default when absent) and is validated against the declared sortable set.

The query spelling **matches Gombit's shipped admin data plane** so the framework speaks one dialect. The shared spelling/SQL now lives in one place (`database.Search`, `database.Ordering`/`ParseOrdering`, `database.FilterEq`); `admin.applySearch`/`applyOrdering` were refactored to delegate to it (duplicate `escapeLike`/`quoteIdent` removed) — the admin wire contract and SPA are unchanged.

Closes #260

## Acceptance criteria

- [x] **Filter** by a declared set of filterable fields and belongs_to FKs (`?<field>=<value>`, exact match)
- [x] **Sort** by a declared set of sortable fields (`?ordering=<field>`, `-` prefix for DESC), replacing the fixed `Order("id")`
- [x] **Search** across a declared set of searchable text fields (`?search=<term>`)
- [x] "Declared" = per-field opt-in (grammar modifiers), mirroring the declared-subset approach
- [x] One contract Forge can drive both #51 (table) and #53 (related records) from — aligned with the admin data-plane spelling

## Scope notes

- Filters are exact-match only; ranges/operators are explicitly deferred (noted in docs).
- Admin's filter *plumbing* still uses its registry-driven `coerceFilter` while the generated handler coerces string query params by declared kind — same spelling/semantics, different internals. Fully unifying that path is a reasonable follow-up; this PR unifies search + ordering.
- Generated frontend list pages are unchanged (server-side contract only; Forge is the consumer).

## Validation

- [x] `go test ./...` — green
- [x] `golangci-lint run` (v2.12.2) — 0 issues
- DB helpers covered on SQLite plus PostgreSQL/MySQL integration entries (`TestListQueryHelpers{SQLite,Postgres,MySQL}`); a Huma runtime contract test proves the generated input shape registers and behaves; resourcegen unit tests cover modifier parsing/validation and handler output (incl. a guard that a resource with no modifiers is byte-unchanged); a goldentest end-to-end fixture compiles a resource using all modifiers + a belongs_to.

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix (SQLite in unit tests; PG/MySQL via `-tags integration` DSN flags)
- [x] Stable features ship docs and appear in an example app (docs/contract.md, tutorial; goldentest compile fixture as the executable example)
- [x] Extraction preserves contracts — admin refactor is move-and-delegate; admin tests unchanged and passing
- [x] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files (no-modifier output is byte-identical; guarded by a regression test)
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR — N/A: no framework endpoint changes; this affects *generated apps*, whose OpenAPI/TS is produced by `gombit client generate` at app build time
- [x] No secrets in generated frontend source; `VITE_*` is treated as public — N/A (no frontend change)
- [x] Scope stays inside the issue's milestone — no M6 battery creep

### DIFF
--- a/cli/make.go
+++ b/cli/make.go
@@ -46,7 +46,7 @@ regex), next to product.Register(app). AutoMigrate is updated the same way.
 
 Field grammar (design §27 subset):
 
-  name:type[:required][,unique][,index]
+  name:type[:required][,unique][,index][,filterable][,sortable][,searchable]
 
 Supported types: string, text, int, int64, bool, uint, decimal, time, enum,
 belongs_to, has_many, many_to_many.
@@ -55,6 +55,18 @@ belongs_to, has_many, many_to_many.
   decimal(p,s)       pin precision/scale, e.g. decimal(10,2).
   time               time.Time (RFC3339 in JSON).
   enum(a,b,c)        string column validated against the listed values.
+
+List-query modifiers opt a field into the generated list handler's declared
+query surface (safe, indexable subset):
+
+  filterable         exact-match ?<field>=<value> query param.
+                     Types: string, int, int64, uint, bool, enum. A belongs_to
+                     foreign key is filterable by default (GET /children?
+                     <parent>_id=<id>) with no modifier needed.
+  sortable           ?sort=<field>&order=asc|desc (replaces the fixed id order;
+                     id stays the default when ?sort= is absent).
+  searchable         case-insensitive ?q=<term> LIKE across searchable text
+                     fields. Types: string, text, enum.
 
 Relations use name:kind:Target, where Target is a model in internal/<target>/:
 
@@ -74,6 +86,8 @@ foreign key / explicit join keys. Point relations at a different package.
 Examples:
 
   gombit make resource Widget name:string:required price:int
+  gombit make resource Article title:string:required,searchable,sortable \
+    status:enum(draft,published):filterable author:belongs_to:Author
   gombit make resource Rental price:decimal:required starts_at:time \
     status:enum(requested,confirmed,active,returned,cancelled) \
     engine:belongs_to:Engine warehouses:many_to_many:Warehouse
--- a/cli/make.go
+++ b/cli/make.go
@@ -57,16 +57,18 @@ belongs_to, has_many, many_to_many.
   enum(a,b,c)        string column validated against the listed values.
 
 List-query modifiers opt a field into the generated list handler's declared
-query surface (safe, indexable subset):
+query surface (safe, indexable subset). The query spelling matches Gombit's
+admin data plane so the two contracts stay in sync:
 
   filterable         exact-match ?<field>=<value> query param.
                      Types: string, int, int64, uint, bool, enum. A belongs_to
                      foreign key is filterable by default (GET /children?
                      <parent>_id=<id>) with no modifier needed.
-  sortable           ?sort=<field>&order=asc|desc (replaces the fixed id order;
-                     id stays the default when ?sort= is absent).
-  searchable         case-insensitive ?q=<term> LIKE across searchable text
-                     fields. Types: string, text, enum.
+  sortable           ?ordering=<field> (prefix with - for DESC, e.g.
+                     ?ordering=-created_at). Replaces the fixed id order; id
+                     stays the default when ?ordering= is absent.
+  searchable         case-insensitive ?search=<term> LIKE across searchable
+                     text fields. Types: string, text, enum.
 
 Relations use name:kind:Target, where Target is a model in internal/<target>/:
 
--- a/cli/make.go
+++ b/cli/make.go
@@ -65,7 +65,7 @@ admin data plane so the two contracts stay in sync:
                      foreign key is filterable by default (GET /children?
                      <parent>_id=<id>) with no modifier needed.
   sortable           ?ordering=<field> (prefix with - for DESC, e.g.
-                     ?ordering=-created_at). Replaces the fixed id order; id
+                     ?ordering=-title). Replaces the fixed id order; id
                      stays the default when ?ordering= is absent.
   searchable         case-insensitive ?search=<term> LIKE across searchable
                      text fields. Types: string, text, enum.
--- a/resourcegen/fields.go
+++ b/resourcegen/fields.go
@@ -19,6 +19,15 @@ type Field struct {
 	Index    bool
 	Nullable bool
 
+	// Filterable / Sortable / Searchable opt this field into the generated list
+	// handler's declared query surface (issue #260): exact-match ?<field>=,
+	// ?sort=<field>&order=, and ?q= respectively. belongs_to fields are
+	// filterable by default (see filterable) so the has_many detail-list case —
+	// GET /children?<parent>_id=<id> — works without extra declaration.
+	Filterable bool
+	Sortable   bool
+	Searchable bool
+
 	// EnumValues holds the allowed values for FieldEnum, in declared order.
 	EnumValues []string
 	// Precision/Scale set the decimal(p,s) column for FieldDecimal.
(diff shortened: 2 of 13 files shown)


leo-aa88 (commented):
## Summary

The declared list-query surface is the right shape: opt-in modifiers, belongs_to filterable by default, filters/search before COUNT, admin spelling (`?search=` / `?ordering=` / `?<field>=`), shared `database` primitives. The SQLite+PG+MySQL helper tests, the Huma runtime contract test, and the no-modifier regression guard are aimed at the actual contract.

The generator then writes those new query params onto the same list-input struct as `page` / `per_page` / `search` / `ordering` with no reservation, and will emit duplicate Go fields and duplicate `query` tags for ordinary names. `gofmt` accepts it. `go build` does not. `reservedFields` already exists for `id` / `created_at` / `updated_at` / `deleted_at` and was not extended.

CI is green. This is not a failing-test problem. The current tests do not exercise a filterable `page` or a filterable `search` next to a searchable title.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

resourcegen/files.go
```diff
@@ -250,9 +285,22 @@ func renderHandler(ctx renderContext) string {
...
-	b.WriteString("\tPerPage int `query:\"per_page\" doc:\"Page size\"`\n}\n\n")
+	b.WriteString("\tPerPage int `query:\"per_page\" doc:\"Page size\"`\n")
+	if len(searchCols) > 0 {
+		b.WriteString("\tSearch string `query:\"search\" doc:\"Search term matched across searchable fields\"`\n")
+	}
+	if len(sortCols) > 0 {
+		b.WriteString("\tOrdering string `query:\"ordering\" doc:\"Field to order by; prefix with - for DESC (allowed: " + strings.Join(sortCols, ", ") + ")\"`\n")
+	}
+	for _, f := range filters {
+		b.WriteString("\t" + f.filterInputField() + " string `" + f.filterQueryTag() + "`\n")
```
leo-aa88, line 301: **[bug]** Filter fields are appended onto the same struct that already declared `Page`, `PerPage`, and (conditionally) `Search` / `Ordering`. `filterInputField()` is `dtoGoName()`, so a declared field whose JSON name is `page`, `per_page`, `search`, or `ordering` becomes a second struct field with the same name and the same `query` tag.

I generated these against this commit (`format.Source` succeeded):

- `page:int:filterable` → two `Page` fields, both `query:"page"`
- `per_page:int:filterable` → two `PerPage` fields
- `title:string:searchable` + `search:string:filterable` → two `Search` fields, both `query:"search"`
- `title:string:sortable` + `ordering:string:filterable` → two `Ordering` fields

`page` as a filterable int is not an exotic name. `reservedFields` already rejects `id` / timestamps because they collide with `gorm.Model`. The list dialect this PR introduces (`page`, `per_page`, `search`, `ordering`) needs the same treatment — reject at `parseFields`, before anything is written. A generator that emits a non-compiling list input is not a declared query surface.

resourcegen/listquery_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func TestListQueryModifierTypeErrors(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name    string
+		spec    string
+		wantErr string
+	}{
+		{"searchable int", "views:int:searchable", "cannot be searchable"},
```
leo-aa88, line 37: **[bug]** Type-mismatch tests only. Nothing here (or in `TestRenderHandlerListQuery`) compiles a filterable `page` / `search` / `ordering` / `per_page` and asserts the generator refuses. `TestRenderHandlerNoListQuery` guards the no-modifier happy path. An implementation that emits duplicate struct fields still passes this file.

resourcegen/fields.go
```diff
@@ -19,6 +19,15 @@ type Field struct {
 	Index    bool
 	Nullable bool
 
+	// Filterable / Sortable / Searchable opt this field into the generated list
+	// handler's declared query surface (issue #260): exact-match ?<field>=,
+	// ?sort=<field>&order=, and ?q= respectively. belongs_to fields are
```
leo-aa88: **[suggestion]** The alignment commit changed the wire spelling to `?search=` / `?ordering=`. This comment (and `typeAllowsSearch` / `searchColumns` / `sortColumns`) still advertise `?q=` and `?sort=&order=`. Docs and generated tags are correct; the type that owns the modifiers is not. Stale comments on a contract type are how the next spelling change drifts.

leo-aa88:
Addressed in 4498700.

- **[bug] duplicate list-input fields** — `page` / `per_page` / `search` / `ordering` are now reserved and rejected at `parseField` (new `reservedQueryFields`, same upfront treatment as `reservedFields` for `gorm.Model`), so a `page:int:filterable` (or any of the four, with or without a modifier) errors before anything is written instead of emitting a second struct field + duplicate `query` tag.
- **[bug] test gap** — added `TestReservedQueryFieldNamesRejected` covering the bare and modifier'd forms of all four names.
- **[suggestion] stale comments** — the `Filterable`/`Sortable`/`Searchable` doc and `typeAllowsSearch` / `searchColumns` / `sortColumns` now say `?search=` / `?ordering=` instead of the pre-alignment `?q=` / `?sort=&order=`.

`go test ./...` green; `golangci-lint` 0 issues.

framework/ablation_bench_test.go
```diff
@@ -0,0 +1,85 @@
+package framework
+
+// Middleware-ablation benchmark (investigation scratch; not part of the
+// committed suite). Builds the Gombit request pipeline one layer at a time on
```
leo-aa88: **[suggestion]** This file is not list-query work. It landed in `fix(resourcegen): reserve list-query param names; refresh stale comments`, and its own comment says "investigation scratch; not part of the committed suite." That is the file telling you not to commit it.

`go test ./framework` now has to compile a middleware-ablation harness that imports `benchmarks/micro/scenario` from the runtime package's test graph. That coupling is a BENCH-1 investigation, not issue #260.

Drop it from this PR. If the ablation is useful, it belongs in `benchmarks/` under BENCH-1, in its own change, without a comment that claims it is not committed.

cli/make.go
```diff
@@ -56,6 +56,20 @@ belongs_to, has_many, many_to_many.
...
+List-query modifiers opt a field into the generated list handler's declared
+query surface (safe, indexable subset). The query spelling matches Gombit's
+admin data plane so the two contracts stay in sync:
+
+  filterable         exact-match ?<field>=<value> query param.
+                     Types: string, int, int64, uint, bool, enum. A belongs_to
+                     foreign key is filterable by default (GET /children?
+                     <parent>_id=<id>) with no modifier needed.
+  sortable           ?ordering=<field> (prefix with - for DESC, e.g.
+                     ?ordering=-created_at). Replaces the fixed id order; id
```
leo-aa88: **[nit]** `?ordering=-created_at` cannot be produced by this generator. `created_at` is already in `reservedFields` (gorm.Model), and the sortable set is only grammar fields with `sortable`. Admin can order by timestamps implicitly; the generated handler cannot. Use `-title` or `-views` in the example, or the help describes a contract the grammar cannot declare.

leo-aa88 (commented):
## Summary

HEAD (`4498700`) closes the previous list-input collision: `reservedQueryFields` rejects `page` / `per_page` / `search` / `ordering` at `parseField`, with tests for bare and modifier forms. The stale `?q=` / `?sort=` comments are gone. The #260 contract now matches the generator.

The same commit also adds `framework/ablation_bench_test.go`, an investigation scratch file whose own comment says it is not part of the committed suite. That file is unrelated to #260 and should not land in this PR.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

The list-query work is otherwise ready. Do not merge the scratch bench in with it.

leo-aa88 (commented):
LGTM

PR #273, opened 2026-09-03
[feature] Declared server-side numeric aggregates for list resources (#272)

### LINKED ISSUE #272: [feature] Declared server-side numeric aggregates for list resources
## Problem

Gombit Forge's dashboard builder (gombit-forge #54) offers **numeric aggregate cards** — e.g. "total invoiced this month", "average order value" — over a resource collection. There is currently no server-side way to compute these: the generated list handler returns rows + a `PageMeta` count, but no `SUM`/`AVG`/`MIN`/`MAX` over a field. Forge will **not** compute aggregates client-side (it would be wrong over a paginated/filtered set and leak all rows to the browser), mirroring the decision that produced declared server-side filtering (#260).

## Requested contract

A declared, safe, opt-in aggregate surface on the generated resource list handler, in the same spirit as #260's `filterable`/`sortable`/`searchable`:

- A field modifier (e.g. `aggregatable`) that opts a numeric field (`int`, `int64`, `uint`, `decimal`) into aggregation, validated per type.
- A way to request one or more aggregates over an aggregatable field on the collection endpoint — e.g. `GET /invoices?aggregate=sum:total,avg:total` — returning the computed values in a stable envelope (alongside or instead of rows).
- Aggregates should respect the **same filters/search** as the list (so "total for filtered set" is possible), computed server-side before/independent of pagination — consistent with how #260 applies filters before the COUNT.
- Shared primitives in the `database` package (as #260 put `FilterEq`/`Search`/`Ordering` there), so generated apps and the admin data plane share one implementation.

## Consumer

- gombit-forge #54 (dashboard numeric-aggregate cards) is the named consumer; the Forge-side work is tracked as a dedicated blocked issue that will build against whatever contract ships here.

Refs: parallels #260 (declared list filtering/sort/search).

leo-aa88:
## Summary

Adds an opt-in **numeric aggregate** surface to the generated resource list handler, in the same spirit as #260's declared `filterable`/`sortable`/`searchable`. A numeric field marked `aggregatable` can be reduced server-side with `SUM`/`AVG`/`MIN`/`MAX` over the collection, respecting the list's filters/search and computed before pagination — the contract Gombit Forge's dashboard numeric-aggregate cards consume (gombit-forge #54).

- **Modifier:** `aggregatable`, valid only on `int`/`int64`/`uint`/`decimal`, validated per type at parse time.
- **Request:** `GET /invoices?aggregate=sum:total,avg:total` (comma-separated `<func>:<field>`; funcs `sum`/`avg`/`min`/`max`).
- **Response:** values in `meta.aggregates`, keyed `"<func>:<field>"`, as canonical decimal strings (`types.Decimal`). **Precision is per driver:** exact for integer aggregates everywhere and for decimal `SUM` on Postgres/MySQL; SQLite computes `AVG` and fractional decimal `SUM` in IEEE float, and `AVG` is an approximation on any driver — see docs/contract.md. Omitted when no `?aggregate=` is requested, so the envelope stays byte-identical to a `PageMeta` response.
- **Scope:** computed over the same filtered/searched query as the list, before pagination (like the COUNT). Empty match set → `0`. Unknown func / undeclared field / malformed pair → D10 `validation_error` (422) on `aggregate`.
- **Shared primitives** live in the `database` package (`ParseAggregates`, `Aggregate`) alongside `FilterEq`/`Search`/`Ordering`; the meta type is the new `contract.ListMeta` (PageMeta + optional `aggregates`). The generator emits the wiring only when a field opts in, so a resource with no aggregatable field is unchanged (regression-guarded).

> Builds on #260's shared list-query primitives (`FilterEq`/`Search`/`Ordering`), now merged to `main`; rebased onto `main`.

Closes #272

## Acceptance criteria

- [x] Field modifier `aggregatable` opts a numeric field (`int`, `int64`, `uint`, `decimal`) into aggregation, validated per type (non-numeric types rejected at parse time).
- [x] Request one or more aggregates on the collection endpoint — `?aggregate=sum:total,avg:total` — returning computed values in a stable envelope (`meta.aggregates`, keyed `"<func>:<field>"`).
- [x] Aggregates respect the same filters/search as the list, computed server-side before/independent of pagination.
- [x] Shared primitives in the `database` package (`ParseAggregates`, `Aggregate`), mirroring where #260 put `FilterEq`/`Search`/`Ordering`.

## Scope notes

- **Admin data plane** is intentionally not wired for aggregates: it uses its own query helpers (`applyFilters`/`applySearch`/`applyOrdering`), and the named consumer (Forge, gombit-forge #54) drives the generated resource handlers, not admin — matching #260's boundary. Can follow separately if admin cards are ever needed.
- Only `sum`/`avg`/`min`/`max`; grouping / per-bucket aggregates are out of scope.

## Validation

- [x] `go test ./...` (database, contract, resourcegen, goldentest, admin, client, framework green; SQLite unit + PG/MySQL integration wrappers for the aggregate helpers)
- [x] `go build ./...`, `go vet`, `gofmt -l` clean on changed files
- [ ] `golangci-lint` (not run locally)

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix (SQLite unit + `TestAggregateHelpers{Postgres,MySQL}` integration wrappers)
- [x] Stable features ship docs and appear in an example app — docs in `docs/contract.md` + tutorial; a goldentest end-to-end **compile fixture** exercises a generated app (parity with #260, which documented rather than adding the modifier to a committed example app)
- [x] Extraction preserves contracts — additive; a resource with no `aggregatable` field generates byte-identical output (regression-guarded)
- [x] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files
- [x] API changes regenerate the OpenAPI doc + TS client in this PR — N/A: the framework's sample app doesn't use the modifier, so `client.SampleApp()` and its generated spec/client are unchanged (verified: client/openapi suites pass, no drift)
- [x] No secrets in generated frontend source; `VITE_*` is treated as public — N/A (no frontend changes)
- [x] Scope stays inside this issue's milestone — no M6 battery creep

### DIFF
--- a/cli/make.go
+++ b/cli/make.go
@@ -76,8 +76,10 @@ aggregates:
   aggregatable       server-side ?aggregate=<func>:<field> (func: sum, avg,
                      min, max), computed over the filtered set before
                      pagination and returned in meta.aggregates. Types: int,
-                     int64, uint, decimal. Exact on Postgres/MySQL; SQLite
-                     computes AVG and fractional decimal SUM in float.
+                     int64, uint, decimal. Integer aggregates and decimal sum
+                     are exact on Postgres/MySQL; SQLite computes avg and
+                     fractional decimal sum in float. avg is an approximation
+                     on any driver.
 
 Relations use name:kind:Target, where Target is a model in internal/<target>/:
 
--- a/cli/make.go
+++ b/cli/make.go
@@ -46,7 +46,7 @@ regex), next to product.Register(app). AutoMigrate is updated the same way.
 
 Field grammar (design §27 subset):
 
-  name:type[:required][,unique][,index][,filterable][,sortable][,searchable]
+  name:type[:required][,unique][,index][,filterable][,sortable][,searchable][,aggregatable]
 
 Supported types: string, text, int, int64, bool, uint, decimal, time, enum,
 belongs_to, has_many, many_to_many.
@@ -69,6 +69,10 @@ admin data plane so the two contracts stay in sync:
                      stays the default when ?ordering= is absent.
   searchable         case-insensitive ?search=<term> LIKE across searchable
                      text fields. Types: string, text, enum.
+  aggregatable       server-side ?aggregate=<func>:<field> (func: sum, avg,
+                     min, max), computed over the filtered set before
+                     pagination and returned in meta.aggregates. Types: int,
+                     int64, uint, decimal.
 
 Relations use name:kind:Target, where Target is a model in internal/<target>/:
 
@@ -90,6 +94,8 @@ Examples:
   gombit make resource Widget name:string:required price:int
   gombit make resource Article title:string:required,searchable,sortable \
     status:enum(draft,published):filterable author:belongs_to:Author
+  gombit make resource Invoice total:decimal:required,aggregatable \
+    quantity:int:aggregatable,filterable customer:belongs_to:Customer
   gombit make resource Rental price:decimal:required starts_at:time \
     status:enum(requested,confirmed,active,returned,cancelled) \
     engine:belongs_to:Engine warehouses:many_to_many:Warehouse
--- a/cli/make.go
+++ b/cli/make.go
@@ -69,10 +69,15 @@ admin data plane so the two contracts stay in sync:
                      stays the default when ?ordering= is absent.
   searchable         case-insensitive ?search=<term> LIKE across searchable
                      text fields. Types: string, text, enum.
+
+The generated list handler (not the admin data plane) also supports numeric
+aggregates:
+
   aggregatable       server-side ?aggregate=<func>:<field> (func: sum, avg,
                      min, max), computed over the filtered set before
                      pagination and returned in meta.aggregates. Types: int,
-                     int64, uint, decimal.
+                     int64, uint, decimal. Exact on Postgres/MySQL; SQLite
+                     computes AVG and fractional decimal SUM in float.
 
 Relations use name:kind:Target, where Target is a model in internal/<target>/:
 
--- a/database/aggregate.go
+++ b/database/aggregate.go
@@ -117,10 +117,17 @@ func ParseAggregates(ctx context.Context, raw string, allowed map[string]Aggrega
 // ("<func>:<field>"). All aggregates are evaluated in a single SELECT with no
 // GROUP BY, so the query returns exactly one row and runs before pagination.
 //
-// Values come back as types.Decimal (exact, JSON-string) regardless of driver
-// or column type, so money totals never round through float. A NULL result — an
-// aggregate over an empty matching set (SUM/AVG/MIN/MAX all yield NULL) —
-// becomes decimal zero, so a card renders 0 rather than null.
+// Values are the database's own aggregate, encoded as a types.Decimal (a
+// canonical JSON string). Precision therefore follows the driver: Postgres and
+// MySQL compute SUM/AVG/MIN/MAX over numeric/decimal in fixed point, so a
+// decimal SUM (and integer aggregates on every driver) is exact; SQLite has no
+// native fixed-point aggregate and computes AVG — and SUM of a fractional
+// decimal column — in IEEE double, so those may carry float rounding and can
+// differ from Postgres/MySQL for the same data. AVG is fractional by nature
+// (rounded to each driver's default scale) and is an approximation everywhere.
+// See docs/contract.md "List query: numeric aggregates" for the per-driver
+// contract. A NULL result — an aggregate over an empty matching set (SUM/AVG/
+// MIN/MAX all yield NULL) — becomes decimal zero, so a card renders 0 not null.
 //
 // An empty specs list is a no-op (nil map, no query). Column names in specs are
 // generator-controlled; nothing here is derived from user input.
@@ -158,11 +165,14 @@ func Aggregate(ctx context.Context, q *gorm.DB, specs []AggregateSpec) (map[stri
 }
 
 // toDecimal normalizes a driver-scanned aggregate value into a types.Decimal.
-// Drivers return aggregate results with different Go types: SQLite yields int64
-// or float64, Postgres yields int64 for SUM(int) and []byte/string for numeric
-// AVG, MySQL yields []byte for its decimal results. A NULL result (nil) becomes
-// decimal zero. AVG is inherently fractional, so a float64 is converted through
-// its shortest exact decimal representation.
+// Drivers return aggregate results with different Go types: Postgres yields int64
+// for SUM(int) and a numeric string/[]byte for fixed-point AVG/decimal SUM,
+// MySQL yields []byte for its decimal results — both exact. SQLite yields int64
+// for integer aggregates but float64 for AVG and for SUM of a fractional decimal
+// column, because it has no fixed-point aggregate; that float64 is encoded via
+// its shortest round-tripping decimal string, so the value is faithful to what
+// SQLite computed (float rounding included) rather than silently reshaped. A
+// NULL result (nil) becomes decimal zero.
 func toDecimal(v any) (types.Decimal, error) {
 	switch val := v.(type) {
 	case nil:
--- a/database/aggregate.go
+++ b/database/aggregate.go
@@ -1,0 +1,185 @@
+package database
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/gombit-dev/gombit/contract"
+	"github.com/gombit-dev/gombit/types"
+	"gorm.io/gorm"
+)
+
+// This file holds the reusable server-side aggregate primitives (SUM / AVG /
+// MIN / MAX over a declared numeric column) that the generated resource list
+// handler calls. Like the list-query helpers in listquery.go, the generator
+// only emits calls for fields a resource declares aggregatable, so the surface
+// stays a safe, declared subset. Aggregates run over the same filtered/searched
+// query the list uses, computed before pagination — so a "total invoiced this
+// month" card reflects the whole matching set, not one page (issue #272).
+
+// AggregateFunc is a supported SQL aggregate function.
+type AggregateFunc string
+
+const (
+	AggSum AggregateFunc = "sum"
+	AggAvg AggregateFunc = "avg"
+	AggMin AggregateFunc = "min"
+	AggMax AggregateFunc = "max"
+)
+
+// aggregateSQL maps a validated func to its SQL keyword. Only these four are
+// reachable — ParseAggregates rejects anything else before it becomes a spec —
+// so the keyword is never derived from user input.
+var aggregateSQL = map[AggregateFunc]string{
+	AggSum: "SUM",
+	AggAvg: "AVG",
+	AggMin: "MIN",
+	AggMax: "MAX",
+}
+
+// AggregateColumn declares one field a resource exposes to aggregates: the DB
+// column the SQL function is applied to. It is generator-controlled — built
+// from a field declared `aggregatable` — so only a declared numeric column can
+// reach the query.
+type AggregateColumn struct {
+	Column string
+}
+
+// AggregateSpec is one resolved aggregate request: Func over Column, returned in
+// the response keyed as "<func>:<field>" (Key).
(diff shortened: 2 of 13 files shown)


database/aggregate.go
```diff
@@ -0,0 +1,185 @@
...
+		return types.Decimal{}, nil
+	case []byte:
+		return types.NewDecimalFromString(strings.TrimSpace(string(val)))
+	case string:
+		return types.NewDecimalFromString(strings.TrimSpace(val))
+	case int64:
+		return types.NewDecimalFromString(strconv.FormatInt(val, 10))
+	case int32:
+		return types.NewDecimalFromString(strconv.FormatInt(int64(val), 10))
+	case float64:
```
leo-aa88, line 188: **[bug]** This is the runtime that the rest of the PR advertises out of existence.

`Aggregate`'s godoc (and `ListMeta`, and `docs/contract.md`) says values come back as `types.Decimal` so money totals never round through float and the string is identical across SQLite/Postgres/MySQL. This case is IEEE float, converted with `FormatFloat(..., 'f', -1, 64)`.

SQLite `AVG()` is a float. The comment ten lines up already admits "SQLite yields int64 or float64" and "AVG is inherently fractional, so a float64 is converted through its shortest exact decimal representation." That is the opposite of "never round through float."

`7/3` on SQLite is `2.3333333333333335`. Postgres `AVG(integer)` is `numeric`, typically a different digit string. `10.10+20.20+0.01` as a `types.Decimal` column on SQLite is `SUM` of REAL/TEXT affinity — also this path. None of that is in the test table: `aggItem.Amount` is `int`, and the only averages are `100/4=25` and `60/3=20`.

CI is green. This is not a failing-test problem. The current tests do not exercise this contract.

Either make the claim true (driver-stable encoding, CAST/text scan for numeric, matrix tests for decimal SUM and a non-integer AVG) or retract it. SQLite cannot do exact decimal `SUM`/`AVG` natively; if the money card has to be exact, say so per driver instead of putting `types.Decimal` on the envelope and then feeding it a float.

**Suggestion:** Add SQLite+PG+MySQL cases for (1) `types.Decimal` SUM of `10.10+20.20+0.01` and (2) AVG of `1,2,4`. Until those strings match the docs, the docs are false. Also lock `?page=1&per_page=1&aggregate=sum:amount` so pagination cannot shrink the aggregate — the generated order looks right, the HTTP test never asks.

docs/contract.md
```diff
@@ -110,6 +110,60 @@ primitives live in `database` (`FilterEq`, `Search`, `Ordering` / `ParseOrdering
...
+```
+
+```
+GET /api/v1/invoices?status=paid&aggregate=sum:total,avg:total,max:total
+```
+
+The values land in `meta.aggregates`, keyed `"<func>:<field>"`, as exact decimal
+strings (the same canonical string encoding as a `decimal` field — trailing
+zeros trimmed, so the value is identical across SQLite/Postgres/MySQL and money
+totals never round through float):
```
leo-aa88: **[bug]** This is the public contract the runtime does not keep.

"exact decimal strings", "identical across SQLite/Postgres/MySQL", "money totals never round through float." The example two lines down is an invoice `sum:total` — a `decimal` column, the Forge card this issue exists to serve.

`toDecimal`'s `float64` branch is the SQLite AVG (and likely SQLite decimal SUM) path. Integer tests with avg `25` cannot falsify this sentence.

**Suggestion:** Do not ship this paragraph until a three-driver test prints the same string for a decimal SUM and a non-integer AVG, or rewrite it to the behavior that actually exists per driver.

cli/make.go
```diff
@@ -69,6 +69,10 @@ admin data plane so the two contracts stay in sync:
                      stays the default when ?ordering= is absent.
   searchable         case-insensitive ?search=<term> LIKE across searchable
                      text fields. Types: string, text, enum.
+  aggregatable       server-side ?aggregate=<func>:<field> (func: sum, avg,
```
leo-aa88, line 76: **[suggestion]** This bullet is listed under "The query spelling matches Gombit's admin data plane so the two contracts stay in sync." Admin `listResources` has no `aggregate` param and never calls `database.Aggregate`. `?aggregate=` on `/api/v1/admin/resources/{slug}` is dropped.

Issue #272 asked for shared primitives "so generated apps and the admin data plane share one implementation" — the functions live in `database`, admin does not use them. #260 actually wired admin. This bullet is sitting in that sync paragraph without that wiring.

**Suggestion:** Move `aggregatable` out of the admin-sync list (it is generated-handler-only, as the PR scope notes say), or wire admin the way #260 did. Do not leave the help text describing a contract the admin runtime does not implement.

leo-aa88 (commented):
## Summary

The list-handler wiring is in the right place: declared `aggregatable`, `ParseAggregates` over a generator-controlled column map, one SELECT before pagination, D10 422 on a bad spec, `ListMeta` omitted when unasked. That is the #260 shape applied to SUM/AVG/MIN/MAX.

It does not do what its own comments promise. `types.Decimal` is in the envelope so money totals never round through float and the string is identical across SQLite/Postgres/MySQL. `toDecimal` then has a `float64` case that runs `strconv.FormatFloat`. SQLite `AVG` is IEEE float — the implementation's own comment says so. The tests only ever SUM/AVG integers that divide evenly (`10+20+30+40`, avg `25`). CI is green. This is not a failing-test problem. The current tests do not exercise this contract.

The motivating consumer is a money card. The default driver is SQLite.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

database/aggregate.go
leo-aa88: Fixed in d943aa1 — you're right, and the honest path is the one you flagged: SQLite has no fixed-point aggregate, so exact cross-driver decimal SUM/AVG is not achievable there, and no CAST/text scan changes that. So I **retracted the claim** rather than pretend otherwise.

- `toDecimal` comment + `Aggregate` godoc now state the real per-driver contract: exact for integer aggregates everywhere and decimal SUM on PG/MySQL; float-approximate for AVG and fractional decimal SUM on SQLite; AVG approximate anywhere. The float64 path is documented as "faithful to what SQLite computed (float rounding included)," not reshaped.
- Added the matrix cases you asked for to `assertAggregateHelpers` (so they run on SQLite + PG + MySQL): a fractional decimal SUM `10.10+20.20+0.01` and a non-integer AVG on a `decimal(19,4)` column. Since the string legitimately differs by driver, the assertion compares the decimal **value** within tolerance — asserting one fixed string would be the false claim again.
- Locked `?page=1&per_page=1&aggregate=sum:amount` in the runtime test: sum stays 60 over all rows with a 1-row page.

docs/contract.md
leo-aa88: Fixed in d943aa1. The paragraph no longer claims exactness or cross-driver-identical strings. It now has a **Precision is per driver** section: exact on Postgres/MySQL (the config a production money card should run on); SQLite computes AVG and fractional decimal SUM in float, with the concrete `30.10.../30.309999999999995` example, and "for an exact money total, aggregate on Postgres/MySQL." The new decimal SUM + non-integer AVG matrix tests back this behavior, so the prose and the runtime now agree.

cli/make.go
leo-aa88: Fixed in d943aa1 — took the first option. Moved `aggregatable` out of the "matches the admin data plane" paragraph into its own note ("The generated list handler (not the admin data plane) also supports numeric aggregates"), so the help no longer implies admin parity it doesn't have. Wiring admin stays deferred per the scope notes — admin has no aggregate declaration surface and the named consumer (Forge) drives the generated handlers.

leo-aa88 (commented):
## Summary

The previous blocker is gone. Docs, `Aggregate`, `ListMeta`, and `toDecimal` now describe the driver that actually exists: exact integer aggregates everywhere, exact decimal SUM on PG/MySQL, IEEE float for SQLite AVG and fractional decimal SUM, AVG approximate everywhere. The admin-sync help paragraph no longer claims `?aggregate=`. The Huma test now locks pagination-immunity.

What replaced the lie is a test that claims to pin "exact on PG/MySQL" and then asserts `Float64()` within epsilon on every driver. That cannot fail a Postgres path that still goes through `FormatFloat`. The CLI one-liner still says "Exact on Postgres/MySQL" without the AVG caveat the docs just added.

Also: the PR summary still advertises "no float rounding; identical across SQLite/Postgres/MySQL". Update it to match HEAD before merge.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

database/aggregate_test.go
```diff
@@ -0,0 +1,354 @@
...
+		wantApprox(t, out, "sum:amount", 30.31, 1e-6)
+		wantApprox(t, out, "avg:amount", 30.31/3, 1e-3)
+	})
+}
+
+// wantApprox parses an aggregate's decimal string and asserts it is within eps of
+// want. Aggregate values are exact on Postgres/MySQL but float-approximate on
+// SQLite (and AVG is approximate everywhere), so the string is not identical
+// across drivers; the numeric value is.
+func wantApprox(t *testing.T, out map[string]types.Decimal, key string, want, eps float64) {
```
leo-aa88, line 173: **[suggestion]** This helper is the lock for the remaining exactness claim, and it does not lock it.

The comment says Postgres/MySQL decimal SUM is exact and SQLite is float-approximate, so the test compares values within a tolerance rather than strings, "which is exactly the contract." No. If the contract is exact on PG/MySQL, those drivers assert `"30.31"`. `db.Driver()` is already on the type. A single `Float64()` ± `1e-6` (and ± `1e-3` for AVG) is the SQLite arm applied to everyone.

A Postgres implementation that still scanned SUM through `float64`/`FormatFloat` would pass this. So would any string that happens to be numerically close. The integer cases (`avg=25`) could not falsify the old claim; this case cannot falsify the new one on the drivers that still promise exactness.

**Suggestion:** Branch on `db.Driver()`. `wantString(..., "30.31")` for Postgres and MySQL SUM. `wantApprox` only for SQLite SUM and for AVG everywhere. That is the contract `docs/contract.md` just wrote down.

cli/make.go
```diff
@@ -70,6 +70,15 @@ admin data plane so the two contracts stay in sync:
   searchable         case-insensitive ?search=<term> LIKE across searchable
                      text fields. Types: string, text, enum.
 
+The generated list handler (not the admin data plane) also supports numeric
+aggregates:
+
+  aggregatable       server-side ?aggregate=<func>:<field> (func: sum, avg,
+                     min, max), computed over the filtered set before
+                     pagination and returned in meta.aggregates. Types: int,
+                     int64, uint, decimal. Exact on Postgres/MySQL; SQLite
```
leo-aa88: **[suggestion]** "Exact on Postgres/MySQL" is the previous overclaim, compressed.

`docs/contract.md` now says AVG is rounded to each driver's numeric scale and is an approximation **everywhere**. `Aggregate`'s godoc says the same. This line has no AVG caveat, so a reader of the CLI help still believes `avg:total` is exact on Postgres.

**Suggestion:** Match the docs: exact for integer aggregates and for decimal SUM on Postgres/MySQL; SQLite AVG and fractional decimal SUM are float; AVG is approximate on every driver.

database/aggregate_test.go
leo-aa88: Fixed in 194b141. `DecimalPrecision` now branches on `db.Driver()`: Postgres and MySQL assert the exact string `"30.31"` for the fractional decimal SUM, so a PG/MySQL path that scanned SUM through `float64`/`FormatFloat` would now fail; SQLite keeps the value-within-tolerance arm; AVG stays `wantApprox` on every driver (it's rounded to each driver's scale). That's the `docs/contract.md` contract, locked per driver instead of the SQLite arm applied to everyone.

cli/make.go
leo-aa88: Fixed in 194b141. The help line now reads: "Integer aggregates and decimal sum are exact on Postgres/MySQL; SQLite computes avg and fractional decimal sum in float. avg is an approximation on any driver." — matching the godoc and docs.

leo-aa88 (commented):
LGTM

Issue #271, opened 2026-09-03
[PERF-13] Decide: move request-input XSS sanitization out of the default pipeline (5 allocs/op per POST, rewrites API input)


leo-aa88:
## Summary

`xssMiddleware` ([framework/xss.go#L56](https://github.com/gombit-dev/gombit/blob/main/framework/xss.go#L56)) still buffers every JSON POST/PUT/PATCH body before Huma reads the same bytes, even after the #241 fast path. Beyond the cost, the feature rewrites API input: XSS is an output-encoding concern, and stripping markup on ingress means a Gombit app cannot faithfully store `{"description": "x < y"}` (see #118, #201, #232 for the class of bugs this keeps producing). This issue is the decision to move body/query sanitization out of the default pipeline and offer it as an explicit, narrower tool.

## Evidence (measured)

- Cumulative ablation: +xss = +5 allocs/op on valid and invalid POST (`io.LimitReader`, `io.ReadAll`, `bytes.NewReader`, `io.NopCloser`, `mime.ParseMediaType`), 0 on GET without a query string.
- Prototype removing it from the stack (after the content-type fast path): POST 46→42 allocs/op, ≈0.4 µs.
- Huma decodes the body once into `any` for schema validation and once into the typed struct; the sanitizer adds a third full read of the same bytes on the fast path and a fourth decode on the slow path.

## Proposed fix

- Default pipeline: no request-body or query sanitization. Huma decodes exactly once. The response side stays as-is (React escapes text; admin templates must escape untrusted data, which they already do).
- Provide explicit alternatives for apps that want ingress stripping:
  - `gombit.SanitizeHTML(s string) string` (the existing `stripHTML` behavior, exported) for handler-level use.
  - A per-app opt-in, e.g. `Security.SanitizeInput = true` / `GOMBIT_SECURITY_SANITIZE_INPUT`, that installs today's middleware unchanged (keeping `WithRawBodyPaths` semantics).
  - Optionally a field-level mechanism in resourcegen (`sanitize` tag or model option) as a follow-up.
- Document the security posture change: what the default protects against (nothing changes on output), and when to enable ingress sanitization.
- Fix #201 (data loss after an unclosed skip element) in the opt-in path or close it with the same rationale.

## Acceptance criteria

- [ ] Default `framework.App` stack has no xss layer; ablation row is absent or 0 allocs.
- [ ] `{"description":"x < y"}` and `{"note":"<b>bold</b>"}` round-trip unchanged through a default app.
[...]

PR #274, opened 2026-09-03
perf: fast-path the JSON content-type check in the XSS sanitizer (#269)

### LINKED ISSUE #269: [PERF-11] isJSONContentType calls mime.ParseMediaType on every JSON write request (1 alloc/op)
## Summary

`isJSONContentType` ([framework/xss.go#L188](https://github.com/gombit-dev/gombit/blob/main/framework/xss.go#L188)) calls `mime.ParseMediaType` on every POST/PUT/PATCH, which allocates a parameter map even for the exact string `application/json`. One avoidable alloc plus parsing on every JSON write request.

## Evidence (measured)

`alloc_objects` on the full-app valid POST: `mime.ParseMediaType` = 1 alloc/op. Prototype with an exact-match fast path: POST 47→46 allocs/op, no change to any scenario assertion.

## Proposed fix

Check the common forms before falling back to `mime.ParseMediaType`:

```go
if ct == "application/json" { return true }
if len(ct) > 16 && strings.EqualFold(ct[:16], "application/json") && (ct[16] == ';' || ct[16] == ' ') { return true }
return isJSONContentTypeSlow(ct) // existing ParseMediaType path
```

Keep the existing behavior for `application/json; charset=utf-8` (already covered by the prefix branch), `APPLICATION/JSON`, and malformed values.

## Acceptance criteria

- [ ] Table test covering `application/json`, `application/json; charset=utf-8`, `Application/JSON`, `application/jsonx`, `application/vnd.api+json` (must remain false as today), and malformed input, asserting identical results to the previous implementation.
- [ ] `mime.ParseMediaType` no longer appears in the valid-POST alloc profile.

## Scope notes

Small, self-contained. Becomes moot for the default pipeline if the sanitizer is made opt-in, but still applies when it is enabled. Part of #264.

Brunotlps:
## Summary

`isJSONContentType` ran `mime.ParseMediaType` on every `POST`/`PUT`/`PATCH`
carrying a body. That parse allocates its parameter map unconditionally on the
success path (`mime/mediatype.go:151`), so the exact string `application/json`
cost one alloc per request and `application/json; charset=utf-8` cost two —
purely to answer a question a string comparison answers.

This adds a fast path for the two shapes that carry essentially all real
traffic (the exact media type, and the media type followed by the `;` that ends
the subtype) and preserves the full-parse decision verbatim as
`isJSONContentTypeSlow`, which stays the only place a content type is actually
decided.

Measured with `testing.AllocsPerRun`:

| Content-Type | before | after |
| --- | ---: | ---: |
| `application/json` | 1 | **0** |
| `application/json; charset=utf-8` | 2 | **0** |
| `Application/JSON` | 2 | 2 (delegated) |
| `application/jsonx` | 1 | 1 (delegated, still `false`) |

The safety property: **the fast path never returns `false`** — it either
short-circuits a `true` or delegates — so no request body can lose
sanitization through it, and the accepted set is identical by construction
rather than by enumeration.

Only `;` is accelerated. A space delimiter was prototyped and dropped: it would
have been correct only via the prefix fallback in the slow path
(`application/json bogus` fails the media-type parse outright and survives only
there), which would have coupled the fast path to that fallback's exact
looseness. It also bought nothing — that input allocates zero either way,
because `checkMediaTypeDisposition` rejects it before the parameter map is made.

Closes #269

## Acceptance criteria

- [x] Table test covering `application/json`, `application/json; charset=utf-8`,
      `Application/JSON`, `application/jsonx`, `application/vnd.api+json` (still
      `false`), and malformed input, asserting identical results to the previous
      implementation. `TestIsJSONContentType` covers those five plus leading
      whitespace, a tab delimiter, an unparseable tail, form-encoded, and the
      empty string — 16 cases, each additionally asserted against
      `isJSONContentTypeSlow`. `FuzzIsJSONContentType` turns "identical results"
      into a machine-checked differential property (223k local executions, no
      divergence).
- [x] `mime.ParseMediaType` no longer appears in the valid-POST alloc profile.
      The micro harness posts exactly `application/json`
      (`benchmarks/micro/scenario/benchmark.go:78`), which returns on the
      exact-match branch before `ParseMediaType` is reached.

## Scope notes

`isJSONContentTypeSlow` carries a pre-existing asymmetry, **unchanged by this
PR**: when `mime.ParseMediaType` fails on a malformed parameter, the fallback
matches by *prefix*, so `application/jsonx; charset` returns `true` while
`application/jsonx` alone returns `false`.

It is fail-safe — an over-eager yes only costs a decode, and an undecodable body
is handed to the handler untouched — so it is not a correctness or security
problem today. It is pinned in `TestIsJSONContentType` as a **characterization
test** (documenting the behavior that exists, not endorsing it) so this PR's
fast path cannot shift it by accident.

Worth a separate decision, not taken here: `mime.ParseMediaType` already returns
the correctly parsed `mediatype` alongside `ErrInvalidMediaParameter`, and the
current code discards it — the information needed for an exact answer is already
available at that call site. Flagged for the maintainer rather than fixed as
unrequested scope; #271 (PERF-13) may also change this surface wholesale.

No new issue was opened for it, per AGENTS.md ("don't create issues beyond §4
backlog entries without asking first").

## Validation

- [x] `go build ./...`
- [x] `go test ./...` — 44 packages ok, 0 failures
- [x] `go test ./framework/ -fuzz FuzzIsJSONContentType -fuzztime 25s` — 223,538 executions, no divergence (no `testdata/` corpus produced, nothing to commit)
- [x] Project `code-review` skill run against this diff — two rounds. The first
      returned REQUEST CHANGES (misleading invariant comment; the space branch
      depending on the slow path's prefix fallback with no test pinning
      `application/json bogus`; universal-equivalence claim proven only over a
      finite table). All three were addressed — the space branch was removed
      rather than tested around — and the second round returned APPROVE with no
      findings.

Not run (not applicable): the SQLite + PostgreSQL + MySQL matrix — this change
does not touch the database.

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — tests added; no DB surface touched, so the matrix does not apply
- [ ] Stable features ship docs and appear in an example app — N/A: no observable behavior change, so there is nothing to document. `docs/router.md`'s XSS notes remain accurate as written
- [x] Extraction preserves contracts — refactor and move, don't rewrite code that already passes its tests — the full-parse decision is preserved verbatim as `isJSONContentTypeSlow`; only the string literals became a named constant
- [ ] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files — N/A: runtime-only change, no generator touched
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR — N/A: internal middleware gate, the HTTP contract is unchanged
- [ ] No secrets in generated frontend source; `VITE_*` is treated as public — N/A: no frontend surface touched
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep — two files, `framework/` only; no benchmark snapshot, README, docs, or DB changes

### DIFF
--- a/framework/xss.go
+++ b/framework/xss.go
@@ -185,15 +185,64 @@ func sanitizeJSONBody(c *gin.Context) {
 	c.Request.Header.Set("Content-Length", strconv.FormatInt(int64(len(cleaned)), 10))
 }
 
+// jsonMediaType is the only media type the sanitizer decodes. Its length
+// doubles as the fast-path prefix window below.
+const jsonMediaType = "application/json"
+
+// isJSONContentType reports whether contentType names the JSON media type.
+//
+// mime.ParseMediaType is not free: it allocates its parameter map
+// unconditionally on the success path (mime/mediatype.go), so even the exact
+// string "application/json" costs one alloc on every POST/PUT/PATCH carrying a
+// body. The fast path below skips that parse for the two shapes that carry
+// essentially all real traffic (issue #269).
+//
+// The fast path never returns false, and it returns true only where the media
+// type is unambiguously "application/json": the whole string, or everything
+// before the ';' that ends the subtype. Everything else is delegated to
+// isJSONContentTypeSlow, which stays the only place a content type is actually
+// decided — a longer subtype ("application/jsonx"), a structured suffix
+// ("application/vnd.api+json"), a cased spelling with no parameters
+// ("Application/JSON", accepted there via ToLower), leading whitespace, a tab
+// delimiter, an unparseable tail ("application/json bogus"), and the empty
+// string. FuzzIsJSONContentType asserts that the two never disagree.
 func isJSONContentType(contentType string) bool {
+	if contentType == jsonMediaType {
+		return true
+	}
+	// Only ';' is accepted as the delimiter. It is the one byte that ends the
+	// subtype under RFC 2045 grammar, so "application/json" before it is the
+	// media type no matter how the parameters that follow are spelled — even
+	// malformed ones, which the slow path still accepts. A space does not carry
+	// that guarantee: "application/json bogus" fails the media-type parse
+	// outright and survives only through the prefix fallback below, so
+	// accelerating it would make this path depend on that fallback's exact
+	// looseness. It is left to the slow path deliberately.
+	if len(contentType) > len(jsonMediaType) &&
+		contentType[len(jsonMediaType)] == ';' &&
+		strings.EqualFold(contentType[:len(jsonMediaType)], jsonMediaType) {
+		return true
+	}
+	return isJSONContentTypeSlow(contentType)
+}
+
+// isJSONContentTypeSlow is the full-parse decision, unchanged from before the
+// fast path landed: it stays the single place a content type is actually
+// decided.
+func isJSONContentTypeSlow(contentType string) bool {
 	if contentType == "" {
 		return false
 	}
 	mediaType, _, err := mime.ParseMediaType(contentType)
 	if err != nil {
-		return strings.HasPrefix(strings.ToLower(strings.TrimSpace(contentType)), "application/json")
+		// A malformed parameter ("application/json; charset", no "=") fails the
+		// parse even though the media type itself is fine, so fall back to a
+		// prefix match and still sanitize those bodies. Deliberately fail-safe:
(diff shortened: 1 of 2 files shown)


leo-aa88 (approved):
LGTM

PR #286, opened 2026-09-05
[HOST-0/HOST-2] Host contracts (ADR-015) + real readiness probe

### LINKED ISSUE #283: [HOST-2] Runtime health convention (/livez + /readyz)
Gombit should define a standard machine-readable health endpoint/convention so hosts (incl. Gombit Cloud) do not require each app to invent custom health semantics (gombit-cloud DESIGN.md §24).

- [ ] Standard health endpoint (e.g. `GET /healthz`)
- [ ] Room for future readiness/liveness split
- [ ] Documented as a framework contract

**Blocks:** gombit-dev/gombit-cloud C0 health gate.

---
**Reconciled per gombit process.** Now backlog item **HOST-2** under `post-v0.1`, gated by **ADR-015**.

**Finding:** Gombit *already* serves `GET /livez` and `GET /readyz` (raw Gin, `framework/app.go`, D10-shaped). So this is **not** a from-scratch endpoint. HOST-2 = (a) document `/livez`+`/readyz` as a stable operational contract (`docs/health.md`), and (b) make `/readyz` **meaningful** — 200 only when datastore reachable + `OnStart` hooks done, else **503** with a D10 `not_ready` error (today it always returns 200, so it can't gate traffic).

**Supersedes Cloud DESIGN.md §24 `/healthz`:** the host consumes the existing `/livez`+`/readyz`; Gombit does not grow a `/healthz` alias. The corresponding gombit-cloud C0 issue should be updated to poll `/readyz`. deps: HOST-0, M1.

### LINKED ISSUE #285: [HOST-0] ADR-015: host/deployment contracts
Lock the shape of three framework-level **host/deployment contracts** so their implementation issues are unblocked, following the ADR-0-gates-implementation precedent (ADR-013 → ADMIN-1..3).

ADR: `docs/adr/015-host-deployment-contracts.md` (**Proposed**). Build-plan track: **HOST** (`docs/GOMBIT_BUILD_PLAN.md` §4, `post-v0.1`).

Covers:
- **HOST-1** #282 — machine-readable application contract (projected from declared config; no repo-trivia inference)
- **HOST-2** #283 — health convention (`/livez`+`/readyz`; make `/readyz` real; supersedes Cloud `/healthz`)
- **HOST-3** #284 — migration safety manifest + verifier (over Atlas SQL; anti-tamper hash; host owns the approval gate)

Honors locked constraints without reopening them: §3.3 generate-vs-runtime, 6.2 no request-time reflection / no repo-trivia inference, C1/OpenAPI (health stays raw-Gin, out of spec), D4 (no drift checksums), D10 envelopes, ADR-012 (no new migration engine). New scope for Gombit, **post-v0.1, off the v0.1 critical path, no M6 battery**. Host-neutral (MIT): a Gombit app stays deployable without any host (L18).

Driven by an external consumer — `gombit-cloud/docs/DESIGN.md` §9/§24/§29–§31 — which requires these contracts to live upstream in Gombit (§3.1/L4/L20). That design is rationale/context only, not a source of additional Gombit scope.

**AC:** ADR accepted (this issue closes on acceptance); HOST-1..3 reconciled to `[HOST-*]` titles under `post-v0.1` (done). **Implementation of HOST-1..3 does not start until this ADR is accepted.**

leo-aa88:
## Summary

Two things, one gate and its first slice:

- **HOST-0** — accept **ADR-015**, establishing a post-v0.1 `HOST` track for three host/deployment contracts (application contract, health convention, migration safety manifest) that a deployment host consumes to build, health-check, and safely migrate an ordinary *handwritten* Gombit app. Adds HOST-0..3 to build plan §4. Host-neutral (MIT), off the v0.1 critical path, no M6 battery.
- **HOST-2** — implement the health convention. `/readyz` was a stub that always returned `200` and could not gate traffic. It now returns `200 {"data":{"status":"ready"}}` only when the app is **not draining** and the configured **datastore answers a bounded ping**; otherwise `503` with a D10 `not_ready` envelope whose message is a **fixed reason** (never the raw datastore error — it can embed a DSN, and the probe is unauthenticated). `framework.WithShutdownDrainDelay` keeps the server accepting during shutdown so a host observes the 503 and deregisters before connections are refused. `/livez` unchanged. Both stay raw Gin, out of OpenAPI. Supersedes Cloud DESIGN.md §24 `/healthz`.

Closes #285
Closes #283

## Acceptance criteria

HOST-0:
- [x] ADR accepted; HOST-1..3 reconciled to `[HOST-*]` titles under `post-v0.1`

HOST-2:
- [x] `/readyz` returns 503 before readiness (draining / unreachable datastore) and 200 when ready — verified on **SQLite** (unit) and **Postgres + MySQL** (`framework/integration_test.go`, wired into the CI `database-postgres` / `database-mysql` jobs)
- [x] readiness check is not per-request-heavy (single bounded `PingContext`, 2s cap; tested)
- [x] 503 reason is fixed and leak-free; the cause is logged (tested against a DSN-bearing error)
- [x] shutdown drain window is host-visible over HTTP (tested)
- [x] `docs/health.md` documents the exact bodies and status codes

## Scope notes

- HOST-1 (application contract) and HOST-3 (migration safety manifest + verifier) are separate backlog items, not in this PR.
- ADR-015 lands with its first implementing slice (HOST-2), the same way ADR-013/ADMIN-0 preceded ADMIN-1.

## Validation

- [x] `go test ./...` — 44 packages ok (`-race` clean on `./framework`)
- [x] `go test -tags integration ./framework -framework.postgres-dsn … -framework.mysql-dsn …` (runs in CI's Postgres/MySQL jobs; skips locally without DSNs)
- [x] `gofmt` / `go vet` clean on changed files
- [x] Project `code-review` skill run against this diff

- [x] New behavior has tests; **DB-touching change now runs the SQLite + Postgres + MySQL matrix** via `framework/integration_test.go` + CI wiring
- [x] Stable feature ships docs (`docs/health.md`) and appears in every app (framework routes)
- [x] Extraction preserves contracts — `/livez` unchanged; only `/readyz` semantics changed as specified
- [ ] Generators — N/A
- [ ] OpenAPI/TS client — N/A (probes intentionally out of OpenAPI)
- [x] No secrets in frontend — N/A; and the 503 reason is explicitly scrubbed of datastore/DSN detail
- [x] Scope stays in `post-v0.1`; no M6 battery
- [x] Links issues and lists satisfied AC


framework/readiness_test.go
```diff
@@ -0,0 +1,115 @@
...
+	if body.Data.Status != "ready" {
+		t.Fatalf("GET /readyz data.status = %q, want %q", body.Data.Status, "ready")
+	}
+}
+
+// An attached but unreachable datastore makes the app not ready: 503 with a D10
+// not_ready envelope. This is the signal a host gates traffic on (§23/§24).
+func TestReadyzUnreachableDatabaseReturns503(t *testing.T) {
+	db, err := database.Open(config.DatabaseConfig{
+		Driver: config.DatabaseDriverSQLite,
```
leo-aa88: **[suggestion]** This is the only not-ready datastore path in the suite, and it is SQLite `Close()` on an in-memory pool.

HOST-2 AC (build plan §4): `/readyz` ready/not-ready is "verified on SQLite + Postgres + MySQL". ADR-015 Consequences: "Tests cover ready/not-ready transitions on the DB matrix." Working agreement item 1: DB-touching changes pass that matrix. D12 already made the three drivers required and CI-gated.

`PingContext` is portable at the call site. That does not make this test representative:

- `sql: database is closed` is a local pool error. It is not "the configured datastore is unreachable" (server down, network timeout, auth, failover). The 2s `readinessTimeout` never fires here; a hung network ping is exactly where that bound matters, and it is untested.
- Default SQLite pool is `MaxOpenConns=1`. Postgres/MySQL default to 25. Per-probe `PingContext` contends for the only SQLite connection; that interaction cannot show up in this file.
- CI `database-postgres` / `database-mysql` already open real servers for `./database`, `./auth`, `./admin`. They will not execute anything in `framework/` unless a `-tags integration` test is added and wired into those jobs, the same way auth/admin did.

The PR restated HOST-2's AC without the matrix clause, then disclosed the skip as "by design of the operation". Closing #283 with that AC unsatisfied is the design. SQLite `Close()` passing does not prove the host contract on the drivers a host actually pings.

**Suggestion:** Add `framework` integration tests (reachable ping + a real unreachable — stop the server or blackhole the DSN after `Open`) on postgres and mysql, and invoke them from the existing CI service jobs. Until then, do not claim the HOST-2 AC is met.

framework/app.go
```diff
@@ -482,6 +488,10 @@ func (a *App) runStartHooks(ctx context.Context) error {
 }
 
 func (a *App) shutdown() error {
+	// Flip readiness to 503 before draining so a host deregisters the instance
+	// while in-flight requests finish (HOST-2 / ADR-015).
+	a.draining.Store(true)
```
leo-aa88, line 525: **[suggestion]** THE COMMENT SAYS: flip to 503 *before* drain so a host deregisters the instance while in-flight requests finish.

THE RUNTIME DOES: set a process-local flag, then immediately `server.Shutdown`, which closes the listener. A polling host's next `/readyz` is almost never this handler returning `{"error":{"code":"not_ready","message":"shutting down"}}`. It is connection refused.

That is not the protocol `docs/health.md` describes ("the probe flips to `503` *before* connections drain, so a host stops routing new requests to the instance while in-flight ones finish"). Connection refused also sheds traffic — for Kubernetes, Docker, and most LBs it is a failed probe — but it is not the D10 `not_ready` contract this PR just stabilized. There is no probe-period delay, so there is no reliable 503 window.

The tests do not catch this:

- `TestReadyzDrainingReturns503` pokes `draining.Store(true)` and calls `ServeHTTP`. No listener, no Shutdown, and despite the comment ("even with a healthy datastore") no datastore.
- `TestShutdownMarksDraining` asserts the flag *after* `RunContext` returns. An implementation that set the flag at the *end* of `shutdown()` would still pass.

The serve-error branch (`serverErr`) never sets `draining` at all.

**Suggestion:** Either implement a host-visible window (set the flag, wait at least one probe interval, *then* `Shutdown`) or stop documenting 503-on-shutdown as the host signal and state that shutdown is "flag + close listener; treat connection failure as not-ready". Either way, add a test that observes `/readyz` returning 503 *while the server is still accepting* after cancel, not just the atomic after exit.

leo-aa88 (commented):
## Summary

HOST-0 is a real ADR: it locks three host contracts without reopening Huma, Atlas, D4, D10, or ADR-013, and HOST-1..3 are already reconciled under `post-v0.1`. HOST-2 turns `/readyz` into an actual traffic gate (draining flag + bounded `PingContext` + D10 `not_ready`) instead of a stub `200`.

Two gaps keep this from LGTM.

1. HOST-2's written AC, ADR-015 Consequences, and working-agreement item 1 all require ready/not-ready on SQLite **and** Postgres **and** MySQL. The tests never leave in-memory SQLite. The PR drops that clause from its own AC list and calls the skip intentional. CI already has the postgres/mysql jobs; they will not run any `framework` readiness test as written.

2. The shutdown half of the host contract is a process-local flag, not a host-visible 503 window. `draining.Store(true)` is immediately followed by `server.Shutdown`, which closes the listener. A polling host almost never sees `{"error":{"code":"not_ready","message":"shutting down"}}`; it sees connection refused. The tests would still pass if the flag were set *after* Shutdown.

The happy-path runtime is fine. The comments and the closed issue AC currently describe a stricter contract than the tests or the shutdown path actually provide.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

leo-aa88 (commented):
## Summary

`f993ba4` actually answers the previous review. Fixed 503 reason (no DSN on the unauthenticated probe), Postgres/MySQL integration tests wired into the existing CI jobs, host-visible drain window as an explicit `WithShutdownDrainDelay` (default 0 is now documented honestly), timeout test, CHANGELOG for `data.status` `ok`→`ready`.

CI is red. Lint fails G101 on the leak-test secret. Every job after lint is skipped, so the matrix this commit added has never run. That is the merge bar, not a style nit — the rest of this repo already writes `#nosec G101 -- fake …` on the same class of fixture.

Do not merge until lint is green and the postgres/mysql `./framework` integration tests have actually executed.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

framework/readiness_test.go
```diff
@@ -0,0 +1,190 @@
...
+		t.Fatalf("GET /readyz data.status = %q, want %q", body.Data.Status, "ready")
+	}
+}
+
+// A failing datastore probe makes the app not ready: 503 with a D10 not_ready
+// envelope. Critically, the raw probe error — which may embed a DSN — must NOT
+// reach the public probe; only the fixed reason does. The cause is logged.
+func TestReadyzUnreachableDatastoreHides503Cause(t *testing.T) {
+	app := newTestApp(t)
+	const secret = "dial tcp 10.0.0.5:5432: host=10.0.0.5 user=admin password=hunter2 dbname=prod"
```
leo-aa88: **[bug]** CI is red on this line: `G101: Potential hardcoded credentials (gosec)`.

The test that proves `/readyz` does not leak `password=` is itself the credential gosec flags. Lint failed; Build/Test/Database (postgres|mysql) were skipped. The matrix this commit added has not run.

This repo already has a convention for fake secrets, e.g. `config/redact_test.go`:

```go
dsn := "host=localhost user=gombit password=s3cret dbname=app sslmode=disable" // #nosec G101 -- fake local test DSN.
```

**Suggestion:** Put `#nosec G101 -- fake probe-error fixture, not a credential` on this const (same pattern as every other G101 in-tree). Then let CI actually run `go test -tags integration ./framework` on postgres and mysql.

framework/app.go
```diff
@@ -482,11 +520,23 @@ func (a *App) runStartHooks(ctx context.Context) error {
...
 	server := a.server
 	timeout := a.shutdownTimeout
+	drainDelay := a.shutdownDrainDelay
 	a.mu.RUnlock()
 
+	// Keep serving for the drain delay so a host polling /readyz sees the 503
+	// and stops routing before the listener closes.
+	if drainDelay > 0 {
+		timer := time.NewTimer(drainDelay)
+		<-timer.C
```
leo-aa88, line 537: **[suggestion]** This sleep is not part of `shutdownTimeout`. Total time from SIGTERM to process exit is `shutdownDrainDelay + shutdownTimeout` (default timeout 10s).

`WithShutdownDrainDelay(30*time.Second)` on a Kubernetes pod with `terminationGracePeriodSeconds: 30` gets SIGKILL during `Shutdown`, not a clean drain. The option doc says "a few seconds" but nothing enforces it.

Not a blocker. Either bound the delay (`delay >= shutdownTimeout` rejected at `New`, or the wait is `min(delay, remaining budget)`) or state in the option godoc that the delay is **in addition to** the shutdown timeout and must fit in the orchestrator's grace period.

leo-aa88 (commented):
LGTM

PR #287, opened 2026-09-05
[HOST-1] Machine-readable application contract (gombit contract app)

### LINKED ISSUE #282: [HOST-1] Machine-readable application contract
Gombit must expose a stable, versioned, machine-readable application description that both Forge and Cloud consume. Per gombit-cloud DESIGN.md §3.1/§10/L4 this contract must live in Gombit, not privately in Cloud.

Shape (illustrative, gombit-cloud DESIGN.md §9):
```json
{ "contract_version": 1, "framework": {"name":"gombit","version":"0.4.0"},
  "build": {"command":"gombit build --embed","artifact":"./dist/app"},
  "runtime": {"http_port":8080,"health_path":"/healthz"},
  "database": {"required":true,"driver":"postgres"},
  "migrations": {"path":"database/migrations"} }
```

- [ ] Define + version the schema (`contract_version`)
- [ ] Emit/expose it from a Gombit project
- [ ] Explicit-failure semantics on incompatible versions

**Blocks:** gombit-dev/gombit-cloud C0 gate (app detection).

---
**Reconciled per gombit process (AGENTS.md / build plan §4).** Now backlog item **HOST-1** under `post-v0.1`, gated by **ADR-015** (`docs/adr/015-host-deployment-contracts.md`). Contract is **projected from declared config** (`gombit.yaml` + resolved framework version, REL-9) — never inferred from repo trivia (§10, principle 6.2). Provisional surface `gombit contract app`. deps: HOST-0, REL-9, M4-1.

leo-aa88:
## Summary

Implements **HOST-1** (ADR-015): `gombit contract app` emits a stable, versioned, machine-readable description of a Gombit app that a deployment host (Gombit Cloud, a container platform, CI) reads to **build, health-check, and migrate** it — without inferring anything from the source tree.

```json
{
  "contract_version": 1,
  "framework": { "name": "gombit", "version": "v0.5.0" },
  "build": { "command": "gombit build --embed", "artifact": "bin/server" },
  "runtime": { "http_port": 8080, "health": { "live": "/livez", "ready": "/readyz" } },
  "database": { "required": true, "driver": "sqlite" },
  "migrations": { "path": "database/migrations" }
}
```

- New `appcontract` package: the versioned `Contract` schema + a `Project` built from declared inputs only (decoupled from `config` so a host can reuse the type).
- `FrameworkVersion` reads the version from the project's `go.mod` require directive — offline, deterministic. A **local-path** replace is surfaced as unresolved; a **version-to-version / published-fork** replace resolves to its target.
- Fails loudly on a non-Gombit app or an unresolvable framework version (DESIGN.md §55).

Closes #282

## Review feedback addressed (leo-aa88)

1. **`database.required` was a JWT proxy (bug).** Now a declared config field — `Config.Database.Required` (`GOMBIT_DATABASE_REQUIRED`, default `true`, shown in `gombit config show`). A postgres CRUD app without auth correctly reports `required: true`; an app that needs no database opts out explicitly. No longer derived from `Auth.Enabled()`.
2. **Blanket replace failure.** The go.mod parser now distinguishes a local filesystem replace (unresolved) from a version replace a host can pin (reports the target version).
3. **`--dir` divergence.** `contract app` now reads both `go.mod` and config `.env` from `--dir`, so a monorepo can't mix one app's go.mod with another's config; process `GOMBIT_*` env still applies. `--out` is resolved before the chdir.

## Acceptance criteria (HOST-1)

- [x] Emits valid JSON whose values come from declared config + `go.mod`
- [x] Unresolvable framework version fails loudly; version-replace resolves
- [x] `database.required` is a declared value, not inferred
- [x] Schema + `contract_version` covered by tests; no tree-walking (§10, 6.2)
- [x] Docs (`app-contract.md`, `cli.md`, `config.md`, index) + CHANGELOG

## Validation

- [x] `go test ./...` — all packages ok
- [x] `golangci-lint v2.12.2` — 0 issues on changed packages
- [x] `gofmt` / `go vet` clean
- [x] Manual: version-replace resolves to the target; `.env` in `--dir` is honored; `database.required` reflects config

- [x] New behavior has tests (`appcontract`, `cli`, `config`). Not DB-touching (pure projection + go.mod parse), so no DB matrix applies.
- [x] Stable feature ships docs; runs in any generated app
- [ ] Generators — N/A
- [ ] OpenAPI/TS client — N/A (CLI projection, not an HTTP route)
- [x] No secrets in frontend — N/A
- [x] Scope stays in `post-v0.1`; no M6 battery
- [x] Links its issue and lists satisfied AC

### DIFF
--- a/appcontract/modversion.go
+++ b/appcontract/modversion.go
@@ -1,0 +1,124 @@
+package appcontract
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// FrameworkModulePath is the module a Gombit application requires.
+const FrameworkModulePath = "github.com/gombit-dev/gombit"
+
+// ErrFrameworkVersionUnresolved is returned when go.mod names the framework but
+// its version cannot be reported as a resolvable module version — a local
+// replace directive (a framework dev checkout) being the common case. A host
+// cannot pin such a build, so this is surfaced rather than guessed.
+var ErrFrameworkVersionUnresolved = errors.New("appcontract: framework version is unresolved (replaced or local)")
+
+// FrameworkVersion reads the version of the Gombit framework the app in workDir
+// builds against, straight from its go.mod require directive. It deliberately
+// parses the *declared* dependency rather than running the go toolchain, so it
+// is offline and deterministic.
+//
+// A replace directive targeting the framework module makes the version
+// unresolvable for a host; that returns ErrFrameworkVersionUnresolved. A go.mod
+// that does not require the framework at all is an error — it is not a Gombit
+// app.
+func FrameworkVersion(workDir string) (string, error) {
+	path := filepath.Join(workDir, "go.mod")
+	data, err := os.ReadFile(path) // #nosec G304 -- go.mod at a caller-provided project dir
+	if err != nil {
+		return "", fmt.Errorf("appcontract: read %s: %w", path, err)
+	}
+	return frameworkVersionFromModfile(string(data))
+}
+
+// frameworkVersionFromModfile extracts the framework require version from go.mod
+// content. Split out for testing without touching the filesystem.
+func frameworkVersionFromModfile(content string) (string, error) {
+	version := ""
+	replaced := false
+
+	inRequireBlock := false
+	inReplaceBlock := false
+
+	scanner := bufio.NewScanner(strings.NewReader(content))
+	for scanner.Scan() {
+		line := stripComment(scanner.Text())
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" {
+			continue
+		}
+
+		switch {
+		case inRequireBlock:
+			if trimmed == ")" {
+				inRequireBlock = false
+				continue
+			}
+			if v, ok := requireVersion(trimmed); ok {
+				version = v
+			}
+		case inReplaceBlock:
+			if trimmed == ")" {
+				inReplaceBlock = false
+				continue
+			}
+			if isFrameworkReplace(trimmed) {
+				replaced = true
+			}
+		case strings.HasPrefix(trimmed, "require ("):
+			inRequireBlock = true
+		case strings.HasPrefix(trimmed, "replace ("):
+			inReplaceBlock = true
+		case strings.HasPrefix(trimmed, "require "):
+			if v, ok := requireVersion(strings.TrimSpace(strings.TrimPrefix(trimmed, "require"))); ok {
+				version = v
+			}
+		case strings.HasPrefix(trimmed, "replace "):
+			if isFrameworkReplace(strings.TrimSpace(strings.TrimPrefix(trimmed, "replace"))) {
+				replaced = true
+			}
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("appcontract: scan go.mod: %w", err)
+	}
+
+	if replaced {
+		return "", ErrFrameworkVersionUnresolved
+	}
+	if version == "" {
+		return "", fmt.Errorf("appcontract: go.mod does not require %s", FrameworkModulePath)
+	}
+	return version, nil
+}
+
+// requireVersion returns the version if entry is a require line for the
+// framework module, e.g. "github.com/gombit-dev/gombit v0.5.0".
+func requireVersion(entry string) (string, bool) {
+	fields := strings.Fields(entry)
+	if len(fields) >= 2 && fields[0] == FrameworkModulePath {
+		return fields[1], true
+	}
+	return "", false
+}
+
+// isFrameworkReplace reports whether entry replaces the framework module, e.g.
+// "github.com/gombit-dev/gombit => ../gombit".
+func isFrameworkReplace(entry string) bool {
+	fields := strings.Fields(entry)
+	return len(fields) >= 1 && fields[0] == FrameworkModulePath
+}
+
+// stripComment removes a trailing // comment (e.g. "// indirect") outside of
+// any string, which go.mod lines never contain.
+func stripComment(line string) string {
+	if i := strings.Index(line, "//"); i >= 0 {
+		return line[:i]
+	}
+	return line
+}
--- a/appcontract/modversion.go
+++ b/appcontract/modversion.go
@@ -31,10 +31,11 @@ var semverPattern = regexp.MustCompile(
 // parses the *declared* dependency rather than running the go toolchain, so it
 // is offline and deterministic.
 //
-// A replace directive targeting the framework module makes the version
-// unresolvable for a host; that returns ErrFrameworkVersionUnresolved. A go.mod
-// that does not require the framework at all is an error — it is not a Gombit
-// app.
+// A replace of the framework module wins over the require: a version-to-version
+// replace (or a published fork with a version) reports its target version,
+// while a local filesystem replace is unresolvable for a host and returns
+// ErrFrameworkVersionUnresolved. A go.mod that does not require the framework at
+// all is an error — it is not a Gombit app.
 func FrameworkVersion(workDir string) (string, error) {
 	path := filepath.Join(workDir, "go.mod")
 	data, err := os.ReadFile(path) // #nosec G304 -- go.mod at a caller-provided project dir
--- a/appcontract/modversion.go
+++ b/appcontract/modversion.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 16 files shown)


cli/contract.go
```diff
@@ -0,0 +1,107 @@
...
+				"framework version — build against a published %s release", err, appcontract.FrameworkModulePath)
+		}
+		return fmt.Errorf("gombit contract app: %w", err)
+	}
+
+	contract, err := appcontract.Project(appcontract.Inputs{
+		FrameworkVersion: version,
+		HTTPAddr:         cfg.HTTP.Addr,
+		DatabaseDriver:   string(cfg.Database.Driver),
+		DatabaseRequired: cfg.Auth.Enabled(),
```
leo-aa88: **[bug]** `database.required` is `cfg.Auth.Enabled()`.

THE FIELD SAYS: this app needs a datastore.
THE RUNTIME DOES: JWT secret is set, so `framework.New` refuses to boot without `WithDatabase`.

Those are not the same predicate. `examples/database` and any generated CRUD app without `GOMBIT_JWT_SECRET` still `database.Open`. The contract they would emit is `{required: false, driver: "postgres"}` plus `migrations.path: database/migrations`. A host that provisions on `required` does not start postgres. The app then dies on Open.

This is not a declared-config projection. Gombit has no "this app uses a database" field, so the PR invented a proxy. Auth is the wrong one. The contradictory state (`required: false` + driver + migrations path) is representable and is the default.

Flagging it in the PR body does not make the host contract true.

**Suggestion:** Do not bind `required` to JWT. Least-wrong without a new config field: treat a configured driver as required (today that is always true — then say so, or drop the field until it can vary). If the intent is "framework will not boot without a DB", name the field that, and do not tell a host it is the provision signal. A dedicated `config`/`gombit.yaml` bit is the actual declared source the ADR asked for.

leo-aa88 (commented):
## Summary

HOST-1's shape is the right one: CLI projection, `contract_version` independent of the framework, no HTTP route, no tree-walk. Reading the app's `go.mod` require (not `cli.Version`) is the correct declared source.

The host-facing fields then lie in two places, and `--dir` does not mean what the docs show.

1. `database.required` is `Auth.Enabled()`. A postgres CRUD app without JWT emits `required: false` and a migrations path. A host that provisions on `required` skips the datastore. That is not a declared-config projection; it is a proxy for JWT.
2. Docs say a **local** `replace` is unresolved. The parser fails on **any** replace of the framework module, including version-to-version replaces a host can pin. The go.mod scanner is a hand-rolled `bufio` loop, which is why that distinction does not exist.
3. `docs/app-contract.md` and `cli.md` present `--dir` as the project. Config and `.env` still come from cwd/process env. `--dir apps/api` from a monorepo root mixes that app's go.mod with the root's `GOMBIT_HTTP_ADDR` / driver.

CI is green. The tests lock the JWT mapping and the blanket replace failure. They do not exercise the host contract.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

appcontract/modversion.go
```diff
@@ -0,0 +1,124 @@
...
+			if isFrameworkReplace(strings.TrimSpace(strings.TrimPrefix(trimmed, "replace"))) {
+				replaced = true
+			}
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("appcontract: scan go.mod: %w", err)
+	}
+
+	if replaced {
```
leo-aa88: **[suggestion]** THE DOCS SAY: a replace that points the framework at a **local checkout** is unresolved.

THE PARSER DOES: any `replace` whose left-hand path is `github.com/gombit-dev/gombit`.

```go
func isFrameworkReplace(entry string) bool {
    fields := strings.Fields(entry)
    return len(fields) >= 1 && fields[0] == FrameworkModulePath
}
```

This go.mod is a pin a host can resolve, and it still errors:

```
require github.com/gombit-dev/gombit v0.5.0
replace github.com/gombit-dev/gombit => github.com/gombit-dev/gombit v0.5.1
```

The tests only cover `=> ../gombit`. They prove the implementation agrees with itself.

The cause is the hand-rolled `bufio` scanner. `golang.org/x/mod/modfile` already distinguishes a directory replace from a module-version replace. Use it. Fail only when the replacement is a filesystem path (the case the docs and the error string actually describe).

docs/app-contract.md
```diff
@@ -0,0 +1,65 @@
...
+Gombit app that a deployment host (a container platform, CI, or
+[Gombit Cloud](adr/015-host-deployment-contracts.md)) reads to learn how to
+**build, health-check, and migrate** the app — without inferring anything from
+the source tree. It is the HOST-1 contract from
+[ADR-015](adr/015-host-deployment-contracts.md).
+
+```bash
+gombit contract app          # print the contract to stdout
+gombit contract app --out app.json
+gombit contract app --dir ./path/to/project
```
leo-aa88, line 13: **[suggestion]** This example tells a host `--dir` selects the project. `cli.md` says the same: "`--dir` selects the project directory."

`runContractApp` uses `--dir` only for `FrameworkVersion(dir)` (the go.mod). `LoadConfig()` still reads process env and `.env` from **cwd**. Flag help admits this; the docs do not.

Concrete miss: from a monorepo root, `gombit contract app --dir apps/api` emits `apps/api`'s require line mixed with the root's `GOMBIT_HTTP_ADDR` / `GOMBIT_DATABASE_DRIVER`. The host health-checks the wrong port.

**Suggestion:** Either load the project's `.env` (and any future file config) from `--dir`, or remove `--dir` until it means the project. Do not document a project-directory flag that only relocates `go.mod`.

cli/contract.go
```diff
@@ -0,0 +1,143 @@
...
+		abs, err := filepath.Abs(out)
+		if err != nil {
+			return fmt.Errorf("gombit contract app: resolve --out: %w", err)
+		}
+		out = abs
+	}
+
+	// Read config (.env) and go.mod from the same project directory so the
+	// contract cannot mix one app's go.mod with another's config.
+	restore, err := chdir(dir)
```
leo-aa88: **[suggestion]** `--dir` now means the project. Implementing that with `os.Chdir` reintroduces a process-global.

`go test ./...` runs packages in parallel. These tests call `runContractApp` with a `t.TempDir()`, which chdirs the **whole process** into that temp while `./database`, `./framework`, etc. still resolve relative paths against cwd. That is a flake generator. `loadDotEnv` then `os.Setenv`s from the project's `.env` and never unsets them, so the leak survives `restore()`.

The production CLI is single-shot so it happens to work. The tests and any in-process reuse do not.

`LoadConfig` is stubbed in every `runContractApp` test, so the `.env`-from-`--dir` path this chdir exists to implement is **untested**.

**Suggestion:** Do not chdir. Open `.env` at `filepath.Join(dir, ".env")` (or teach `loadDotEnv` a directory), pass `dir` to `FrameworkVersion(dir)`, and add one test that writes a `.env` under `--dir` and asserts the projected port/driver without stubbing `LoadConfig`.

docs/app-contract.md
```diff
@@ -0,0 +1,68 @@
...
+```json
+{
+  "contract_version": 1,
+  "framework": { "name": "gombit", "version": "v0.5.0" },
+  "build": { "command": "gombit build --embed", "artifact": "bin/server" },
+  "runtime": {
+    "http_port": 8080,
+    "health": { "live": "/livez", "ready": "/readyz" }
+  },
+  "database": { "required": false, "driver": "sqlite" },
```
leo-aa88: **[suggestion]** Canonical example still emits `"required": false`.

Default is now `true` (`DefaultFor`, `GOMBIT_DATABASE_REQUIRED`, the table three lines below). This JSON is the leftover from when `required` was `Auth.Enabled()` (default false). A host copying the example will think sqlite apps report `required: false` by default. They do not.

Show `true`, or label this as an explicit opt-out.

leo-aa88 (commented):
## Summary

`b55785d` actually answers the previous review. `database.required` is a declared `Config.Database.Required` (default true). Local-path replace is unresolved; version-to-version replace reports the target (tested). `--dir` now aims both go.mod and `.env` at the same project.

What remains is how `--dir` was implemented, plus two leftovers that still contradict the new contract.

1. `os.Chdir` is process-global. `go test ./...` runs packages in parallel. These tests chdir into a temp dir while other packages use cwd. `loadDotEnv` also `os.Setenv`s from that `.env` and never unsets. Load `.env` from `filepath.Join(dir, ".env")`; pass `dir` to `FrameworkVersion`. Do not mutate the process.
2. The canonical example in `docs/app-contract.md` still shows `required: false`. Default is now `true`. That is the old JWT-proxy example, left in place.
3. Exported `FrameworkVersion` godoc still says any replace is unresolved. The function no longer does that.

Lint and Test are green. Previous blockers are closed. Do not treat `--dir` via `Chdir` as done.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

leo-aa88 (commented):
LGTM

PR #290, opened 2026-09-08
fix: scope security headers by response kind (PERF-9, #267)

### LINKED ISSUE #267: [PERF-9] Security headers push every response past 8 headers, costing 5 allocs/op of map growth; scope browser policy to HTML routes
## Summary

`securityHeadersMiddleware` ([framework/http_middleware.go#L108](https://github.com/gombit-dev/gombit/blob/main/framework/http_middleware.go#L108)) assigns six shared, pre-allocated slices and allocates nothing itself, yet the ablation attributes **5 allocs/op and ≈0.7 µs** to it on every scenario. The cause is header count: request-id + trace-id + six security headers + Content-Type = 9 entries, one past the 8-slot group of Go's swiss map. The handler header map grows (≈3 allocs) and the `Header.Clone` net/http performs on `WriteHeader` grows too (≈2 allocs). Every API response pays this.

Separately, `Content-Security-Policy: default-src 'self'`, `X-Frame-Options`, and `X-Download-Options` are browser-rendering policies that do nothing useful on a JSON response to `GET /api/projects`. `X-Download-Options` is an IE-only header no maintained browser honors.

## Evidence (measured)

- Cumulative ablation: +security_headers = +5 allocs (POST 45→50, plaintext 25→30).
- `alloc_objects` profile: all 7 `Header.Set` allocs come from only 3 `Set` calls (request-id, trace-id, Content-Type); the surplus 4 is map growth triggered by the 9th key. `Header.Clone` = 5 allocs/op.
- Prototype dropping only `X-Download-Options` (response back to exactly 8 headers): POST 52→47, plaintext 27→22 allocs, plaintext ≈2.8 µs → 2.1 µs.
- Prototype with an API-only set (`X-Content-Type-Options`, `Strict-Transport-Security`, `Content-Security-Policy: default-src 'none'; frame-ancestors 'none'`): same alloc count as the 8-header set, so the win is the map threshold, not the header bytes.

## Proposed fix

1. Drop `X-Download-Options` unconditionally (dead header). Zero behavior risk, restores the ≤8 threshold for the common API response.
2. Split the header set by response kind:
   - API/JSON default: `X-Content-Type-Options: nosniff`, `Strict-Transport-Security` (production), `Content-Security-Policy: default-src 'none'; frame-ancestors 'none'` (the OWASP REST minimum; `frame-ancestors 'none'` subsumes `X-Frame-Options: DENY`).
   - HTML routes (`/admin/*`, `/docs/*`, embedded SPA, `applySPAContentSecurityPolicy`): the full browser policy as today, including `Referrer-Policy` and `X-Frame-Options` for legacy UAs.
3. Document the header budget: a route that adds `Set-Cookie`, `Cache-Control`, `Link`, etc. will cross 8 again; keep the default API response at ≤6 headers so there is headroom. Make the shared-slice invariant test (`TestSecurityHeaderSharedValueContract`) cover the new values.

## Acceptance criteria

- [ ] Default JSON response carries ≤ 8 headers in total with correlation IDs and Content-Type included; ablation shows the security_headers layer at 0 allocs/op.
- [ ] HTML/admin/docs/SPA responses keep the full browser policy; existing security tests updated rather than deleted, with a test per response kind.
- [ ] `docs/` security page lists which headers apply to which response kind and why.
- [ ] README/CHANGELOG note the removed `X-Download-Options` and the CSP change for API responses.

## Scope notes

Runtime + security. Highest allocs-per-risk ratio in #264; do this first. Build plan §5 applies.

leo-aa88:
## Summary

Scopes the runtime security headers **by response kind** (issue #267 / PERF-9): API/JSON responses get a strict, minimal policy while HTML responses (the SPA, the admin SPA, and Huma's `/docs`) keep the full browser policy. This holds the common JSON response under Go's 8-header swiss-map threshold, removing ~5 allocs/op of header-map growth from every API response, and drops the dead IE8-only `X-Download-Options` header.

Before this change `securityHeadersMiddleware` set one fixed policy (CSP `default-src 'self'` + `Referrer-Policy` + `X-Frame-Options` + `X-Content-Type-Options` + HSTS) on **every** response — the ninth header that tipped API responses over the swiss-map group boundary.

### What changed

- **`framework/http_middleware.go`** — `securityHeadersMiddleware(includeHSTS, docsEnabled bool)`:
  - **API/JSON (default, hot path):** `Content-Security-Policy: default-src 'none'; frame-ancestors 'none'` + `X-Content-Type-Options: nosniff` (+ HSTS in production). No `X-Frame-Options` or `Referrer-Policy` — `frame-ancestors 'none'` subsumes the former, and a `default-src 'none'` document needs neither. ≤ 6 headers with correlation IDs + Content-Type, so the layer allocates nothing.
  - **`/docs` (only when `DocsEnabled`):** the browser hardening headers Huma's docs renderer omits (`Referrer-Policy`, `X-Frame-Options`); Huma sets its own CSP for the Swagger UI assets. Classification is gated on `DocsEnabled`, so with docs disabled a `/docs` request is an ordinary not-found **API** response and gets the API policy — the response kind follows what is actually served, not the URL prefix.
- **`framework/embed.go` / `framework/adminui.go`** — HTML documents the framework serves through its own handlers (embedded SPA `index.html`, admin SPA) start from the API baseline and are promoted to the full browser policy (SPA CSP + `Referrer-Policy` + `X-Frame-Options`) via `applyBrowserSecurityHeaders`.
- **`X-Download-Options` removed** — IE8-only, honored by no supported browser, and part of what pushed the API response over the header threshold.
- **Tests** — a test per response kind (`TestDefaultRouterAddsAPISecurityHeaders`, `TestBrowserResponseKindsCarryFullPolicy` for SPA/admin/docs, `TestDisabledDocsPathGetsAPIPolicy` for the disabled-docs adversarial case), a header-budget invariant (`TestDefaultJSONResponseStaysUnderHeaderBudget`), an allocation-regression test proving the layer adds 0 allocs/op (`TestSecurityHeadersLayerAllocatesNothing`), and the shared-value contract re-anchored on the API CSP header.
- **Docs** — new [`docs/security.md`](docs/security.md) with the per-response-kind header table + rationale; `docs/build.md`, `docs/openapi.md`, the docs index, and `CHANGELOG.md` updated.

Fixes #267.

## Validation

- [x] `go test ./...`

- [x] New behavior has tests (a test per response kind + header-budget + 0-alloc regression tests)
- [x] Stable features ship docs — new `docs/security.md`, CHANGELOG note for the removed header and the CSP change
- [x] Extraction preserves contracts — HTML responses keep the full browser policy; only the API response kind is narrowed
- [x] Scope stays inside this issue's milestone — runtime security headers only
- [x] This PR links its issue (#267)

### DIFF
--- a/framework/http_middleware.go
+++ b/framework/http_middleware.go
@@ -101,10 +101,24 @@ var (
 	referrerPolicyValue     = []string{"strict-origin-when-cross-origin"}
 	hstsHeaderValue         = []string{"max-age=315360000; includeSubDomains"}
 	contentTypeOptionsValue = []string{"nosniff"}
-	downloadOptionsValue    = []string{"noopen"}
 	frameOptionsValue       = []string{"DENY"}
 )
 
+// securityHeadersMiddleware sets the framework's fixed set of security
+// headers on every response (API and HTML/SPA alike). X-Download-Options
+// ("noopen") is IE8-only guidance for the long-retired IE8 downloads
+// behavior; no supported browser honors it, so it is not set at all rather
+// than carried forward as dead weight on every response (issue #267). Every
+// other header here is still meaningful in current browsers:
+//   - Content-Security-Policy and X-Frame-Options are the load-bearing
+//     clickjacking/injection defenses.
+//   - Referrer-Policy and X-Content-Type-Options are cheap, still-honored
+//     hardening with no compatibility downside.
+//
+// HTML responses (the embedded SPA) override Content-Security-Policy via
+// applySPAContentSecurityPolicy (see embed.go) to allow the styling the SPA
+// needs; JSON API responses keep the strict default-src 'self' policy set
+// here. See docs/security.md for the full per-response-type header table.
 func securityHeadersMiddleware(includeHSTS bool) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		header := c.Writer.Header()
@@ -114,7 +128,6 @@ func securityHeadersMiddleware(includeHSTS bool) gin.HandlerFunc {
 			header["Strict-Transport-Security"] = hstsHeaderValue
 		}
 		header["X-Content-Type-Options"] = contentTypeOptionsValue
-		header["X-Download-Options"] = downloadOptionsValue
 		header["X-Frame-Options"] = frameOptionsValue
 		c.Next()
 	}
--- a/framework/http_middleware.go
+++ b/framework/http_middleware.go
@@ -163,14 +163,15 @@ func securityHeadersMiddleware(includeHSTS, docsEnabled bool) gin.HandlerFunc {
 	}
 }
 
-// isDocsPath reports whether urlPath is Huma's interactive docs UI. Docs is an
-// HTML document served by Huma (contract.DocsPath, root-mounted), so — when it
-// is enabled — its response kind is decided here by path rather than by an
-// override in a framework handler. The caller gates this on docsEnabled: a
-// /docs path with docs disabled has no docs handler behind it and is not an
-// HTML response.
+// isDocsPath reports whether urlPath is Huma's interactive docs UI. Huma
+// registers exactly one route, contract.DocsPath ("/docs") — not the "/docs/"
+// subtree — so only that exact path is an HTML document. A descendant like
+// /docs/not-a-route is served by no handler (404) and is an ordinary API
+// response, not HTML; matching the whole prefix would hand those 404s the
+// browser policy. The caller additionally gates this on docsEnabled: with docs
+// disabled the route is not registered at all.
 func isDocsPath(urlPath string) bool {
-	return urlPath == contract.DocsPath || strings.HasPrefix(urlPath, contract.DocsPath+"/")
+	return urlPath == contract.DocsPath
 }
 
 // httpMetrics accumulates request metrics without locking the request path.
--- a/framework/http_middleware.go
+++ b/framework/http_middleware.go
@@ -129,11 +129,15 @@ var (
 //     response with correlation IDs and Content-Type at <= 6 headers, one
 //     under the swiss-map 8-slot group boundary, so the layer allocates
 //     nothing (the map and net/http's WriteHeader Header.Clone no longer grow).
-//   - Interactive docs (/docs): the full browser policy, applied here by path
-//     because Huma owns that handler. Huma's docs renderer sets its own
-//     Content-Security-Policy (it must allow the Swagger UI assets), so it
-//     overrides the CSP set here; what this branch contributes that Huma does
-//     not is Referrer-Policy and the legacy X-Frame-Options: DENY.
+//   - Interactive docs (/docs), only when docs are enabled: the full browser
+//     policy, applied here by path because Huma owns that handler. Huma's docs
+//     renderer sets its own Content-Security-Policy (it must allow the Swagger
+//     UI assets), so it overrides the CSP set here; what this branch
+//     contributes that Huma does not is Referrer-Policy and the legacy
+//     X-Frame-Options: DENY. When docs are disabled (the production default)
+//     no /docs route exists, so a /docs request is an ordinary API/not-found
+//     response and gets the API policy — the classification follows the
+//     response that will actually be served, not the URL alone.
 //
 // HTML documents the framework serves through its own handlers — the admin SPA
 // and the embedded frontend — start from the API baseline set here and then
@@ -141,14 +145,14 @@ var (
 // and adminui.go), which is where the richer SPA CSP, Referrer-Policy, and
 // X-Frame-Options are added. X-Download-Options ("noopen") is IE8-only and set
 // nowhere: no supported browser honors it (issue #267).
-func securityHeadersMiddleware(includeHSTS bool) gin.HandlerFunc {
+func securityHeadersMiddleware(includeHSTS, docsEnabled bool) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		header := c.Writer.Header()
 		header["X-Content-Type-Options"] = contentTypeOptionsValue
 		if includeHSTS {
 			header["Strict-Transport-Security"] = hstsHeaderValue
 		}
-		if isDocsPath(c.Request.URL.Path) {
+		if docsEnabled && isDocsPath(c.Request.URL.Path) {
 			header["Content-Security-Policy"] = browserContentSecurityPolicyValue
 			header["Referrer-Policy"] = referrerPolicyValue
 			header["X-Frame-Options"] = frameOptionsValue
@@ -160,9 +164,11 @@ func securityHeadersMiddleware(includeHSTS bool) gin.HandlerFunc {
 }
 
 // isDocsPath reports whether urlPath is Huma's interactive docs UI. Docs is an
-// HTML document served by Huma (contract.DocsPath, root-mounted), so its
-// response kind is decided here by path rather than by an override in a
-// framework handler.
+// HTML document served by Huma (contract.DocsPath, root-mounted), so — when it
+// is enabled — its response kind is decided here by path rather than by an
+// override in a framework handler. The caller gates this on docsEnabled: a
+// /docs path with docs disabled has no docs handler behind it and is not an
+// HTML response.
 func isDocsPath(urlPath string) bool {
 	return urlPath == contract.DocsPath || strings.HasPrefix(urlPath, contract.DocsPath+"/")
 }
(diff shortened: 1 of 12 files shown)


framework/http_middleware.go
```diff
@@ -101,10 +101,24 @@ var (
 	referrerPolicyValue     = []string{"strict-origin-when-cross-origin"}
 	hstsHeaderValue         = []string{"max-age=315360000; includeSubDomains"}
 	contentTypeOptionsValue = []string{"nosniff"}
-	downloadOptionsValue    = []string{"noopen"}
 	frameOptionsValue       = []string{"DENY"}
 )
 
+// securityHeadersMiddleware sets the framework's fixed set of security
```
leo-aa88: **BLOCKING.** The linked issue requires different policies for API/JSON and HTML/admin/docs/SPA responses. This comment admits the implementation still sets one fixed header set on every response, and the code below keeps `default-src 'self'` for JSON instead of the required API policy `default-src 'none'; frame-ancestors 'none'`. Removing the IE8-only header is only the first step. Implement the response-kind boundary and test both policies before closing #267.

framework/http_middleware.go
```diff
@@ -101,10 +101,24 @@ var (
...
+// other header here is still meaningful in current browsers:
+//   - Content-Security-Policy and X-Frame-Options are the load-bearing
+//     clickjacking/injection defenses.
+//   - Referrer-Policy and X-Content-Type-Options are cheap, still-honored
+//     hardening with no compatibility downside.
+//
+// HTML responses (the embedded SPA) override Content-Security-Policy via
+// applySPAContentSecurityPolicy (see embed.go) to allow the styling the SPA
+// needs; JSON API responses keep the strict default-src 'self' policy set
+// here. See docs/security.md for the full per-response-type header table.
```
leo-aa88: **MAJOR.** `docs/security.md` does not exist on this branch. The issue explicitly requires that page to list the per-response-type headers and also requires README/CHANGELOG notes; this PR adds none of them. A source comment pointing to imaginary documentation is itself a documentation bug. Add the required page and release notes, then make this reference real.

framework/observability_security_test.go
```diff
@@ -144,6 +143,11 @@ func TestDefaultRouterAddsSecurityHeaders(t *testing.T) {
 	if got := rec.Header().Get("X-XSS-Protection"); got != "" {
 		t.Fatalf("X-XSS-Protection = %q, want empty deprecated header", got)
 	}
+	// X-Download-Options is IE8-only guidance; no supported browser honors
```
leo-aa88, line 159: **MAJOR.** This proves only that one deprecated header disappeared. It does not assert the issue's actual invariants: total default JSON headers remain at or below eight, API and HTML responses receive different policies, and the `security_headers` ablation costs zero allocations. The current test would remain green if the response crossed the map-growth threshold again tomorrow. Add tests per response kind plus the allocation/ablation guard required by #267.

leo-aa88 (commented):
# REQUEST CHANGES

This patch removes one dead header, but the linked issue is a response-kind security-policy change with four acceptance criteria. The implementation stops after the easiest deletion and then documents behavior it did not build.

## 1. The response-kind contract is not implemented

**BLOCKING.** #267 requires API/JSON responses to use the API policy (`default-src 'none'; frame-ancestors 'none'`) while HTML/admin/docs/SPA responses retain the browser policy, with the default response kept within the eight-header allocation budget and the security-header ablation measured at zero allocations. This function still installs one fixed policy on every response, and the patch contains no per-response-kind tests or benchmark evidence. Deleting `X-Download-Options` may recover the map threshold, but it is only step 1 of the issue's proposed fix. Split the policy at the runtime boundary, prove both response kinds, and pin the header-count/allocation invariant.

## 2. The patch publishes nonexistent documentation and omits required release notes

**MAJOR.** The new comment sends maintainers to `docs/security.md`, which does not exist. #267 also explicitly requires that security page plus README/CHANGELOG notes for the removed header and CSP change; none are present. The PR body does not use the repository template or enumerate the acceptance criteria, and its title describes an ablation benchmark that this diff does not contain. Public comments and PR metadata are part of the contract; they cannot describe a different change.

# VERDICT

The local deletion is plausible, but the PR is specification-incomplete and its documentation claims are false. This is not the issue it says it closes.

Do not merge until the API/HTML split, invariant tests and benchmark evidence, required documentation, and accurate PR contract are all present.

framework/http_middleware.go
```diff
@@ -94,32 +95,78 @@ func GetTraceIDFromContext(ctx context.Context) string {
...
-		header["Content-Security-Policy"] = cspHeaderValue
-		header["Referrer-Policy"] = referrerPolicyValue
+		header["X-Content-Type-Options"] = contentTypeOptionsValue
 		if includeHSTS {
 			header["Strict-Transport-Security"] = hstsHeaderValue
 		}
-		header["X-Content-Type-Options"] = contentTypeOptionsValue
-		header["X-Download-Options"] = downloadOptionsValue
-		header["X-Frame-Options"] = frameOptionsValue
+		if isDocsPath(c.Request.URL.Path) {
```
leo-aa88: ## 1. This classifies a path prefix, not a response kind

**MAJOR.** `isDocsPath` runs before dispatch and knows nothing about whether the docs handler exists. Production defaults set `DocsEnabled=false`, so Huma registers no `/docs` route; nevertheless `/docs` and every `/docs/...` 404 are assigned the HTML-only `Referrer-Policy`/`X-Frame-Options` set and pay the extra header-map cost. The docs and comment claim headers are scoped by response kind, but an API/not-found response is treated as browser HTML solely because of its URL.

Thread `cfg.API.DocsEnabled` into this decision (at minimum) and test disabled docs, or move promotion to the actual docs handler boundary as the SPA/admin handlers already do. A path that would have been HTML under a different configuration is not an HTML response in this one.

leo-aa88 (commented):
# REQUEST CHANGES

The runtime implementation now covers the intended API, SPA, admin, and enabled-docs paths, and the allocation/header tests are materially better. Current HEAD still does not actually classify by response kind, however, and the PR metadata describes an entirely different patch: the title says ablation benchmark, while the body says this only updates one deprecated-header assertion. That violates this repository's mandatory PR template and makes the claimed contract unusable for review.

# VERDICT

The remaining code issue is an implementation/abstraction mismatch at the docs boundary; the review-description issue is specification and process. Fix the disabled-docs path classification, add the adversarial case, and rewrite the title/body against the actual runtime/security change before merge. Because this is the author's own PR, GitHub does not permit a formal request-changes event; this comment records the equivalent verdict.

framework/http_middleware.go
```diff
@@ -94,32 +95,84 @@ func GetTraceIDFromContext(ctx context.Context) string {
...
-		header["Content-Security-Policy"] = cspHeaderValue
-		header["Referrer-Policy"] = referrerPolicyValue
+		header["X-Content-Type-Options"] = contentTypeOptionsValue
 		if includeHSTS {
 			header["Strict-Transport-Security"] = hstsHeaderValue
 		}
-		header["X-Content-Type-Options"] = contentTypeOptionsValue
-		header["X-Download-Options"] = downloadOptionsValue
-		header["X-Frame-Options"] = frameOptionsValue
+		if docsEnabled && isDocsPath(c.Request.URL.Path) {
```
leo-aa88, line 155: ## 1. `DocsEnabled` does not make every `/docs/*` response HTML

**BLOCKING.** Huma registers the interactive UI at the exact configured docs path; it does not register the whole `/docs/` subtree. With docs enabled, a request such as `/docs/not-a-route` therefore returns 404, but this condition is true because `isDocsPath` accepts every descendant. I reproduced that on this HEAD: the response was 404 and still carried `Referrer-Policy: strict-origin-when-cross-origin` plus the browser CSP.

The comment says policy follows what is actually served, but runtime behavior still follows a guessed prefix. Gate on the actual registered route (or at minimum the exact Huma docs path), and add an enabled-docs unknown-path test that requires API policy.

leo-aa88 (commented):
# REQUEST CHANGES

The configuration plumbing now correctly prevents disabled docs paths from receiving browser policy. The implementation still classifies nonexistent descendants as HTML whenever docs are enabled, so it has not yet reached the claimed response-kind contract.

CI is green. This is not a failing-test problem; the added regression test covers the disabled case but not an unknown path under enabled docs.

# VERDICT

The remaining problem is an implementation/abstraction mismatch: middleware predicts response kind from a broad URL prefix that the router does not actually serve. Do not merge until policy selection follows the registered docs route and the enabled-docs 404 case is tested.

GitHub does not permit a formal request-changes review on a pull request authored by the reviewing account, so this is posted as a comment review.

leo-aa88 (commented):
LGTM

Issue #291, opened 2026-09-08
[bug] Benchmark publishability guard never judges the host: a developer-laptop snapshot publishes banner-free, and results/README.md describes a different run


Brunotlps:
## Summary

Two related defects in the benchmark **publishability guard**, found while investigating #266.

The guard exists so a substandard snapshot can never quietly wear the authority of the canonical run: `make benchmark-report` stamps `⚠️ UNPUBLISHABLE DEVELOPMENT RUN` for a dirty tree and `Reduced development snapshot` for a narrower protocol ([report.go#L96-L124](https://github.com/gombit-dev/gombit/blob/main/benchmarks/internal/report/report.go#L96-L124)). It covers **how** a run was executed, but nothing about **where**:

1. **`reducedFrom` never judges the host.** A run on an ordinary developer laptop that happens to use the canonical protocol publishes with **no banner at all** — visually indistinguishable from a dedicated-host run. That is the state of the committed snapshot today.
2. **`benchmarks/results/README.md` describes a snapshot that is not the committed one.** Its prose calls the current snapshot a "reduced run" on "a contended WSL2 laptop"; `metadata.json` says the protocol was canonical and the kernel is not a WSL2 kernel. The page that explains the snapshot to a newcomer is describing a different run.

Both point at the same gap: the project has recorded **twice, in prose**, that the committed snapshot is not from dedicated hardware — and the generated artifact says nothing.

## Evidence (measured)

**1. The guard is protocol-only.** [`reducedFrom`](https://github.com/gombit-dev/gombit/blob/main/benchmarks/internal/report/report.go#L135) compares exactly four fields:

```go
meta.Concurrency      vs CanonicalProtocol.Concurrency
meta.Trials           vs CanonicalProtocol.Trials
meta.DurationSeconds  vs CanonicalProtocol.DurationSeconds
meta.WarmupSeconds    vs CanonicalProtocol.WarmupSeconds
```

No hardware term. [`writeMethodology`](https://github.com/gombit-dev/gombit/blob/main/benchmarks/internal/report/report.go#L310) *prints* the host (CPU model, logical CPUs, RAM, kernel), but nothing ever *evaluates* it.

[...]

Issue #270, opened 2026-09-03
[PERF-12] Decide: make the per-handler request timeout opt-in (4–5 allocs/op on every request by default)


leo-aa88:
## Summary

`requestTimeoutMiddleware` ([framework/http_middleware.go#L29](https://github.com/gombit-dev/gombit/blob/main/framework/http_middleware.go#L29)) runs on every request with the default `HTTP.RequestTimeout = 60s` ([config/config.go#L303](https://github.com/gombit-dev/gombit/blob/main/config/config.go#L303)). #242 removed the redundant re-wrap, but on the common path (no inbound deadline) `context.WithTimeout` is intrinsic: a `timerCtx`, a runtime timer, the cancel closure, and a `Request.WithContext`. This issue is the decision on whether the per-handler deadline should be on by default.

## Evidence (measured)

- Cumulative ablation: +request_timeout = +5 allocs/op on every scenario (after the merge in the request_context issue, +4), ≈0.45 µs.
- `alloc_objects`: `context.WithDeadlineCause` 3/op, `time.newTimer` 1/op, `Request.WithContext` 1/op.
- The httptest harness understates production cost: under a real `http.Server` the parent context is the connection's cancel context, so `propagateCancel` also registers the child (one more map/entry alloc per request).
- Prototype with the timeout disabled: plaintext 22→18 allocs, ≈2.1 µs → 1.7 µs.

## Proposed fix

Decide between:

- **A. Opt-in (recommended for the milestone target).** Default `HTTP.RequestTimeout = 0`; document that the `http.Server` `ReadHeaderTimeout`/`ReadTimeout`/`WriteTimeout`/`IdleTimeout` remain the safety net, and that setting `GOMBIT_HTTP_REQUEST_TIMEOUT` enables context propagation of a per-handler deadline into DB/cache calls. Scaffolded apps (`gombit new`) may set it explicitly in their generated config so new projects keep today's behavior.
- **B. Keep on by default** and close as won't-fix with the rationale recorded (cancellation of handler work is worth 4 allocs per request).

Either way, make the disabled case a true no-op: `runtimeMiddlewareStack` should omit the layer entirely when the timeout is 0 rather than install a pass-through.

Trade-off to state in the docs: with A, a slow handler under no per-handler deadline keeps running after `WriteTimeout` closes the connection; long-running DB queries are not cancelled unless the app enables the timeout or sets its own deadlines.

## Acceptance criteria

[...]

Issue #268, opened 2026-09-03
[PERF-10] request_context is 11 allocs/op: drop gin.Context.Set, single ID string, direct header slices, merge timeout wrap


leo-aa88:
## Summary

`requestContextMiddleware` ([framework/request_id.go#L38](https://github.com/gombit-dev/gombit/blob/main/framework/request_id.go#L38)) is the single largest allocator in the stack at **11 allocs/op**, and `requestTimeoutMiddleware` ([framework/http_middleware.go#L29](https://github.com/gombit-dev/gombit/blob/main/framework/http_middleware.go#L29)) then performs a second `Request.WithContext` on the same request. Both IDs are already propagated through the standard context, so the two `c.Set` calls (Gin `Keys` map + two boxed strings) are redundant with `GetRequestIDFromContext` / `GetTraceIDFromContext`.

## Evidence (measured)

Per-request allocations today, from `alloc_objects`: `newRequestID` string copy (1), `newTraceID` string copy (1), `gin.Context.Set` map (1) + boxed strings (2), `Header.Set` ×2 (2), `context.WithValue` (1), boxed `requestMeta` (1), `Request.WithContext` (1) = 10–11; then the timeout layer adds another `Request.WithContext` (1) on top of its timer/context allocs.

Escape analysis (`go build -gcflags=-m`): `make([]byte, 36)` and `make([]byte, 32)` already do not escape; only `string(buf)` allocates. Switching to array literals gains nothing.

Prototype (`leanRequestContext` in the ablation variants file) combining the changes below: POST 60→52, plaintext 35→27 allocs/op, plaintext ≈3.4 µs → 2.8 µs, all scenario assertions and correlation-header checks passing.

## Proposed fix

- Remove both `c.Set` calls. Reimplement `GetRequestID(*gin.Context)` and `GetTraceID(*gin.Context)` as `GetRequestIDFromContext(c.Request.Context())` / `GetTraceIDFromContext(...)`. Public API unchanged. (−3 allocs)
- Encode both IDs into one stack buffer (`[68]byte`) and take one `string()`; sub-slice it into requestID and traceID. (−1)
[...]

PR #292, opened 2026-09-08
fix(bench): per-unit benchmark provenance + refresh the framework-tax snapshot (#266)

### LINKED ISSUE #266: [PERF-8] results/latest snapshot predates PERF-1..4; README framework-tax table overstates Gombit by ~33 allocs
## Summary

`benchmarks/results/latest` (and therefore the README framework-tax table) was captured at 88442ae on 2026-08-27, before PERF-1 through PERF-4 (#239, #240, #241, #242) merged. The README still shows Gombit at 93 allocs / ≈23 µs on valid POST and 45 allocs on plaintext; current `main` measures 60 and 35 allocs. A third of the reported tax is already gone.

## Evidence (measured)

- `benchmarks/results/latest/metadata.json`: `git_commit: 88442ae…`, `timestamp: 2026-08-27T21:57:34Z`.
- `git merge-base --is-ancestor 88442ae 872bb98` → true (snapshot predates the first perf PR).
- `go test ./benchmarks/micro/gombit -bench BenchmarkFrameworkTax -benchmem` on current main: valid-post 60 allocs/op, plaintext 35 allocs/op.

## Proposed fix

Re-run the BENCH-1 snapshot on the reference machine with the documented procedure and commit the refreshed `results/latest` plus README table. Do it once now, and again after each child of #264 lands so the README tracks the milestone. Consider adding a CI check that warns when `metadata.json.git_commit` is older than the last commit touching `framework/*middleware*.go`, `framework/request_id.go`, or `framework/xss.go`.

## Acceptance criteria

- [ ] `results/latest/metadata.json` `git_commit` is at or after the PERF-4 merge.
- [ ] README framework-tax table reflects the refreshed numbers, including allocs/op.
- [ ] (Optional) staleness warning in CI.

## Scope notes

Docs/bench chore. Part of #264.

Brunotlps:
## Summary

`benchmarks/results/latest/` holds three independently produced groups at very different costs: the framework tax (`make benchmark-micro`, minutes of pure Go), CRUD (`make benchmark-crud-all`, hours of Docker) and footprint (`make benchmark-footprint`, minutes of Docker). `metadata.json` carried one `git_commit` and host for all three, so refreshing the cheap table meant either claiming the CRUD sweep ran at the new commit or publishing fresh numbers under the old one. Nobody did either, which is how the README kept advertising pre-PERF-1..4 numbers.

This PR decouples them, then does the refresh:

- **Provenance per unit, stamped by the writer that merges the row.** `metadata.json` gains `groups.<group>.<unit>` (commit, dirty state, host, Go toolchain). Each unit equals its data file's merge key: microbench stack namespace (`scripts/microbench`), CRUD framework (`scripts/run-crud`), footprint `framework:variant` (`scripts/footprint`).
- **Per-table captions from the units a table publishes.** Comparable units (same commit, dirtiness, host and toolchain, not the clock) collapse to one caption with the measurement span. Anything else refuses a table-wide claim and lists each unit.
- **All-or-nothing fallback.** Only a snapshot that records no unit is read through its top-level block. Otherwise an unrecorded unit is captioned as unrecorded, never with whatever last rewrote the top level.
- **Fail closed.** One reader, `metadata.ReadFile`, rejects an unreadable `metadata.json` before any writer touches its data file.
- **The refresh.** `make benchmark-micro` re-run on current `main`; only the four microbench units moved. `schema_version` stays `1` (additive).

Validated typed POST, allocs/op (README before → after):

| stack | before | after | B/op after | vs net/http |
| --- | ---: | ---: | ---: | ---: |
| net/http | 18 | 21 | 2097 | 1.0× |
| Gin | 26 | 29 | 2321 | 1.8× |
| Huma + Gin | 31 | 37 | 2316 | 2.6× |
| **Gombit** | **93** | **42** | **3401** | **3.1×** (was 9.5×) |

The baseline rungs moved against the old snapshot because the Go toolchain changed: the old snapshot records `go1.27.0` and this one `go1.26.1`, while Gin, Huma and the serial benchmark code are unchanged. Across the two runs on this branch under the same toolchain (2ee9f8b, then e9a98f7) they reproduced their allocs/op exactly, so Gombit's 59 → 42 comes from PERF-9 (#267), PERF-10 (#268) and PERF-12 (#270), whose fixes merged in between.

Measured on e9a98f7: clean tree, idle host, `MICRO_COUNT=10`. CoV is 0.2–1.7% across all 20 stack × scenario series and 0.3–0.5% on the published row; no row is flagged. It is the only run taken on that commit.

Closes #266

## Review round 2

**Blocking finding: footprint provenance was coarser than footprint writes.** Correct on every point. Provenance was aligned with the make target (one target, one stamp) instead of the row merge (one row, one stamp).

- **Fix (2ee9f8b):** stamping moved from the shell into the three row-writers, through one read-modify-write (`metadata.StampUnitFile`). `footprint-all.sh` and the `Makefile` are no longer touched by this PR, and the CRUD "known limitation" prose is gone. The rule is enforced in code for all three producers.
- **Why not the atomic alternative:** `APPS=` subsets are a supported path that CI's `benchmark-smoke` uses, and "the full set" is configurable, so atomic replacement has no fixed definition. Refusing to publish detects the bad state; per-unit provenance makes it unrepresentable. An interrupted loop is the same state as a subset: each row keeps the provenance of the run that wrote it.
- **"Add a test that starts with multiple rows and refreshes one":** every producer now has one that starts from the committed snapshot and runs the real writer. `TestSubsetRefreshThroughTheWriterNeverRelabelsUntouchedRows` covers footprint, `TestSubsetRunNeverReattributesRowsItDidNotMeasure` covers `run-crud`, and `TestPartialWriteThroughTheWriterNeverRelabelsOtherStacks` covers microbench (one stack, and the ablation ladder). The footprint and microbench tests are mutation-checked against a group-wide stamp.

Auditing the fix before pushing found more defects. Each one passed a green suite:

1. **The fallback reopened the blocker** (b6c9eb6). The committed snapshot's 12 CRUD and footprint units fell back to the top-level block, which `run-crud` rewrites on every app (including a one-app `APPS=` subset) and `make benchmark-metadata` rewrites with a collection that measured nothing. Either run captioned untouched rows, and the untouched footprint table, with the new commit and host. The fallback is now all-or-nothing, and the 12 units carry explicit entries, byte-identical to their existing captions.
2. **Comparability judged the clock** (390f588). The four `benchmark-micro` stacks finish minutes apart, so "not measured together" fired on every ordinary run.
3. **The footprint unit ignored the variant** (ad595a0). An embedded-binary run would have relabelled the published container row.
4. **`make benchmark-metadata` erased all provenance** (62db5c9).
5. **Writers failed after writing rows** (4a130ed). `run-crud` silently replaced a corrupt `metadata.json`; `microbench` and `footprint` merged rows before discovering it.
6. **A dirty PERF-7 (#265) ablation run flagged the published table UNPUBLISHABLE** (2ab0c49). The banner now judges only published units.
7. **The snapshot went stale again while the PR was open** (eef6ff8). `main` was merged in and the table re-measured, instead of publishing 59 allocs/op when `main` measures 42.
8. **Two stray build binaries were committed in 2ee9f8b** and removed in 747afac; `.gitignore` now anchors the command names to the root.

`main` was merged in (eee9396, e9a98f7) rather than rebased, so no pushed commit is rewritten and the SHA recorded in `metadata.json` exists.

## Acceptance criteria

- [ ] **`results/latest/metadata.json` `git_commit` is at or after the PERF-4 merge.** @leo-aa88, this needs your confirmation: it is the author's reading of the criterion, not a decision you made. The commit that answers "when was the framework-tax table measured" is now `groups.microbench.<stack>.git_commit` = e9a98f7. The top-level `git_commit` stays at 88442ae because it is no longer the caption for any table, and advancing it would claim that the unchanged CRUD and footprint rows ran at a commit and host they never ran on. If you want the literal top-level field advanced instead, that is a small follow-up.
- [x] **README framework-tax table reflects the refreshed numbers, including allocs/op.** Regenerated with `make benchmark-report`.
- [ ] **(Optional) staleness warning in CI.** Deferred to #293. It must stay a warning, and BENCH-1 plan §3 rules out perf gates in PR CI; no workflow file is touched.

## Scope notes

- **Re-running after the remaining #264 children.** #266 also asks for a re-run after each #264 child lands, and #271 is still open. That recurring re-run is #264's last checklist item, and #293 would flag a stale snapshot. Closing #266 does not mean the table is final.
- **Host.** #266 asks for the "reference machine", which no doc defines; the previous snapshot also came from a developer laptop. The framework-tax table is now from a different developer host (i5-11400H) than the CRUD and footprint tables (i5-8250U), stated in each caption. It publishes with no banner, which is the gap tracked in #291. This PR also makes `benchmarks/results/README.md`'s "single developer host" wording inaccurate; that file's description of the snapshot is #291's second criterion, so it is left for that decision.
- **Residual trade-off.** In an `OUT_DIR` holding only a `metadata.json`, `make benchmark-metadata` keeps the existing top-level block, because in a pre-provenance snapshot it is every row's only provenance. A first collection taken on a dirty tree therefore keeps the banner until a row-writer runs. It is not reachable from the committed snapshot, which records every unit.
- **Found, not fixed:** `run-crud`'s `mergeRows` drops a framework's rows for every workload (#361). It is latent while `crud-list` is the only workload.
- **Pre-existing:** `benchmarks/scripts/footprint-all_test.sh` fails under `LANG=pt_BR.UTF-8` (`awk` decimal separator), identically on `main`; it passes with `LC_ALL=C`, which is the CI condition.
- **CHANGELOG N/A:** the benchmark harness has no CHANGELOG entries (BENCH-1, #289).

## Validation

- [x] `gofmt -l benchmarks/` — clean
- [x] `go build ./...` — OK
- [x] `go test ./...` — pass
- [x] `make benchmark-report-check` — README matches the data
- [x] `LC_ALL=C bash benchmarks/scripts/{run-suite,benchmark-target,run-crud-all,footprint-all}_test.sh` — OK
- [x] Contract drift: `go run ./cmd/gombit client check --write --spec examples/client/openapi.json --out examples/client/frontend/src/api/generated` leaves `git status --porcelain` empty; `go test ./goldentest/...` passes
- [x] End to end against a copy of the committed snapshot with the real writers, reading the rendered README: a one-app `scripts/footprint` run yields "Rows were not measured together" with five rows kept at 88442ae; `make benchmark-metadata` leaves all three captions unchanged
- [x] Project `code-review` skill run against this diff, three passes. Round 1 returned REQUEST CHANGES (fixed in 62db5c9, ec5217c). A pass on the round-2 branch returned REQUEST CHANGES with seven findings, fixed as items 1 and 5–7 above plus docs and comments. A final pass before push returned COMMENT, with no code defect: it found the stale microbench cost and unit in docs (fixed in 56c6f95) and the disclosures now in Scope notes.

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix — 35 new Go tests across `internal/metadata`, `internal/report` and all four writers; the scenario tests start from the committed snapshot, and the fallback, fail-closed, banner and writer tests fail against the previous code. DB matrix N/A: no persistence code touched.
- [x] Stable features ship docs and appear in an example app — `methodology.md` documents per-unit provenance, the all-or-nothing fallback and how to read `metadata.json`. Example app N/A: benchmark harness, not a framework feature.
- [x] Extraction preserves contracts — refactor and move, don't rewrite code that already passes its tests — `run-crud`'s metadata merge moved into `metadata.Merge`; the Postgres-sentinel test still asserts the same semantics and now also checks the returned error.
- [ ] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files (if this PR touches generators) — N/A: no generator touched; `go test ./goldentest/...` passes.
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR — N/A: no API change; regeneration verified byte-identical.
- [ ] No secrets in generated frontend source; `VITE_*` is treated as public — N/A: no frontend file touched.
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep (jobs, events, scheduler, mail, storage, gRPC, multi-tenancy, i18n) — `post-v0.1`, part of #264; only `benchmarks/`, `README.md` and `.gitignore` change, with no workflow, no ADR and no locked decision touched.


benchmarks/scripts/footprint-all.sh
```diff
@@ -263,6 +263,13 @@ main() {
   for app in $APPS; do
     measure_container "$app"
   done
+  # Record which commit and host produced these rows. Footprint is a separate
+  # measurement group from the CRUD sweep and the microbenchmark, so it stamps
+  # its own provenance and leaves the rest of metadata.json untouched — a
+  # snapshot-wide commit would vouch for numbers that never ran at it
+  # (issue #266).
+  mkdir -p "$(dirname "$OUT")"
+  go run ./benchmarks/scripts/collect-host-info -group footprint -out "$(dirname "$OUT")/metadata.json"
```
leo-aa88, line 272: **BLOCKING.** This stamps the whole footprint table after a row-level merge. `record_footprint` preserves rows for frameworks not remeasured, while `APPS` explicitly supports subset runs. Concrete failure: six rows exist from commit A; `APPS=gombit make benchmark-footprint` at B replaces only Gombit, then this line records `groups.footprint = B`; the report captions all six rows as B. An interrupted loop similarly leaves mixed rows with stale provenance. Align provenance with the row merge (per-row provenance), or write the entire group atomically and refuse to publish partial/subset replacements under a group-wide caption. Add a test that starts with multiple rows and refreshes one.

leo-aa88 (changes requested):
# REQUEST CHANGES

Per-group provenance fixes the microbenchmark refresh, but the abstraction is not sound for every producer it claims to cover. The footprint target can update one row and then relabel the entire table.

## 1. Footprint provenance is coarser than footprint writes

**BLOCKING.** `footprint` merges by `(framework, variant)`, and both the Makefile and `footprint-all.sh` document `APPS=...` as a supported subset run. After such a run, this new line stamps one table-wide `groups.footprint` value. Start with six rows measured at commit A, run `APPS=gombit make benchmark-footprint` at commit B, and five untouched A rows are now captioned as measured at B. An interrupted full run has the same failure mode. That is the exact confident misattribution this PR exists to eliminate.

The architecture must align provenance ownership with write ownership. Either store provenance per footprint row, or make a footprint group replacement atomic and prevent subset/partial results from being published under a table-wide caption. The same rule should be applied consistently to CRUD rather than leaving correctness dependent on a prose warning.

# VERDICT

This is an abstraction mismatch between row-level mutation and table-level provenance. The metadata/report unit tests are green, but they never execute the supported subset-refresh path, so they prove only that the new representation agrees with itself.

Do not merge while a documented benchmark command can produce a falsely attributed canonical table.

benchmarks/scripts/footprint-all.sh
Brunotlps: Agreed on every point. Fixed by aligning provenance with the row merge instead of the make target. This line is gone and `footprint-all.sh` is no longer touched by the PR. `scripts/footprint` now stamps exactly the row it merged, under its full merge key (`framework:variant`), and `run-crud` and `microbench` do the same for their units.

An interrupted loop is now the same state as a subset: each row keeps the provenance of the run that wrote it, and the report refuses a table-wide caption when the published rows disagree.

The test you asked for is `TestSubsetRefreshThroughTheWriterNeverRelabelsUntouchedRows`. It copies the committed six-row snapshot, re-measures `gombit` through the real writer, and renders the README. It is mutation-checked against a group-wide stamp, and CRUD and microbench have the same kind of test. Auditing the fix also showed that the top-level fallback reopened this failure through `run-crud` and `make benchmark-metadata`; that is fixed too. Details are in the PR body under "Review round 2".

leo-aa88 (approved):
LGTM

PR #295, opened 2026-09-12
fix: CSRF cookie reuse + request-time read + 403 recovery (#250)

### LINKED ISSUE #250: [bug] Cookie-mode CSRF breaks with a second tab or after cookie expiry: /auth/csrf always rotates and SPAs cache the token in memory with no 403 recovery
### What happened

In cookie mode, opening the admin (or the generated cookie-mode SPA) in a
**second tab** permanently breaks every write (POST/PATCH/DELETE) in the first
tab: they all fail `403 {"error":{"code":"authorization_error","message":"csrf
token missing or invalid"}}` until that tab is hard-reloaded. The same failure
hits a single long-lived tab once the CSRF cookie's 24h `Max-Age`
(`csrfCookieMaxAge`, `auth/cookie.go`) elapses.

Three pieces combine into this:

1. `GET /auth/csrf` **always mints a fresh token pair** and re-sets the
   `gombit_csrf` cookie (`issueCSRFToken`, `auth/csrf.go:157`) — it never
   reuses a still-valid cookie. Every new tab's bootstrap therefore rotates
   the cookie shared by all tabs.
2. Both SPAs keep the CSRF token **in module memory only** and treat a
   present token as final: `bootstrapCSRF()` is a no-op when
   `getCSRFToken()` is non-empty (`internal/adminui/src/api/client.ts:194`,
   `scaffold/templates/frontend/src/api/client.ts.tmpl:118`). A tab whose
   in-memory token no longer matches the (rotated or re-minted) cookie keeps
   double-submitting the stale value forever. The cookie is deliberately
   *not* HttpOnly precisely so the SPA could read it (`csrfCookie`,
   `auth/cookie.go:68`), but neither SPA ever does.
3. Neither client has a **403 recovery path**: the fetch wrappers only handle
   401 (silent refresh). A CSRF 403 is surfaced to the user as a failed save
   and every retry fails the same way.

The 24h single-tab variant: the session outlives the CSRF cookie
(refresh tokens last 7 days), the expired cookie is re-minted with a *new*
value by `ensureCSRFCookie` on the next safe GET (`auth/csrf.go:89`), and the
tab's stale in-memory token then mismatches on every write.

Expected: multiple admin tabs work concurrently, and a tab recovers from a
rotated/expired CSRF cookie without a manual page reload.

### Steps to reproduce

```shell
1. gombit new demo --auth cookie && cd demo && gombit dev
2. gombit createsuperuser --email a@b.c --password secret123 --no-input
3. Open http://127.0.0.1:8080/admin/ in tab A, log in, verify editing a resource works.
4. Open http://127.0.0.1:8080/admin/ in tab B (same browser). Its bootstrap
   calls GET /auth/csrf, which sets a NEW gombit_csrf cookie.
5. Back in tab A, save any change.
   # => 403 {"error":{"code":"authorization_error","message":"csrf token missing or invalid"}}
   # Every subsequent write in tab A fails until the page is reloaded.
```

Same effect with one tab: leave it open >24h (or delete the gombit_csrf
cookie in devtools, then make any GET so the middleware re-mints it), then try
a write.

### gombit version

### Environment

```text
go 1.x, linux; any browser
sqlite default app, --auth cookie
```

### Database

not database related

### Auth mode

cookie (session + CSRF)

### Area

auth, admin, frontend (Vite / React)

### Logs or output

```text
POST /api/v1/admin/resources/widgets/1 -> 403
{"error":{"code":"authorization_error","message":"csrf token missing or invalid","request_id":"..."}}
```

### Suggested fix

Any one of these breaks the failure mode; together they make it robust:

- Server: `GET /auth/csrf` should return the existing cookie's value when the
  request already carries a valid signed `gombit_csrf` cookie, instead of
  always rotating (double-submit does not require rotation per bootstrap).
- Client: read the token from `document.cookie` at request time (the cookie
  is intentionally JS-readable) instead of caching it in module memory.
- Client: on a 403 from an unsafe request, clear the cached token,
  re-bootstrap CSRF, and retry once — mirroring the existing 401 silent
  refresh.

### Before submitting

- [x] I searched existing issues for a duplicate (#107, #115, #119, #226 are
  adjacent but none cover cookie rotation across tabs / expiry recovery)

leo-aa88:
## Summary

Cookie-mode CSRF broke every write in a first tab once a second tab opened (and hit a single long-lived tab after the 24h cookie `Max-Age`): `GET /auth/csrf` always minted a fresh token and re-set the shared `gombit_csrf` cookie, while both SPAs cached the token in module memory and had no 403 recovery. Fixed with the three coordinated changes the issue suggests:

- **Server** (`auth/csrf.go`): `GET /auth/csrf` now **reuses** a still-valid signed `gombit_csrf` cookie (read via a cookie input) and only mints when the cookie is missing/invalid; re-setting the same value refreshes `Max-Age`. Double-submit needs a *shared* signed value, not a per-call-unique one.
- **Clients** (`internal/adminui` + `scaffold/templates/.../client.ts.tmpl`): read the token from the JS-readable `gombit_csrf` cookie **at request time** (in-memory is only a fallback), so a rotated/expired cookie is always matched; `bootstrapCSRF` keys off cookie presence (not memory) with a `force` flag; and a **403 on an unsafe request** clears the mirror, force-re-bootstraps CSRF, and retries once — mirroring the existing 401 silent-refresh path.

Closes #250

## Acceptance criteria

- [x] Multiple admin tabs work concurrently (a second tab's bootstrap no longer rotates the shared cookie; the server reuses it)
- [x] A tab recovers from a rotated/expired CSRF cookie without a manual reload (request-time cookie read + 403 re-bootstrap-and-retry)

## Scope notes

Both the framework-owned admin SPA (`internal/adminui`) and the scaffold cookie-mode template carry the client fix; the `new-cookie` golden is regenerated. No API contract/envelope change.

## Validation

- [x] `go test ./auth/...` (server reuse + invalid-cookie replacement tests) — pass
- [x] `npm test` + `npm run typecheck` in `internal/adminui` (cookie-read, 403-retry, no-retry-on-safe-403) — pass
- [x] `go test ./goldentest` (regenerated `new-cookie` client.ts) — pass
- [x] `golangci-lint run ./auth/...` — 0 issues
- [x] Project `code-review` skill run against this diff

- [x] New behavior has tests (Go auth tests + adminui vitest); not a DB-touching change, so the DB matrix is N/A
- [x] Stable features ship docs — `docs/auth-cookie.md` updated to the reuse + request-time-read + 403-recovery behavior; the cookie-mode behavior is exercised by the scaffold template + golden
- [x] Extraction preserves contracts — refactor/extend, existing auth tests still pass
- [ ] Generators: N/A — no generator logic changed (the scaffold client template is edited, and its golden regenerated)
- [x] API changes regenerate OpenAPI + TS client — N/A: the `gombit_csrf` cookie input is ambient and `/auth/csrf` is not part of the committed sample OpenAPI; no generated contract changed
- [x] No secrets in generated frontend source; `VITE_*` is public
- [x] Scope stays inside M5 (auth/frontend) + admin — no M6 battery
- [x] Links its issue (#250) and states the AC it satisfies

### DIFF
--- a/internal/adminui/src/api/client.ts
+++ b/internal/adminui/src/api/client.ts
@@ -139,19 +139,32 @@ export function createAdminClient() {
       await bootstrapCSRF();
     }
 
+    // Bounded recovery: a CSRF 403 (rotated/expired cookie) and a session 401
+    // (expired access cookie) can each occur — and can occur consecutively on
+    // the same write, e.g. a 403 whose recovered retry then 401s. Recover each
+    // kind at most once, in either order, re-evaluating after every retry and
+    // rebuilding the request (init() re-reads the live cookie). The auth-path
+    // exclusion applies only to the 401 refresh loop, not to CSRF recovery (#250).
     let response = await fetch(url, init());
-    if (response.status === 401 && !isAuthPath(path)) {
-      const ok = await refreshSession();
-      if (ok) {
+    let csrfRetried = false;
+    let authRetried = false;
+    for (let i = 0; i < 2; i++) {
+      if (response.status === 403 && isUnsafeMethod(method) && !csrfRetried) {
+        csrfRetried = true;
+        setCSRFToken(undefined);
+        await bootstrapCSRF(true);
         response = await fetch(url, init());
+        continue;
       }
-    } else if (response.status === 403 && isUnsafeMethod(method)) {
-      // A CSRF 403 means our cookie was rotated/expired out from under us.
-      // Drop the stale in-memory mirror, force a fresh bootstrap, and retry
-      // once — mirroring the 401 silent-refresh path (#250).
-      setCSRFToken(undefined);
-      await bootstrapCSRF(true);
-      response = await fetch(url, init());
+      if (response.status === 401 && !isAuthPath(path) && !authRetried) {
+        authRetried = true;
+        if (!(await refreshSession())) {
+          break;
+        }
+        response = await fetch(url, init());
+        continue;
+      }
+      break;
     }
 
     const parsed: unknown = await readJSON(response);
--- a/internal/adminui/src/api/client.ts
+++ b/internal/adminui/src/api/client.ts
@@ -57,8 +57,10 @@ function apiPath(path: string): string {
 
 /**
  * Thin same-origin fetch wrapper around D10 `{data, meta, error}`.
- * Cookie session + CSRF double-submit (X-CSRF-Token) on unsafe methods.
- * Concurrent 401s share one refreshInFlight promise. CSRF stays in memory.
+ * Cookie session + CSRF double-submit (X-CSRF-Token) on unsafe methods. The
+ * token is read from the JS-readable gombit_csrf cookie at request time.
+ * Concurrent 401s share one refreshInFlight promise; a 401 silent-refreshes and
+ * a 403 re-bootstraps CSRF, each retrying the request once.
  */
 export function createAdminClient() {
   const baseUrl = apiOrigin();
@@ -99,13 +101,16 @@ export function createAdminClient() {
     options?: { body?: unknown; query?: Record<string, string | number | undefined> },
   ): Promise<Envelope<T, M>> {
     const url = buildURL(baseUrl, path, options?.query);
-    const init = (csrf?: string): RequestInit => {
+    const init = (): RequestInit => {
       const headers = new Headers();
       if (options?.body !== undefined) {
         headers.set("Content-Type", "application/json");
       }
       if (isUnsafeMethod(method)) {
-        const token = csrf ?? getCSRFToken();
+        // Read the token at request time from the JS-readable gombit_csrf
+        // cookie (the authoritative double-submit value), so a cookie rotated
+        // or re-minted by another tab / after expiry is always matched (#250).
+        const token = currentCSRFToken();
         if (token) {
           headers.set("X-CSRF-Token", token);
         }
@@ -128,6 +133,13 @@ export function createAdminClient() {
       if (ok) {
         response = await fetch(url, init());
       }
+    } else if (response.status === 403 && isUnsafeMethod(method)) {
+      // A CSRF 403 means our cookie was rotated/expired out from under us.
+      // Drop the stale in-memory mirror, force a fresh bootstrap, and retry
+      // once — mirroring the 401 silent-refresh path (#250).
+      setCSRFToken(undefined);
+      await bootstrapCSRF(true);
+      response = await fetch(url, init());
     }
 
     const parsed: unknown = await readJSON(response);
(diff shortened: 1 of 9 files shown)


scaffold/templates/frontend/src/api/client.ts.tmpl
```diff
@@ -86,6 +91,23 @@ export function createAppClient(): ApiClient {
       return request;
     },
     async onResponse({ request, response }) {
+      if (
+        response.status === 403 &&
+        isUnsafeMethod(request.method) &&
+        !isAuthURL(request.url)
```
leo-aa88: ## 1. Auth requests are excluded from the recovery contract

**BLOCKING.**

This condition retries a 403 only when the request is *not* an auth URL. The claimed contract says an unsafe request with a stale/rotated CSRF pair is bootstrapped and retried once, and the middleware explicitly protects `/auth/login`, `/auth/register`, `/auth/logout`, and `/auth/refresh` too.

Concrete failure: leave a non-empty but invalid `gombit_csrf` cookie, then submit login. `bootstrapCSRF()` sees a cookie and does nothing; login gets the CSRF 403; this branch rejects the retry solely because the URL is `/auth/login`. `refreshSession()` has the same hole because its direct refresh fetch treats a CSRF 403 as refresh failure and clears the session. The browser is stranded until the cookie is cleared by some out-of-band action. Why is the recovery code refusing to recover the requests that need it?

Keep the auth-path exclusion for recursive **401 refresh** prevention, not for one-shot CSRF recovery. Retry CSRF failures on auth requests once as well, and make the direct refresh path force-bootstrap/retry once on 403. Add tests for login/register/logout and refresh with an invalid existing cookie.

scaffold/templates/frontend/src/api/client.ts.tmpl
```diff
@@ -142,8 +164,25 @@ export function bootstrapCSRF(): Promise<void> {
...
+/**
+ * Reads the JS-readable gombit_csrf cookie — the authoritative double-submit
+ * value. The cookie is deliberately not HttpOnly so the SPA can echo it.
+ */
+function readCSRFCookie(): string {
+  if (typeof document === "undefined") {
+    return "";
+  }
+  const match = document.cookie.match(/(?:^|;\s*)gombit_csrf=([^;]*)/);
+  return match ? decodeURIComponent(match[1]) : "";
```
leo-aa88: ## 2. A malformed cookie throws before the server can replace it

**MAJOR.**

`decodeURIComponent` throws `URIError` for malformed percent escapes. With `gombit_csrf=%`, both `bootstrapCSRF()` and `currentCSRFToken()` reject before issuing a fetch. The server-side handler deliberately replaces unsigned/forged cookie values, but this parser prevents the request from ever reaching that handler. The admin client duplicates the same code.

The token alphabet generated by the server is hex plus `.`, so decoding is unnecessary; otherwise catch decoding failure and treat the cookie as invalid/missing. Cover malformed percent escapes in both the admin client and generated scaffold tests.

leo-aa88 (commented):
# REQUEST CHANGES

The server-side token reuse is coherent, but the scaffolded client still does not implement the claimed one-shot recovery for every unsafe request. CI is green. This is not a failing-test problem; the current tests do not exercise malformed or auth-path CSRF recovery.

# VERDICT

The remaining defects are in the client implementation and its adversarial test coverage. The recovery path must work for the authentication endpoints that are themselves CSRF-protected, and invalid cookie input must reach the server-side replacement path instead of crashing in JavaScript. Do not merge until both paths are covered.

leo-aa88 (commented):
# REQUEST CHANGES

The two previous findings are fixed, and the added tests prove each recovery branch independently. The request state machine still fails when those supported branches occur consecutively on the same logical write.

# VERDICT

This is an implementation and cross-feature test defect. CSRF recovery and session recovery are composed as a one-shot `if/else`, so satisfying one contract disables the other for the retry response. Do not merge until the logical request can consume at most one recovery of each kind without losing the original body.

internal/adminui/src/api/client.ts
```diff
@@ -128,6 +145,13 @@ export function createAdminClient() {
       if (ok) {
         response = await fetch(url, init());
       }
+    } else if (response.status === 403 && isUnsafeMethod(method)) {
```
leo-aa88: ## 1. Recovering CSRF can expose an unhandled session expiry

**BLOCKING.**

This dispatch examines only the first response. After the 403 branch retries, the code jumps straight to `readJSON`; it never applies the 401 recovery contract to that retry response. The scaffold middleware has the same one-pass architecture because its manual `fetch` result is returned directly.

Concrete failure: an unsafe request carries an invalid non-empty CSRF cookie and an expired session. The first attempt dies in CSRF middleware with 403. The client force-bootstraps a valid cookie and retries. That retry now reaches auth and returns 401. Instead of silent-refreshing and replaying the buffered request, this code throws the 401. The separate 403 and 401 tests both pass while the supported composition fails.

Use a bounded recovery loop/state machine with independent `csrfRetried` and `authRetried` flags, rebuilding the request body each time. Re-evaluate the response after each retry, allow each recovery at most once, and add this 403→401 case for both the admin client and generated openapi-fetch client.

leo-aa88 (commented):
LGTM

PR #296, opened 2026-09-12
fix: admin hard-deletes so the database enforces foreign keys (#220)

### LINKED ISSUE #220: [bug] Soft-delete makes FK/restrict constraints never fire, leaving live rows referencing API-404 parents
### What happened

`gorm.Model` gives every model a `DeletedAt`, so admin/generated deletes are
soft (they set `deleted_at`; the row physically remains). But Atlas still emits
real `ON DELETE NO ACTION` foreign keys for associations. Because no physical
DELETE ever happens, the restrict constraint **never fires**, and a child row is
left pointing at a parent the API now reports as 404.

So a schema that looks referentially safe is not: deleting a referenced parent
"succeeds", and you silently have a live row referencing an API-invisible parent.

Expected: soft-delete and FK/restrict semantics should be coherent, either the
delete is soft-delete-aware and blocks/handles referencing rows, or the
interaction is documented and there is a hard-delete/restrict path that actually
enforces the constraint.

### Steps to reproduce

```shell
gombit new demo --database sqlite --auth cookie --skip-tidy
cd demo
gombit make resource Engine name:string:required
# add a self-referential FK to internal/engine/engine.go:
#   ParentEngineID *uint
#   ParentEngine   *Engine `gorm:"foreignKey:ParentEngineID"`
# register Engine on the admin, migrate, log in, then via the admin data plane:
#   create parent P, create child C with parent_engine_id = P.id
#   DELETE /api/v1/admin/resources/engines/{P.id}     -> 200 {ok:true}  (soft delete)
#   GET    /api/v1/admin/resources/engines/{P.id}     -> 404 unknown resource
#   GET    /api/v1/admin/resources/engines/{C.id}     -> still parent_engine_id = P.id
```

### gombit version

### Environment

```text
go:       go1.25.7
platform: linux/amd64
```

### Database

sqlite (any)

### Auth mode

cookie (session + CSRF)

### Area

database / GORM, admin, migrations (Atlas)

### Logs or output

```text
DELETE /api/v1/admin/resources/engines/5 -> 200 {"data":{"ok":true}}
sqlite> SELECT id,name,deleted_at FROM engines WHERE id=5;
5|Tower|2026-08-28 20:18:16   # row still present, FK ON DELETE NO ACTION never triggered
GET /api/v1/admin/resources/engines/5 -> 404 {"error":{"code":"not_found","message":"unknown resource"}}
child engine 6 still has parent_engine_id = 5
```

leo-aa88:
## Summary

The admin data plane now **hard-deletes** (`Unscoped().Delete`) so the database's real foreign keys enforce referential integrity atomically (#220). `gorm.Model` made the default delete a **soft** delete — `deleted_at` is set with no physical `DELETE` — so `ON DELETE RESTRICT`/`NO ACTION` never fired and a referenced parent could be "deleted" while a live child kept pointing at an API-404 row. With a real `DELETE`:

- **RESTRICT / NO ACTION**: deleting a referenced row fails atomically → `409 conflict` (via `MapDeleteError`). The database constraint *is* the check — no detached pre-scan, no TOCTOU race.
- **CASCADE / SET NULL**: actually execute at the database.

Admin soft-delete was incidental to `gorm.Model` and never a feature (there is no restore/trash path), so physical deletion removes nothing.

Closes #220

## Acceptance criteria

- [x] soft-delete and FK/restrict are coherent — the delete is physical, so the database constraint is the enforced invariant
- [x] deleting a referenced parent returns 409; once the child is gone the parent is **physically** deleted (not left soft-deleted)

## Scope notes

- `ON DELETE CASCADE` / `SET NULL` / `SET DEFAULT` execute at the database rather than being skipped.
- Scoped to the framework-owned admin data plane; the thin generated per-resource handlers are user-owned and out of scope.
- No restore/trash path existed; hard delete removes no capability.

## Validation

- [x] `go test ./admin/...` — RESTRICT→409 then physical delete (Unscoped can't find it), CASCADE removes children, SET NULL nulls the child FK, self-referential RESTRICT; full admin suite green
- [x] `golangci-lint run ./admin/...` — 0 issues

- [x] New behavior has tests. DB-touching: a plain `DELETE` enforced by the database; covered by the SQLite admin suite and portable across the Postgres/MySQL matrix
- [x] Stable features ship docs — `docs/admin.md` "Deleting referenced records" updated to the hard-delete/DB-enforced contract
- [x] Extraction preserves contracts — replaced the guard with a narrower, DB-enforced delete; prior admin tests still pass
- [ ] Generators: N/A
- [x] API changes regenerate OpenAPI + TS client — N/A (no contract/route/shape change; same 409 envelope)
- [x] Scope stays inside the admin milestone — no M6 battery
- [x] Links its issue (#220) and states which acceptance criteria it satisfies

### DIFF
--- a/admin/resources.go
+++ b/admin/resources.go
@@ -270,65 +270,20 @@ func (h *handlers) deleteResource(ctx context.Context, input *itemInput) (*delet
 	if err != nil {
 		return nil, contract.WithContext(ctx, contract.Internal("admin database is not attached"))
 	}
-	pk, err := coercePathID(input.ID, m.pkType)
-	if err != nil {
-		return nil, contract.WithContext(ctx, contract.NotFound("unknown resource"))
-	}
-	if err := h.blockIfReferenced(ctx, db, m, pk); err != nil {
-		return nil, err
-	}
-	if err := db.WithContext(ctx).Delete(inst).Error; err != nil {
+	// Hard delete so the database's real foreign keys enforce referential
+	// integrity atomically (#220). gorm.Model makes the default Delete a SOFT
+	// delete — it sets deleted_at with no physical DELETE — so ON DELETE
+	// RESTRICT/NO ACTION never fires and a referenced parent could be "deleted"
+	// while a live child keeps pointing at an API-404 row. Unscoped issues a real
+	// DELETE, so the database enforces RESTRICT in one statement (a referenced row
+	// errors, mapped to 409 by MapDeleteError) and actually executes a declared
+	// CASCADE / SET NULL. There is no app-layer pre-scan to race: the invariant is
+	// the database constraint itself. The admin never exposed soft-delete recovery
+	// (no restore path), so this removes no feature.
+	if err := db.WithContext(ctx).Unscoped().Delete(inst).Error; err != nil {
 		return nil, database.MapDeleteError(ctx, err, "resource is still referenced by other records", "delete resource")
 	}
 	return &deleteOutput{Body: contract.Data[deleteResult]{Data: deleteResult{OK: true}}}, nil
-}
-
-// blockIfReferenced enforces ON DELETE RESTRICT at the application layer for the
-// generic admin delete. GORM soft-deletes (gorm.Model.DeletedAt) set deleted_at
-// without a physical DELETE, so the database's real foreign keys never fire and a
-// referenced parent could be "deleted" while live children keep pointing at an
-// API-404 row (#220). Before deleting, it scans every registered model for a
-// belongs_to FK targeting this model and rejects with the same 409 the
-// hard-delete path returns when any live (non-soft-deleted) row still references
-// the target. An explicit ON DELETE CASCADE / SET NULL is respected (not blocked)
-// — the author opted into that behavior.
-func (h *handlers) blockIfReferenced(ctx context.Context, db *gorm.DB, target *registered, pk any) error {
-	targetSch, err := parseSchema(target.newInstance())
-	if err != nil {
-		return contract.WithContext(ctx, contract.Internal("inspect model schema"))
-	}
-	for _, other := range h.reg.all() {
-		otherSch, err := parseSchema(other.newInstance())
-		if err != nil {
-			continue
-		}
-		for fkCol, rel := range belongsToByFK(otherSch) {
-			if rel.FieldSchema == nil || rel.FieldSchema.Table != targetSch.Table {
-				continue
-			}
-			if c := rel.ParseConstraint(); c != nil {
-				switch strings.ToUpper(strings.TrimSpace(c.OnDelete)) {
-				case "CASCADE", "SET NULL", "SET DEFAULT":
-					continue
-				}
-			}
-			q := db.WithContext(ctx).Model(other.newInstance()).
-				Where(clause.Eq{Column: clause.Column{Name: fkCol}, Value: pk})
-			if otherSch.Table == targetSch.Table {
-				// Self-referential FK: a row's reference to itself must not block
-				// its own deletion; only other rows count.
-				q = q.Where(clause.Neq{Column: clause.Column{Name: other.pkColumn}, Value: pk})
-			}
-			var n int64
-			if err := q.Count(&n).Error; err != nil {
-				return contract.WithContext(ctx, contract.Internal("check referencing records"))
-			}
-			if n > 0 {
-				return contract.WithContext(ctx, contract.Conflict("resource is still referenced by other records"))
-			}
-		}
-	}
-	return nil
 }
 
 func (h *handlers) modelForAction(
--- a/admin/resources.go
+++ b/admin/resources.go
@@ -270,10 +270,65 @@ func (h *handlers) deleteResource(ctx context.Context, input *itemInput) (*delet
 	if err != nil {
 		return nil, contract.WithContext(ctx, contract.Internal("admin database is not attached"))
 	}
+	pk, err := coercePathID(input.ID, m.pkType)
+	if err != nil {
+		return nil, contract.WithContext(ctx, contract.NotFound("unknown resource"))
+	}
+	if err := h.blockIfReferenced(ctx, db, m, pk); err != nil {
+		return nil, err
+	}
 	if err := db.WithContext(ctx).Delete(inst).Error; err != nil {
 		return nil, database.MapDeleteError(ctx, err, "resource is still referenced by other records", "delete resource")
 	}
 	return &deleteOutput{Body: contract.Data[deleteResult]{Data: deleteResult{OK: true}}}, nil
+}
+
+// blockIfReferenced enforces ON DELETE RESTRICT at the application layer for the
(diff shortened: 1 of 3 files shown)


admin/resources.go
```diff
@@ -270,12 +270,67 @@ func (h *handlers) deleteResource(ctx context.Context, input *itemInput) (*delet
...
+		if err != nil {
+			continue
+		}
+		for fkCol, rel := range belongsToByFK(otherSch) {
+			if rel.FieldSchema == nil || rel.FieldSchema.Table != targetSch.Table {
+				continue
+			}
+			if c := rel.ParseConstraint(); c != nil {
+				switch strings.ToUpper(strings.TrimSpace(c.OnDelete)) {
+				case "CASCADE", "SET NULL", "SET DEFAULT":
```
leo-aa88: ## 1. “Respecting” CASCADE/SET NULL means executing nothing

**BLOCKING.**

The target is soft-deleted by updating `deleted_at`; no physical `DELETE` reaches the database. Therefore `ON DELETE CASCADE`, `SET NULL`, and `SET DEFAULT` cannot fire. Skipping the only application-side check here leaves the child unchanged and live, pointing at a parent the API now returns as 404—the exact state this PR claims to prevent. That isn't respecting the constraint. That's constraint-themed documentation. ಠ_ಠ

Concrete example: a live child has `OnDelete:SET NULL` and references parent 7. Delete parent 7 through admin. This branch continues, GORM updates only the parent's `deleted_at`, and the child's FK remains 7 forever.

For a soft-delete target, either emulate the declared action transactionally or conservatively block. Alternatively use a physical delete and let the database perform its actual `ON DELETE` contract. Add tests for CASCADE, SET NULL, and SET DEFAULT on soft-deleted parents.

admin/resources.go
```diff
@@ -270,12 +270,67 @@ func (h *handlers) deleteResource(ctx context.Context, input *itemInput) (*delet
 	if err != nil {
 		return nil, contract.WithContext(ctx, contract.Internal("admin database is not attached"))
 	}
+	pk, err := coercePathID(input.ID, m.pkType)
+	if err != nil {
+		return nil, contract.WithContext(ctx, contract.NotFound("unknown resource"))
+	}
+	if err := h.blockIfReferenced(ctx, db, m, pk); err != nil {
```
leo-aa88: ## 2. A detached COUNT is not referential-integrity enforcement

**BLOCKING.**

`blockIfReferenced` and `Delete` are separate statements with no transaction or coordinating write invariant. A concurrent child insert can occur after the count returns zero and before the parent is soft-deleted. The database permits that insert because the parent row still physically exists; the delete then succeeds and produces a live child pointing at an API-hidden parent.

The same hole exists for writes outside generic admin because nothing requires them to participate in this preflight protocol. The code comment says “enforces,” but runtime behavior is a TOCTOU check. CI cannot rescue an invariant that the architecture does not make atomic.

Prefer a hard delete and database-enforced FKs. If soft delete is mandatory, all referencing create/update paths and deletion must coordinate through a transaction/locking design that actually excludes the race; a pre-delete scan alone cannot provide the advertised guarantee.

leo-aa88 (commented):
# REQUEST CHANGES

This patch does not enforce referential integrity under soft delete. It performs a best-effort preflight and then explicitly skips relationships whose database action can never run. CI is green; the tests cover the static RESTRICT happy path, not the contradictory actions or concurrent writes.

# VERDICT

The deepest problem is architectural: database `ON DELETE` semantics cannot be recreated by a detached `COUNT` before a soft-delete `UPDATE`. The current abstraction promises enforcement but admits dangling live references. Do not merge until deletion uses an atomic, enforceable invariant.

admin/resources.go
```diff
@@ -270,7 +270,17 @@ func (h *handlers) deleteResource(ctx context.Context, input *itemInput) (*delet
 	if err != nil {
 		return nil, contract.WithContext(ctx, contract.Internal("admin database is not attached"))
 	}
-	if err := db.WithContext(ctx).Delete(inst).Error; err != nil {
+	// Hard delete so the database's real foreign keys enforce referential
```
leo-aa88, line 273: ## 1. The PR description still promises the implementation you deleted

**BLOCKING.**

This line correctly says the admin now performs a physical hard delete. The PR summary still says it enforces RESTRICT with an application-layer scan over registered models; its acceptance criteria say a successful parent delete leaves the row soft-deleted; its scope notes say CASCADE/SET NULL are merely skipped; and its validation text names the old soft-delete guard. None of that code exists at this head.

That mismatch is especially unacceptable for deletion semantics: a maintainer reading the PR record will believe rows remain recoverable through GORM's soft-delete representation when runtime has permanently removed them. The working agreement requires the PR to state which acceptance criteria it actually satisfies.

Rewrite the title/summary, acceptance criteria, scope notes, validation, and checklist around `Unscoped().Delete`: physical deletion, database-owned RESTRICT/NO ACTION, actual CASCADE/SET NULL behavior, and no restore path. The code and `docs/admin.md` already describe the correct contract; make the PR description agree with them.

leo-aa88 (commented):
# REQUEST CHANGES

The hard-delete replacement fixes the two previous architectural defects: the database now owns the invariant atomically, and CASCADE/SET NULL actually execute. The PR's stated contract was not updated with the implementation.

# VERDICT

The remaining blocker is specification/documentation, not runtime code. Hard delete versus soft delete is an ownership and lifetime decision, not editorial trivia. Update the permanent PR record to describe the destructive semantics actually being merged; then this is mergeable.

admin/resources_test.go
```diff
@@ -765,6 +767,229 @@ func TestResourceDeleteForeignKeyViolationIsConflict(t *testing.T) {
 	assertError(t, rec, http.StatusConflict, "conflict")
 }
 
+// TestResourceDeleteSoftDeleteBlockedWhenReferenced is the #220 regression:
+// with gorm.Model soft-delete, deleting a referenced parent never triggers the
+// database's ON DELETE RESTRICT (no physical DELETE happens), which would leave
+// a live child pointing at an API-404 parent. The admin must enforce RESTRICT at
+// the application layer: a 409 while referenced, success once the child is gone.
```
leo-aa88: ## 1. This comment describes the architecture the patch removed

Lines 770–774 name a nonexistent `TestResourceDeleteSoftDeleteBlockedWhenReferenced` and say the admin “must enforce RESTRICT at the application layer.” The current test and implementation do the opposite: `Unscoped().Delete` issues a physical delete and the database enforces RESTRICT atomically. The immediately following comment already states the correct contract. Delete this stale block; leaving both makes the test claim two incompatible ownership models for the invariant.

leo-aa88 (commented):
# COMMENT

The hard-delete implementation and the updated PR contract now agree. One stale test comment still describes the discarded application-layer design.

# VERDICT

The implementation, database-owned invariant, and adversarial delete tests satisfy #220. This is mergeable; remove the contradictory stale comment so the test documents the architecture it actually exercises.

leo-aa88 (commented):
LGTM

Issue #352, opened 2026-09-13
[RESGEN-1] Model-derived, generator-owned resource contracts (supersedes contract-drift in #218)


leo-aa88:
## Thesis

**Resource contracts become model-derived and generator-owned. Human customization happens through explicit extension points (hooks/services), never by editing generated CRUD plumbing.** This is *not* "rewrite the drift checker better" — it removes the need for a drift checker by removing the multiple hand-editable representations it was trying to reconcile.

## Problem

A generated resource today has **several independently hand-editable representations of the same contract**:

```
GORM model  ·  request DTO  ·  create mapper  ·  response DTO  ·  response mapper  ·  OpenAPI schema
```

`make resource` generates a coordinated set, then **relinquishes ownership** of the handler/DTOs to the developer while still assuming invariants about them. When the model later changes, nothing keeps the representations in agreement, producing the failure modes in #218:

- a new model field is rejected on input (`422 unexpected property`),
- a field is silently dropped from responses,
- a required NOT NULL column is silently zero-filled on create,
- (and, once you look closer) a field present in every DTO but unmapped is accepted and discarded.

## Why not fix it by inspecting the generated handler (the #298 lesson)

PR #298 attempted to prove, after the fact, that the hand-owned representations still agree — by parsing Go. The review thread walked it through an escalating ladder, each fix exposing the next semantic layer:

```
DTO membership → composite-literal presence → named constructor → constructor+Create occurrence
→ ordered syntax → alias checks → exact grammar → Huma schema reflection → mapper-arrow analysis …
```

That is the architectural verdict, not reviewer pedantry: **once a check needs JSON tags, GORM column names, constructor/response-mapper RHS expressions, receiver identity, call ordering, aliases, pointer receivers, `defer`, branches, server-managed mutations, relationships, and custom persistence calls, it is no longer a drift guard — it is an ad-hoc static analyzer for arbitrary Go.** The generator cannot give away ownership of CRUD plumbing and later recover its invariants by inspection.

#298 is therefore closed as an architectural dead end (it earned that conclusion), **not as "fixed."** This issue supersedes the **broad contract-drift** scope of #218.

## Proposed direction: model-first generation with an ownership boundary

Have **one authoritative representation and derive the others.** The semantics to lock in (surface syntax to be decided in the ADR — the example below is illustrative, not final):

```
Source of truth:
  GORM schema owns:   DB-backed / relationship / PK / column name / nullability / default
[...]

PR #354, opened 2026-09-13
feat(resourcepolicy): field-policy tag + resolver — RESGEN-1 slice 1 (#352)

leo-aa88:
## Summary

**RESGEN-1 slice 1 of #352** (model-first resource generation, [ADR-016](../blob/main/docs/adr/016-model-first-resource-generation.md)) — the foundational, pure policy layer everything else derives from. No generation yet; nothing is wired into `make resource`.

Adds package `resourcepolicy`:

- The **`gombit` field-policy tag**: `read` (in response), `write` (settable via create request), `server` (set server-side in a hook), `-` (API-hidden). Surface syntax per ADR-016 (semantics locked, spelling not fossilized).
- `Resolve(FieldFacts, tag)` layers API behavior over **schema-owned** persistence facts (DB-backed / PK / auto-time / soft-delete / column). Persistence facts are supplied by the caller (slice 2 reads them from the model); policy never restates them.
- **Kind defaults**: a content column and a **manual** primary key → read + write from the request; an **auto-increment** key and auto timestamps → read-only; soft-delete → hidden; a relationship field → not a DTO column.
- **Fails closed on contradictions**: `write`+`server`, writing/server-setting a GORM-managed or non-creatable column, `read` on a non-readable column, a required column with no create source, `-` combined with `read`/`write` (note `-,server` is allowed — hidden but hook-set), empty/unknown tokens.
- `RequestColumns` / `ResponseColumns` / `ServerColumns` expose the derived surfaces that slices 3–4 consume.

## Why this shape

Per ADR-016: one owner per fact. The GORM schema owns persistence facts; `gombit` policy owns only API behavior. This slice is the boundary in code — pure, deterministic, unit-tested — with no AST detective work and no wiring, so it can't regress `make resource` while the rest of the epic lands on top.

## Validation

- [x] `go test ./resourcepolicy/...` — defaults, hidden, write-only/server directions, manual-vs-auto keys, timestamp/soft-delete (incl. pointer `gorm.DeletedAt`), GORM `->`/`<-` permissions, required-source validation, empty/unknown-token rejection, and every contradiction; `ResolveAll` + column-surface helpers + a `gorm/schema.Parse`+SQLite adapter (`canonicalFields`) proving tags/permissions survive and resolver ⟺ runtime.
- [x] `golangci-lint run ./resourcepolicy/...` — 0 issues.

## Scope notes

- Foundational slice only. Slice 2 reads real model fields/tags into `FieldFacts`; slices 3–4 derive DTOs/mappers/handler+hooks; slice 5 adds `gombit generate --check`; slice 6 removes `resourcecheck` and writes the breaking-change migration guide. Tracked in #352.

- [x] New behavior has tests — pure unit tests for the whole resolver
- [x] Stable features ship docs — package doc comment specifies the tag semantics and defaults; ADR-016 already merged. User-facing CLI docs follow when the tag is wired (slice 2+)
- [x] Extraction preserves contracts — additive new package; nothing else touched
- [x] Scope inside the RESGEN-1 epic (#352, post-v0.1); no M6 battery
- [x] Links its issue (#352) and states which slice it is


resourcepolicy/policy.go
```diff
@@ -0,0 +1,217 @@
...
+	"strings"
+)
+
+// FieldFacts are the persistence facts about a model field, owned by the GORM
+// schema. Policy never restates these; it only layers API behavior on top.
+type FieldFacts struct {
+	GoName     string // Go field name, e.g. "Title"
+	Column     string // DB column name, e.g. "title"; empty for a relationship/association field
+	DBBacked   bool   // maps to a database column (false for has_many/many2many/belongs-to object fields)
+	PrimaryKey bool
```
leo-aa88: ## 1. `PrimaryKey` is not an auto-generation fact

**BLOCKING.**

This collapses independent GORM facts—`PrimaryKey`, `AutoIncrement`, and `HasDefaultValue`—and `autoManaged()` then treats every key as database/GORM-filled. That is false for an ordinary manual key. GORM parses a string `ID` carrying `gorm:"primaryKey" gombit:"read,server"` as primary-key=true, auto-increment=false, has-default=false; `Resolve` nevertheless rejects `server` (and `write`), leaving the generated create path with no way to supply the key.

The package contract also says nullability/default are schema-owned facts, but `FieldFacts` cannot represent either. Consequently a required string carrying `gorm:"not null" gombit:"read"` resolves to `CreateSourceNone`; generated create plumbing will leave the Go zero value and reproduce the silent required-column zero-fill that #352 exists to eliminate. What exactly is this resolver guaranteeing if it cannot distinguish safe omission from missing data?

Carry auto-increment/default/nullability (or the equivalent safe-omission facts) separately. Base auto-management on actual generation behavior, allow an explicit request/server source for manual keys, and reject non-auto required fields with no request/server/default source. Add tests for manual string/UUID keys and required read-only/hidden columns.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return r, nil
+}
+
+// ResolveAll resolves every field, returning results in the input order. It fails
+// on the first contradictory field so the generator never emits from a bad policy.
+func ResolveAll(fields []FieldFacts, tags map[string]string) ([]Resolved, error) {
+	out := make([]Resolved, 0, len(fields))
+	for _, f := range fields {
+		r, err := Resolve(f, tags[f.GoName])
```
leo-aa88: ## 2. Bare Go names do not identify schema fields

**BLOCKING.**

`tags[f.GoName]` loses the field-to-tag association after GORM flattens embedded structs. This is reproducible with one embedded struct containing a `Code` field carrying `gorm:"column:public_code" gombit:"read"` and another containing a `Code` field carrying `gorm:"column:secret_code" gombit:"-"`. GORM exposes two DB-backed fields whose `Name` is `Code` but whose binding paths, columns, and policies differ. This map can hold only one value, so `ResolveAll` applies that policy to both: either the secret field is exposed or the public field disappears.

Keep each raw tag paired with its exact field facts, or key it by GORM's unambiguous binding path and reject duplicate/unused entries. Add an embedded-name-collision test before slice 2 builds model reading on this API.

leo-aa88 (commented):
# REQUEST CHANGES

This policy layer cannot serve as the authoritative model-to-generator boundary yet. It discards both schema source facts and field identity before resolution, so later slices would be forced either to guess or to build a second validation mechanism around it.

## 1. The schema representation cannot determine create ownership

**BLOCKING.**

`PrimaryKey` is treated as synonymous with auto-generated even though GORM represents `PrimaryKey`, `AutoIncrement`, and `HasDefaultValue` independently. The same representation also omits the nullability/default facts ADR-016 says the schema owns, so it cannot reject a required field whose policy provides no create source.

## 2. `ResolveAll` keys policy by a non-unique field name

**BLOCKING.**

GORM preserves multiple flattened embedded fields with the same `Name` but different binding paths and columns. A `map[string]string` keyed by `GoName` cannot represent their distinct tags, and this function silently applies one field's policy to both.

# VERDICT

The deepest problem is the abstraction: the supposedly authoritative resolver accepts a lossy schema projection. The existing tests prove the intended happy path, but they do not exercise real GORM states that the next slice must feed into this API.

Do not merge this foundation until it preserves actual generation/default facts and exact field identity, validates missing create sources, and adds adversarial coverage for manual primary keys, required non-request fields, and embedded-name collisions. The functional CI jobs are green; the red benchmark-smoke job is an unrelated Composer/GitHub 504 failure.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,255 @@
...
+)
+
+// FieldFacts are the persistence facts about a model field, owned by the GORM
+// schema. Policy never restates these; it only layers API behavior on top. The
+// facts are kept independent (not collapsed) because create-source correctness
+// depends on their exact combination.
+type FieldFacts struct {
+	GoName        string // Go field name, e.g. "Title"
+	Column        string // DB column name, e.g. "title"; empty for a relationship/association field
+	DBBacked      bool   // maps to a database column (false for has_many/many2many/belongs-to object fields)
```
leo-aa88: ## 1. `DBBacked` does not imply `Creatable` or `Readable`

**BLOCKING.**

This still is not a lossless projection of the GORM facts needed to honor the tag contract. GORM exposes `schema.Field.Creatable` and `Readable` independently of `DBName`. For a common computed/read-only field carrying `gorm:"->"`, GORM reports a real DB column with `Creatable=false`; these facts map to this struct exactly like an ordinary content column, so an empty `gombit` tag resolves to `InRequest=true` and `CreateSourceRequest`. GORM's create callback calls `SelectAndOmitColumns(true, false)` and explicitly omits every field with `Creatable=false`. The API therefore accepts and maps a value that persistence discards. That is the original drift bug in a new descriptor.

The inverse is broken too: `gorm:"->:false;<-:create"` is creatable but `Readable=false`. An explicit `read,write` policy is accepted and emits a response field, while GORM query scans skip non-readable fields, so later GET/list responses can only expose a zero value. The code says `read` and `write`; runtime does neither operation as advertised.

Carry at least the schema's create/read capabilities in `FieldFacts`. Default non-creatable fields to no request source, reject `write`/`server` when GORM will omit them from `Create`, and reject `read` when queries cannot populate them. Add adversarial tests built from `gorm/schema.Parse` for `->`, `<-:update`, and `->:false;<-:create` so this boundary is tested against the actual schema semantics slices 2–4 will consume.

leo-aa88 (commented):
# REQUEST CHANGES

The two prior blockers are corrected: manual/generated keys are now distinct, required no-source columns are rejected, and policy remains paired with the exact field. The resolver still loses GORM persistence-access permissions, so its public `read`/`write` contract can disagree with what GORM actually stores and loads.

## 1. `DBBacked` does not imply `Creatable` or `Readable`

**BLOCKING.**

GORM represents these as independent schema facts. The current projection cannot distinguish an ordinary column from a read-only/computed column or a create-only/non-readable column, yet `Resolve` derives request and response surfaces as if all DB-backed content columns support both operations.

# VERDICT

The remaining problem is still the abstraction boundary: schema information is discarded before policy resolution, forcing the generated contract to promise behavior the persistence layer will not perform. This is exactly the class of accepted-but-discarded or surfaced-but-unpopulated field drift RESGEN-1 is supposed to eliminate.

Do not merge until GORM create/read capabilities survive into `FieldFacts`, contradictory policies fail closed, defaults respect those capabilities, and real GORM permission-tag cases are covered. All completed CI jobs are green; benchmark smoke and SQLite migrations were still running at review time.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,286 @@
...
+	return f.AutoIncrement || f.AutoTime || f.SoftDelete
+}
+
+// requiresCreateValue reports whether the column must be given a value on create
+// or it will be silently zero-filled: NOT NULL (a primary key is implicitly so),
+// no database default, not auto-generated, and one GORM actually writes on INSERT
+// (Creatable). A non-creatable column (`gorm:"->"`) is filled by the DB/GORM, not
+// by us, so we make no create-source demand on it.
+func (f FieldFacts) requiresCreateValue() bool {
+	if !f.DBBacked || !f.Creatable {
```
leo-aa88: ## 1. Non-creatable does not mean database-generated

**BLOCKING.**

This early return destroys the required-column invariant. `schema.Field.Creatable=false` means GORM omits the column from INSERT; it does **not** mean the database or GORM supplies a value. A model field such as `Code string` carrying `gorm:"not null;<-:update"` produces `DBBacked=true`, `Creatable=false`, `NotNull=true`, `HasDefault=false`, and no auto-generation fact. `Resolve(facts, "")` accepts it because this branch says no create value is required. At runtime GORM emits `INSERT INTO models DEFAULT VALUES`, and SQLite returns `NOT NULL constraint failed: models.code`. I reproduced that exact path.

You fixed the accepted-but-discarded request case and replaced it with an accepted-but-impossible model state. The comment above is simply false: non-creatable says who may write, not who generates. Compute requiredness from `NotNull/PrimaryKey`, `HasDefault`, and actual auto-generation independently. If such a field is non-creatable, reject the model because neither request nor server can supply it; if it is creatable, retain the existing no-source check. Add a GORM+SQLite test for a NOT NULL `<-:update` field with no default, plus the analogous manual-key state.

leo-aa88 (commented):
# REQUEST CHANGES

The prior GORM access-permission blocker is mostly addressed: request/response defaults now honor `Creatable`/`Readable`, and contradictory tags fail closed. But the fix incorrectly equates non-creatable with database-generated, so it disables required-column validation for models the generated create path can never persist.

## 1. Non-creatable does not mean database-generated

**BLOCKING.**

`Creatable=false` only tells GORM to omit a field from INSERT. Without an auto-generation mechanism or database default, a required column is unsatisfiable. The resolver currently accepts that state and the database rejects every create.

# VERDICT

This is an implementation and invariant defect in the new permission handling. The abstraction now carries the necessary facts, but combines them incorrectly: omission capability is being treated as value generation. That contradicts the runtime and bypasses the exact fail-closed guarantee RESGEN-1 exists to provide.

Do not merge until non-creatable required columns without an actual default/auto-generation fact are rejected, and the behavior is covered through a real GORM create test. CI was still running at review time; the focused package tests, race test, vet, and lint pass because this malformed cross-fact state is not exercised.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,297 @@
...
+	Creatable bool
+	Readable  bool
+}
+
+// autoGenerated reports whether the database or GORM fills the column itself, so
+// it is never settable through the create request and needs no create source. A
+// primary key is auto-generated only when it is an auto-increment/identity key —
+// a manual string/UUID key is not.
+func (f FieldFacts) autoGenerated() bool {
+	return f.AutoIncrement || f.AutoTime || f.SoftDelete
```
leo-aa88, line 80: ## 1. Managed fields are not necessarily generated during create

**BLOCKING.**

This predicate is still doing two incompatible jobs: (1) deciding that request/server writes are forbidden, and (2) deciding that a required column is guaranteed to receive a valid value on INSERT. The second claim is false for two GORM states I reproduced end-to-end.

- `CreatedAt time.Time` with `gorm:"not null;autoCreateTime;<-:update"` parses as `AutoTime=true`, `Creatable=false`, `NotNull=true`, and no default. `Resolve` enters the auto-generated branch and returns success. GORM omits the field because it is non-creatable, emits `INSERT ... DEFAULT VALUES`, and SQLite rejects it with `NOT NULL constraint failed: models.created_at`. GORM can only apply auto-time while writing the field.
- `DeletedAt gorm.DeletedAt` with `gorm:"not null"` parses as soft-delete and no default. `Resolve` again returns success, but GORM inserts `NULL` and SQLite rejects `NOT NULL constraint failed: models.deleted_at`. Soft-delete is managed on deletion; that does not generate a valid create-time value.

Split “policy-managed / never user-settable” from “has a valid persistence create source.” Auto-increment is DB-generated; auto-time only supplies a create value when GORM can write it; soft-delete ordinarily relies on nullability/default and must not waive a contradictory NOT NULL constraint. Run requiredness validation before returning from this branch and add the two GORM+SQLite cases above. Otherwise the resolver still approves models whose generated create path cannot work.

leo-aa88 (commented):
# REQUEST CHANGES

The previously reported non-creatable required-column case is fixed and now covered through the real database path. One related blocker remains: `autoGenerated()` still conflates fields that policy must protect from user writes with fields guaranteed to receive a valid value during INSERT. GORM does not make that guarantee for every auto-time or soft-delete combination.

## 1. Managed fields are not necessarily generated during create

**BLOCKING.**

The auto-generated branch returns before requiredness validation. A non-creatable auto-time field and a required soft-delete field both reach runtime without a valid inserted value, despite this resolver accepting them.

# VERDICT

This is an abstraction and invariant defect. One predicate answers two materially different questions—whether the API may source a value and whether persistence supplies a valid create-time value—and the resulting early return bypasses the fail-closed contract.

Do not merge until those concepts are separated and requiredness is validated against the actual create-time source for auto-time and soft-delete fields. CI was still running at review time; focused tests, race tests, vet, and lint pass because these cross-fact combinations are absent from the suite.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,294 @@
...
+// time or the INSERT is impossible / silently zero-filled: NOT NULL (a primary
+// key is implicitly so), no database default, and not supplied by the DB/GORM.
+// It is INDEPENDENT of Creatable — Creatable governs whether GORM writes a value
+// the API provides, not whether one exists. A required column that neither the DB
+// supplies nor the API can source is unsatisfiable; Resolve reports how.
+func (f FieldFacts) requiresCreateValue() bool {
+	if !f.DBBacked {
+		return false
+	}
+	return (f.NotNull || f.PrimaryKey) && !f.HasDefault && !f.dbSuppliesCreateValue()
```
leo-aa88, line 114: **BLOCKING.** `HasDefault` is not evidence that the default supplies a valid create value. GORM parses a field tagged `gorm:"not null;default:null;<-:update"` as `NotNull=true`, `HasDefaultValue=true`, `DefaultValue="null"`, and `Creatable=false`. This predicate therefore says no create value is required, and `Resolve` accepts the field with `CreateSourceNone`. Against an actual GORM+SQLite database, `Create` then emits `INSERT INTO widgets DEFAULT VALUES RETURNING code,id` and fails with `NOT NULL constraint failed: widgets.code`.

The new invariant still stops one layer short: a default can itself be unusable for a required column. `FieldFacts` has already collapsed the relevant schema fact to a boolean, so `Resolve` cannot distinguish a non-null default from `DEFAULT NULL`. Preserve enough default semantics (or an explicit schema-derived "default supplies a non-null create value" fact), reject this state, and add the same resolver ↔ runtime regression test used for the other unsatisfiable combinations. CI is green; the tests simply do not exercise this contract.

leo-aa88 (commented):
# REQUEST CHANGES

The previous managed-vs-create-source defect is fixed, and the new tests cover the exact auto-time and soft-delete failures. The default branch now repeats the same mistake one fact later: it equates “a default exists” with “persistence supplies a valid required value.” Those are not the same contract.

## 1. A database default can still leave a required column unsatisfied

**BLOCKING.**

`requiresCreateValue` exempts every `HasDefault` column without preserving or inspecting the default. GORM parses `gorm:"not null;default:null;<-:update"` as `NotNull=true`, `HasDefaultValue=true`, `DefaultValue="null"`, and `Creatable=false`. The resolver therefore accepts the field with no create source. A real GORM+SQLite create emits `INSERT ... DEFAULT VALUES` and fails `NOT NULL constraint failed`.

The claimed contract is stronger: accepted required fields must receive a valid value during create. `HasDefault` only proves that a default clause exists. It does not prove that the clause satisfies `NOT NULL`. Preserve enough schema information to distinguish a usable default from a known `NULL` default (or carry the already-resolved “default supplies a non-null create value” fact), reject the contradictory state, and cover it through the same resolver-to-runtime test used for the other unsatisfiable combinations.

# VERDICT

The remaining defect is in the abstraction. `FieldFacts` collapses default semantics before the policy layer asks the exact question it exists to answer, so the resolver can certify a create path that the database rejects.

Do not merge until requiredness is based on a valid create-time default, not merely the presence of a default. CI is fully green. This is not a failing-test problem; the current tests do not exercise this contract.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,304 @@
...
+	return f.AutoIncrement || (f.AutoTime && f.Creatable)
+}
+
+// hasUsableDefault reports whether the column's DB default actually supplies a
+// value that satisfies NOT NULL. A `default:` clause exists (HasDefault) is not
+// enough — an explicit `default:null` provides no value, so GORM still emits
+// DEFAULT VALUES and the constraint fails. Any non-"null" default is treated as
+// usable (including an empty-string default).
+func (f FieldFacts) hasUsableDefault() bool {
+	return f.HasDefault && !strings.EqualFold(strings.TrimSpace(f.DefaultValue), "null")
```
leo-aa88: **BLOCKING.** This is still syntax pretending to be semantic validation. Comparing the raw expression only with `null` rejects one spelling, then declares every other SQL expression usable. GORM preserves `gorm:"not null;default:(NULL);<-:update"` as `HasDefaultValue=true`, `DefaultValue="(NULL)"`, `Creatable=false`; this function returns true, `Resolve` accepts `CreateSourceNone`, and GORM+SQLite emits `INSERT INTO widgets DEFAULT VALUES RETURNING code,id` before failing `NOT NULL constraint failed: widgets.code`. `default:(coalesce(NULL,NULL))` follows the same accepted-then-fails path, so stripping one pair of parentheses is not an architectural fix.

A raw SQL expression cannot prove that a non-null value reaches INSERT. Either scope this policy to create-source ownership—where a DB default is an external source and this layer stops claiming it validates the expression's result—or represent default validity as a closed state (`known non-null`, `known null`, `unknown`) and fail closed for unknown expressions when no request/server source exists. Add the parenthesized/expression runtime case. CI can stay green because the current test exercises only the exact token this comparison was written around.

leo-aa88 (commented):
# REQUEST CHANGES

The exact `default:null` regression is now rejected and covered through GORM+SQLite. The fix does not close the contract: it compares one raw spelling while claiming to determine whether an arbitrary SQL default actually supplies a non-null value.

## 1. Raw default text cannot prove a valid create value

**BLOCKING.**

`hasUsableDefault` returns true for every expression except the exact case-insensitive token `null`. GORM preserves `default:(NULL)` as `DefaultValue="(NULL)"`, so the resolver accepts a required, non-creatable field with `CreateSourceNone`. SQLite evaluates that default to NULL and rejects the resulting `INSERT ... DEFAULT VALUES` on the `NOT NULL` constraint. The same mismatch is reproducible with `default:(coalesce(NULL,NULL))`.

The code says it knows whether the default “actually supplies a value that satisfies NOT NULL”; runtime says otherwise. A raw, dialect-specific SQL expression cannot support that abstraction. Either limit this layer to source ownership and stop claiming it validates the default's result, or introduce a closed validity state with an explicit unknown case and fail closed when a required field has no other source. Do not grow a parenthesis-stripping SQL evaluator here.

# VERDICT

The remaining problem is abstraction plus adversarial coverage. The previous boolean lost too much information; replacing it with an unchecked SQL string and one lexical comparison still does not establish the promised invariant.

Do not merge until the default contract is scoped to facts this pure layer can actually know, or unknown/default-expression states are represented and handled honestly. Focused tests, race tests, and vet pass; all completed CI jobs are green, with several matrix jobs and benchmark smoke still running at review time. The current tests prove only the spelling the implementation special-cases.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,307 @@
...
+	tag, err := parsePolicyTag(gombitTag)
+	if err != nil {
+		return Resolved{}, err
+	}
+	r := Resolved{FieldFacts: facts, CreateSource: CreateSourceNone}
+
+	// A relationship/association field is not a scalar wire column; it is not part
+	// of the derived request/response DTOs (belongs-to exposes its FK column, which
+	// is itself DB-backed and handled as a normal field).
+	if !facts.DBBacked {
```
leo-aa88: **BLOCKING.** `DBBacked` and `Column` encode the same relationship/column distinction independently, but `Resolve` never enforces that they agree. Through the public API, `FieldFacts{GoName:"Ghost", DBBacked:true, Column:"", Creatable:true, Readable:true}` resolves successfully with `InRequest=true`, `InResponse=true`, and `CreateSourceRequest`; `RequestColumns` and `ResponseColumns` then both return empty because `columnsWhere` drops the empty column. The authoritative field policy now gives two incompatible answers depending on which public surface the next generator slice consumes.

This is exactly the malformed-state boundary a foundational resolver must close. Remove redundant `DBBacked` and derive it from `Column != ""`, or validate the equivalence before any policy resolution (and reject DB-only facts on a non-DB-backed field). Add a test proving `Resolve` cannot produce an included field that the column helpers silently discard.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,307 @@
...
+		switch strings.TrimSpace(tok) {
+		case "read":
+			p.read = true
+		case "write":
+			p.write = true
+		case "server":
+			p.server = true
+		case "-":
+			p.hidden = true
+		case "":
```
leo-aa88, line 186: **MAJOR.** This accepts every empty token, not merely one trailing comma. `gombit:","` sets `present=true` but no policy flags, so a normal nullable field silently resolves fully hidden with no error—an undocumented alternate spelling for `-`. `gombit:"read,,write"` and leading commas are also accepted. A typo in the authoritative tag grammar must not quietly change the generated API contract. Validate token position: if trailing-comma compatibility is intentional, permit only one final empty token after at least one real token; reject leading, interior, repeated, and all-empty token lists, with adversarial tests.

resourcepolicy/policy_test.go
```diff
@@ -0,0 +1,408 @@
...
+	if f == nil {
+		t.Fatalf("field %q not found in schema", goName)
+	}
+	return resourcepolicy.FieldFacts{
+		GoName:        f.Name,
+		Column:        f.DBName,
+		PrimaryKey:    f.PrimaryKey,
+		AutoIncrement: f.AutoIncrement,
+		AutoTime:      f.AutoCreateTime != 0 || f.AutoUpdateTime != 0,
+		SoftDelete:    f.Name == "DeletedAt",
```
leo-aa88: **MAJOR.** This is not how GORM identifies a soft-delete column, and the comment on `factsOf` says this is the mapping slice 2 will use.

GORM attaches soft-delete through the `gorm.DeletedAt` type (`QueryClauses` / `DeleteClauses`), not the Go name. I parsed:

```go
RemovedAt gorm.DeletedAt
```

and got `DBName="removed_at"`, `Creatable=true`, `DeleteClauses=1`, `QueryClauses=1`. This line reports `SoftDelete=false`. Default `Resolve` then treats it as a normal content column (`InRequest=true`) — the opposite of “soft-delete is hidden.”

Same helper looks the field up in `FieldsByName`, which is the non-unique key `ResolveAll` was already rewritten to stop using.

Detect the type/clauses GORM actually uses, pass `*schema.Field` (or iterate `schema.Fields`), and do not advertise this as the slice-2 mapping until it is lossless.

resourcepolicy/policy.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+	}
+	if tag.read && !facts.Readable {
+		return Resolved{}, fmt.Errorf("resourcepolicy: column %q is not readable (GORM skips it on SELECT) and cannot be in the response", facts.Column)
+	}
+
+	// Surfaces + create source. Defaults respect capabilities: readable and not
+	// soft-delete → response; creatable and not GORM-managed → request.
+	if tag.present {
+		r.InResponse = tag.read
```
leo-aa88, line 248: **MAJOR.** These two assignments are the actual contract (`read` and `write` are independent). The tests never hit them in the `write`-only or bare-`server` directions.

Covered: untagged defaults, `read`, `read,server`, `-,server`, `-`, and GORM `->:false` write-only. Missing: `gombit:"write"` (password / write-only) and `gombit:"server"` (hook-set, not in the response).

An implementation that did `InResponse = tag.read || tag.write` would still pass. Add those two cases; they are the grammar, not decorations.

leo-aa88 (commented):
# COMMENT

The previous blockers are actually gone. Persistence facts are no longer collapsed, create-source and GORM-managed are separate questions, `Column == ""` is the only relationship/non-column state, empty tag tokens fail closed, and the default-SQL retreat is honest: this layer owns source assignment, not expression evaluation.

What remains is not another invented invariant in `Resolve`. It is that the only GORM→`FieldFacts` mapping in the tree still lies about two facts later slices will copy, and the tag grammar's independent `write` / `server` directions are still untested.

## 1. The documented GORM mapping is still lossy

**MAJOR.**

`factsOf` is commented as “the mapping slice 2 will use.” It is not a lossless projection of the schema this package claims to consume.

- Soft-delete is detected by `f.Name == "DeletedAt"`. GORM attaches soft-delete via the `gorm.DeletedAt` type (`QueryClauses` / `DeleteClauses`), not the field name. `RemovedAt gorm.DeletedAt` parses with `DBName="removed_at"`, `Creatable=true`, and live delete/query clauses. This helper reports `SoftDelete=false`, so default `Resolve` puts the column on the create request — the opposite of the kind default the package documents.
- Lookup is `sch.FieldsByName[goName]`. That is the same non-unique key `ResolveAll` was rewritten to stop using. GORM flattens embedded structs onto one Go name; `FieldsByName` cannot feed this resolver.

THE COMMENT SAYS: this is how slice 2 will read the schema.

THE SCHEMA SAYS: otherwise.

If the next PR copies this helper, renamed soft-delete columns become writable DTO fields and embedded collisions collapse again. Detect soft-delete from the type/clauses GORM actually uses, take `*schema.Field` (or iterate `schema.Fields`), and stop advertising this as the slice-2 mapping until it is one.

## 2. Independent `write` and bare `server` tags are untested

**MAJOR.**

The grammar claims `read`, `write`, and `server` are independent. The implementation does the right thing:

```go
r.InResponse = tag.read
r.InRequest = tag.write
```

The tests never ask for `gombit:"write"` (password / write-only) or bare `gombit:"server"` (hook-set, not in the response). Covered instead: untagged defaults, `read`, `read,server`, `-,server`, hidden, and GORM `->:false` write-only. An implementation that set `InResponse = tag.read || tag.write` would still be green.

CI is green (Test, Lint, Build, SQLite/Postgres/MySQL). This is not a failing-test problem. The current tests do not exercise those tag directions.

# VERDICT

The remaining problem is tests and the mapping sketch, not the resolver's architecture. `Resolve` now matches the contract it states: schema facts stay independent, contradictions fail closed, and SQL defaults are not pretended to be evaluated.

Fix the helper so it cannot be copied into slice 2 as-is, and add the two missing tag-direction cases. I would not block merge of the production package on those, but I also will not LGTM a foundation whose only GORM adapter still disagrees with GORM.

resourcepolicy/policy_test.go
```diff
@@ -0,0 +1,465 @@
...
+// a `RemovedAt gorm.DeletedAt` field is handled correctly. Slice 2 can iterate
+// sch.Fields and reuse this mapping as-is.
+func factsOf(f *schema.Field) resourcepolicy.FieldFacts {
+	return resourcepolicy.FieldFacts{
+		GoName:        f.Name,
+		Column:        f.DBName,
+		PrimaryKey:    f.PrimaryKey,
+		AutoIncrement: f.AutoIncrement,
+		AutoTime:      f.AutoCreateTime != 0 || f.AutoUpdateTime != 0,
+		SoftDelete:    f.FieldType == reflect.TypeOf(gorm.DeletedAt{}),
```
leo-aa88: **MAJOR.** This is still not how GORM detects soft delete. GORM instantiates `reflect.New(f.IndirectFieldType)` and checks the clause interfaces; it does not compare `FieldType`. Consequently `RemovedAt *gorm.DeletedAt` gets GORM query/delete clauses but maps here as `SoftDelete=false`, and the resolver defaults it onto request/response surfaces. Use the indirect type at minimum (or the same capability check GORM uses) and add the pointer case.

resourcepolicy/policy_test.go
```diff
@@ -0,0 +1,465 @@
...
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/schema"
+)
+
+// factsOf maps a parsed GORM *schema.Field to FieldFacts. It is a lossless,
+// name-independent projection — soft-delete is detected by the gorm.DeletedAt
+// TYPE (how GORM attaches its query/delete clauses), not by the Go field name, so
+// a `RemovedAt gorm.DeletedAt` field is handled correctly. Slice 2 can iterate
+// sch.Fields and reuse this mapping as-is.
```
leo-aa88: **MAJOR.** `sch.Fields` is not the canonical persisted-field set. GORM keeps shadowed embedded fields here even when several map to the same DB column; it chooses the effective field through `FieldsByDBName` and `DBNames`. Two embedded `Code` fields both mapped to `code` will therefore be resolved twice by the suggested iteration although GORM persists only one. The existing collision test uses distinct columns and misses this. Define iteration over effective DB fields (while separately preserving relationships/binding paths) and cover same-column shadowing.

leo-aa88 (commented):
# COMMENT

The resolver itself now keeps the relevant policy facts separate and rejects the previously demonstrated contradictions. The final test-only GORM projection still overclaims fidelity, however: it disagrees with GORM's own field and soft-delete machinery in two concrete states that slice 2 will have to handle.

## 1. Soft-delete detection does not match GORM's dispatch

**MAJOR.**

The helper recognizes only an exact, non-pointer `gorm.DeletedAt`:

```go
f.FieldType == reflect.TypeOf(gorm.DeletedAt{})
```

The comment claims this is how GORM attaches query/delete clauses. It is not. GORM instantiates `reflect.New(f.IndirectFieldType)` and dispatches clause interfaces. Therefore `RemovedAt *gorm.DeletedAt` receives GORM's soft-delete query/delete behavior while this projection reports `SoftDelete=false`; default resolution then puts that lifecycle column in request and response surfaces.

Use the indirect type at minimum for core `gorm.DeletedAt`, or detect the same clause capability GORM uses. Add the pointer case so the test proves the claimed boundary.

## 2. `schema.Fields` is not the canonical persisted-field set

**MAJOR.**

The new comment tells slice 2 to iterate `sch.Fields`. GORM retains every parsed field there, including embedded fields shadowed by another field with the same DB column. It separately selects the effective field in `FieldsByDBName` and records unique persisted names in `DBNames`.

For two embedded `Code` fields both mapped to `code`, iterating `sch.Fields` feeds both to `ResolveAll`, while GORM persists only the prioritized field. The next generator would emit duplicate or contradictory DTO/mapping state for one actual column. The existing collision test uses different columns, so it cannot catch this.

Define the caller contract around GORM's effective DB fields, for example ordered `DBNames` plus `FieldsByDBName`, while separately preserving relationships and binding paths. Add a same-column embedded-shadowing test.

# VERDICT

This is a mapping/test-boundary defect, not a defect in `Resolve`'s current policy logic. CI is green, and focused tests, race tests, and vet pass. That does not validate the two GORM states above; the current tests encode the same incorrect projection they claim to verify.

The resolver slice is otherwise mergeable, but the advertised GORM adapter is not complete or reusable as-is. Correct these two claims before slice 2 builds production model reading on them.

resourcepolicy/policy_test.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+}
+
+// canonicalFields maps a schema to resolver input the way slice 2 should: over
+// GORM's effective persisted columns (ordered sch.DBNames → sch.FieldsByDBName),
+// so a duplicate-column or shadowed field is resolved once, not twice.
+func canonicalFields(sch *schema.Schema) []resourcepolicy.Field {
+	out := make([]resourcepolicy.Field, 0, len(sch.DBNames))
+	for _, name := range sch.DBNames {
+		out = append(out, resourcepolicy.Field{FieldFacts: factsOf(sch.FieldsByDBName[name])})
```
leo-aa88: **MAJOR.** This helper is now advertised as the schema-to-resolver adapter slice 2 should copy, but it discards the API policy: every returned `resourcepolicy.Field.Tag` is empty. A model field carrying `gombit:"-"` or `gombit:"read,server"` therefore reaches `ResolveAll` as untagged and gets kind defaults instead. The persistence-field selection is now correct; the end-to-end adapter is not. Either explicitly limit this helper/comment to canonical persistence-fact selection, or populate `Tag` from the selected `schema.Field.Tag` and add a schema-derived test proving distinct policies survive into resolution.

leo-aa88 (commented):
# COMMENT

The two previously reported GORM-fidelity defects are fixed: pointer soft-delete fields now follow `IndirectFieldType`, and persisted-field enumeration now follows GORM's effective `DBNames` / `FieldsByDBName` set. The replacement adapter still stops one layer short of its own claim, and the PR description no longer describes HEAD accurately.

## 1. The canonical schema adapter discards every `gombit` policy

**MAJOR.**

`canonicalFields` is documented as mapping a schema to the resolver input slice 2 should use, but it constructs:

```go
resourcepolicy.Field{FieldFacts: factsOf(sch.FieldsByDBName[name])}
```

`Field.Tag` is therefore empty for every field. Given:

```go
Secret string `gombit:"-"`
```

the advertised adapter feeds `ResolveAll` an untagged content column, so it becomes request-writable and response-readable. The persistence selection is now faithful; the API policy vanishes before resolution. That is the exact kind of represented-at-one-layer, ignored-at-the-next failure this package exists to prevent.

Either narrow the helper/comment explicitly to canonical persistence-field selection, or carry the selected schema field's `gombit` tag and add a real schema-derived test proving distinct tags survive into the resolved surfaces.

## 2. The PR description contradicts the implementation

**MAJOR.**

The PR still claims:

- “primary key ... → read-only,” while HEAD intentionally makes manual string/UUID primary keys writable by default and only auto-increment keys read-only;
- “`-` combined with other tokens” fails, while HEAD intentionally permits `-,server`.

Those are not cosmetic details; they are the public policy grammar and defaults under review. Update the description so the claimed contract matches the code.

# VERDICT

The production resolver is no longer the problem. The remaining defects are in the mapping/test boundary and review contract: the new “canonical” adapter drops the very policy being resolved, and the PR description advertises superseded semantics.

Focused tests, race tests, vet, lint, and build pass locally/CI; the full Test job is still running. Do not treat this slice as fully reviewed until the adapter claim is made honest and the PR description matches HEAD.

leo-aa88 (commented):
LGTM

PR #355, opened 2026-09-14
feat(resourcepolicy): GORM model → policy adapter — RESGEN-1 slice 2 (#352)

leo-aa88:
## Summary

**RESGEN-1 slice 2 of #352** ([ADR-016](../blob/main/docs/adr/016-model-first-resource-generation.md)) — the GORM model → field-policy adapter. It promotes the schema→`Field` mapping that the slice-1 review vetted (`factsOf`/`canonicalFields`, previously test-only) into the production `resourcepolicy` API, so slice 3 can turn a real model into resolved API policy. Still no generation and nothing wired into `make resource`.

New in `resourcepolicy` (all backed by `gorm/schema.Parse`, no hand-rolled tag parsing):

- **`FactsFromSchema(*schema.Field) FieldFacts`** — the lossless projection: soft-delete by the `gorm.DeletedAt` **type** via `IndirectFieldType` (value or pointer), create/read capability from GORM's own `Creatable`/`Readable`, keys/timestamps/nullability/default from the parsed field. Nothing is re-derived by hand.
- **`FromSchema(*schema.Schema) []Field`** — enumerates GORM's **effective persisted columns** (`sch.DBNames` → `sch.FieldsByDBName`), so a duplicate-column or shadowed field is included **once**, each carrying its own `gombit` tag. Relationship/association fields (no column) are excluded; a belongs-to FK column is included.
- **`FromModel(model) ([]Field, error)`** and **`ResolvedFromModel(model) ([]Resolved, error)`** — parse a model (`&Book{}`) and optionally resolve, failing closed on a contradictory tag or an unsatisfiable required column.

## Why this shape

Per the slice-1 review thread: the only faithful source of persistence facts is GORM's own `schema.Parse` — so the adapter reads `Creatable`/`Readable`, the `gorm.DeletedAt` type, and the effective `DBNames`/`FieldsByDBName` set rather than re-interpreting struct tags. It carries each field's `gombit` tag through to resolution so policy can't vanish between selection and resolution.

## Validation

- [x] `go test ./resourcepolicy/...` — `policy_test.go` is now pure resolver coverage; `schema_test.go` exercises the adapter against real `gorm/schema.Parse` (+ SQLite): end-to-end `ResolvedFromModel` over a `gorm.Model` resource, contradiction rejection, soft-delete by type (value + pointer), duplicate-column collapse, tag survival, `->`/`<-` permissions, and resolver ⟺ runtime unsatisfiable cases.
- [x] `go vet` / `golangci-lint run ./resourcepolicy/...` — 0 issues.

## Scope notes

- Adapter only. Slice 3 uses `ResolvedFromModel` to generate `*.gen.go` DTOs/mappers; slice 4 the CRUD handler + hooks; slice 5 `gombit generate --check`; slice 6 removes `resourcecheck` + the breaking-change migration guide. Tracked in #352.
- Where the *generator process* gets the model instance to parse (a loader/Program-Mode step, like `atlas-provider-gorm`) is a slice-3+ concern; this adapter takes the model/schema it's given.

- [x] New behavior has tests — adapter unit + schema-derived + SQLite round-trip
- [x] Stable features ship docs — package/function doc comments specify the mapping; user-facing CLI docs land when the tag is wired (slice 3+)
- [x] Extraction preserves contracts — additive to `resourcepolicy`; resolver unchanged
- [x] Scope inside the RESGEN-1 epic (#352, post-v0.1); no M6 battery
- [x] Links its issue (#352) and states which slice it is


resourcepolicy/schema.go
```diff
@@ -0,0 +1,70 @@
...
+
+// FromSchema maps a parsed GORM schema to resolver input over its EFFECTIVE
+// persisted columns — ordered sch.DBNames → sch.FieldsByDBName — so a
+// duplicate-column or shadowed field is included once, each carrying its own
+// `gombit` tag. Relationship/association fields have no column and so are not
+// persisted columns and not included (a belongs-to FK column is). Feed the result
+// to ResolveAll (or use ResolvedFromModel).
+func FromSchema(sch *schema.Schema) []Field {
+	out := make([]Field, 0, len(sch.DBNames))
+	for _, name := range sch.DBNames {
```
leo-aa88, line 122: **BLOCKING.** This persisted-column-only iteration discards explicit policy on associations before the resolver can validate it. `Resolve` promises to reject a tagged relationship, and `ResolvedFromModel` says contradictory policy fails closed, but GORM stores associations outside `DBNames`. For example, a normal belongs-to field declared as `Author Author` with `gombit:"read"` is omitted here; `ResolvedFromModel(&Book{})` returns no error and only emits `id`/`author_id`. The authoritative policy has become a silent no-op. ಠ_ಠ Enumerate and validate non-column/relationship fields as well (while still emitting only effective persisted columns), and add end-to-end cases proving tagged belongs-to/has-one/has-many associations fail closed.

leo-aa88 (commented):
# REQUEST CHANGES

The persisted-column projection is mostly faithful, but it violates its own fail-closed contract by filtering associations out before policy validation. This is not mergeable while authoritative policy can disappear silently.

## 1. Relationship policy is silently discarded

**BLOCKING.**

`resourcepolicy/schema.go:41` iterates only `sch.DBNames`. GORM association fields do not appear there, so their `gombit` tags never reach `Resolve`. That contradicts both `Resolve`, which explicitly rejects policy on relationships, and `ResolvedFromModel`, which claims contradictory policy fails closed.

A normal belongs-to model with `Author Author` tagged `gombit:"read"` returns no error from `ResolvedFromModel`; the adapter emits only `id` and `author_id`. The explicit policy is ignored. Selection and validation need separate ownership: validate explicit policy on non-column/relationship fields, then emit only the canonical persisted-column set. Add end-to-end cases for tagged belongs-to, has-one, and has-many associations.

CI is green apart from the still-running benchmark smoke job. This is not a failing-test problem. The current tests do not exercise the advertised relationship-validation contract.

# VERDICT

The defect is in the adapter abstraction: it conflates “not emitted as a DTO column” with “does not need validation.” That turns malformed authoritative metadata into a silent no-op—the exact class of drift this redesign is supposed to eliminate.

Do not merge until relationship policy survives far enough through model reading to be rejected explicitly.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// Selection and validation are separate: association fields live outside DBNames,
+// so a `gombit` policy on one would never reach Resolve. FromSchema therefore
+// VALIDATES every relationship field first (a relationship cannot carry API
+// policy — the same rule Resolve enforces) and fails closed on a tagged one,
+// rather than silently discarding the policy, before emitting the persisted set.
+func FromSchema(sch *schema.Schema) ([]Field, error) {
+	// Validate relationship fields (deterministic order for stable errors).
+	rels := make([]*schema.Field, 0, len(sch.Relationships.Relations))
+	for _, rel := range sch.Relationships.Relations {
```
leo-aa88: **BLOCKING.** This still validates through a lossy name-keyed index instead of the fields that carry policy. GORM collapses same-named embedded associations in `Relationships.Relations`; the full paths live under `EmbeddedRelations`. I reproduced `Left.Author` tagged `gombit:"read"` plus untagged `Right.Author`: the top-level map retains only `Right.Author`, and `FromSchema` returns success. A `gorm:"-" gombit:"write"` scalar also vanishes because it is in neither `DBNames` nor `Relationships.Relations`. So the fix closes the four happy-path relationship shapes but not the invariant: every explicit policy must be validated exactly once. Traverse `sch.Fields` by field identity/bind path and validate every policy-bearing field with no effective column (or equivalently traverse all embedded relations plus ignored fields), then use `DBNames` only for emission. Add the embedded same-name collision and ignored-field cases; slice 1 already established that Go names are not unique.

leo-aa88 (commented):
# REQUEST CHANGES

The previous association-policy blocker is fixed only for top-level, uniquely named relationships. The implementation still validates through GORM's lossy relationship-name map, so explicit policy can disappear in valid embedded models and on other non-column fields.

## 1. Validation is still keyed by a non-unique relationship name

**BLOCKING.**

`resourcepolicy/schema.go:49` walks only `sch.Relationships.Relations`. GORM keys that map by relationship name. Two embedded structs can both flatten an `Author` association; GORM retains the full paths in `EmbeddedRelations` but the top-level `Relations["Author"]` entry represents only one field.

I reproduced a model containing tagged `Left.Author` and untagged `Right.Author`. The top-level map retained only `Right.Author`, and `FromSchema` returned success. A scalar tagged `gorm:"-" gombit:"write"` is also silently ignored because it appears in neither `DBNames` nor `Relationships.Relations`.

The contract is not “validate the relationship entries that survived a map collision.” It is “validate every explicit field policy.” Traverse `sch.Fields` by field identity/bind path and validate every policy-bearing field without an effective column, then use `DBNames` solely for canonical persisted-column emission. Add tests for the embedded same-name collision and a tagged GORM-ignored field.

CI is green. This is not a failing-test problem. The current tests cover four top-level association kinds but do not attack the name-collision invariant already documented by slice 1.

# VERDICT

This remains an abstraction defect in model traversal. Selection and validation are nominally separated, but validation still consumes a lossy secondary index instead of the authoritative fields carrying the tags. That recreates the exact silent-policy-loss failure this adapter claims to eliminate.

Do not merge until every explicit policy reaches validation regardless of relationship name collisions or GORM persistence exclusion.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,98 @@
...
+// Selection and validation are separate. An explicit `gombit` policy on a field
+// that is not an effective persisted column (a relationship, a `gorm:"-"` ignored
+// field, or a column shadowed by another field) would otherwise be silently
+// dropped — it lives outside the emitted set. FromSchema therefore validates
+// EVERY policy-bearing field from GORM's authoritative per-field set (sch.Fields),
+// keyed by field identity, not a name-keyed secondary index that collapses
+// same-named embedded fields. A policy that cannot be honored fails closed. Then
+// it emits the effective persisted columns.
+func FromSchema(sch *schema.Schema) ([]Field, error) {
+	for _, f := range sch.Fields {
```
leo-aa88, line 103: **BLOCKING.** `sch.Fields` still is not “every policy-bearing field” from the model. GORM replaces an embedded container with its flattened child fields, and it does not propagate non-`gorm` struct tags to those children. I reproduced an embedded `Audit` field carrying both `gorm:"embedded"` and `gombit:"read"`, where `Audit` contains `Note string`: `sch.Fields` contains only `ID` and `Audit.Note`, neither carries the policy tag, and `FromSchema` returns success with `note` exposed by defaults. An unsupported policy location has silently changed the API again. The previous collision/ignored-field reproducers are fixed, but the invariant remains incomplete. Validate legal `gombit` tag locations against the model structure before GORM discards embedded-container metadata (or retain that metadata in the model-reading representation), and add anonymous plus named embedded-container cases. If container tags are unsupported—and current semantics provide no way to honor them—they must fail closed, not vanish.

leo-aa88 (commented):
# REQUEST CHANGES

The `sch.Fields` traversal fixes the previous relationship-name collision, ignored-field, and shadowed-column failures. It still does not validate every policy-bearing source field as claimed: GORM erases embedded-container metadata before this traversal runs.

## 1. Embedded-container policy still disappears before validation

**BLOCKING.**

`resourcepolicy/schema.go:49` treats `sch.Fields` as GORM's authoritative per-field set. For an embedded struct, however, GORM replaces the container field with its flattened children and does not propagate the container's non-`gorm` struct tags.

I reproduced an embedded `Audit` field carrying `gorm:"embedded" gombit:"read"`, with `Audit` containing `Note string`. The parsed schema contains only `ID` and `Audit.Note`; neither has the `gombit` tag. `FromSchema` returns success and exposes `note` using defaults. The explicit policy is still a silent no-op.

Current semantics cannot honor policy on an embedded container, so it must be rejected before schema flattening discards it. Validate legal `gombit` tag locations against the model structure—or preserve container metadata in the model-reading representation—and add anonymous and named embedded-container tests.

Focused package tests, race tests, vet, diff hygiene, and the primary CI test pass. This is not a failing-test problem. The tests do not exercise policy attached to the embedding field itself.

# VERDICT

The defect remains in the model-reading abstraction: `schema.Schema` is authoritative for persistence facts, but `sch.Fields` is not a lossless inventory of source-level policy locations. Treating it as one lets unsupported policy mutate the generated contract silently.

Do not merge until embedded-container policy fails closed rather than vanishing.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,159 @@
...
+// column (time.Time, gorm.DeletedAt, a Valuer/Scanner type) is neither.
+func isEmbeddedContainer(f reflect.StructField) bool {
+	ft := f.Type
+	for ft.Kind() == reflect.Pointer {
+		ft = ft.Elem()
+	}
+	if ft.Kind() != reflect.Struct {
+		return false
+	}
+	if f.Anonymous {
```
leo-aa88: **BLOCKING.** This contradicts the comment immediately above it. GORM does **not** flatten every anonymous struct: scalar struct types such as `time.Time`, `gorm.DeletedAt`, and Scanner/Valuer implementations remain ordinary persisted fields. I confirmed a model with an anonymous `time.Time` carrying `gombit:"read"` parses as an effective field named `Time` with DB column `time` and the policy tag intact. This branch nevertheless classifies it as an embedded container and rejects a valid field policy before `Resolve` sees it. The guard must identify whether GORM actually flattened the field, not approximate that as `Anonymous && struct`; correlate the source bind path with `sch.Fields`/the effective field set (including nested paths), or otherwise use GORM’s parsed classification rather than duplicating only half of its Valuer/data-type rules. Add anonymous scalar cases for `time.Time` and a custom Scanner/Valuer type.

leo-aa88 (commented):
# REQUEST CHANGES

The embedded-container guard fixes the previous silent-loss case, but it now rejects valid scalar fields because it approximates GORM's embedding rules as “every anonymous struct is flattened.” GORM does not behave that way.

## 1. Anonymous scalar struct fields are misclassified as containers

**BLOCKING.**

`resourcepolicy/schema.go:128` returns true for every anonymous struct. GORM keeps scalar struct types— including `time.Time`, `gorm.DeletedAt`, and Scanner/Valuer implementations—as persisted columns instead of flattening them.

I confirmed a model with an anonymous `time.Time` carrying `gombit:"read"` parses to an effective `Time` field with DB column `time` and the policy intact. `FromModel` then rejects it as an embedded container. The comment says scalar struct columns are not containers; runtime does the opposite.

Determine whether GORM actually flattened the source field by correlating its bind path with `sch.Fields`/the effective field set, rather than duplicating an incomplete subset of GORM's type classification. Add anonymous `time.Time` and custom Scanner/Valuer cases.

Focused package tests, race tests, vet, and diff hygiene pass. This is not a failing-test problem. The tests cover actual containers but not the opposite branch: anonymous structs GORM persists as scalars.

# VERDICT

The defect is a semantic mismatch between the new reflection classifier and GORM's runtime schema. This fix trades a false negative for a false positive because it duplicates GORM's embedding decision without reproducing its scalar-type rules.

Do not merge until container validation follows GORM's parsed classification exactly.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,160 @@
...
+// are recognized); create/read capability comes from GORM's own Creatable/
+// Readable (the `->`/`<-` permission tags); nothing is re-derived by hand.
+func FactsFromSchema(f *schema.Field) FieldFacts {
+	return FieldFacts{
+		GoName:        f.Name,
+		Column:        f.DBName,
+		PrimaryKey:    f.PrimaryKey,
+		AutoIncrement: f.AutoIncrement,
+		AutoTime:      f.AutoCreateTime != 0 || f.AutoUpdateTime != 0,
+		SoftDelete:    f.IndirectFieldType == reflect.TypeOf(gorm.DeletedAt{}),
```
leo-aa88: **BLOCKING — soft-delete classification disagrees with GORM runtime behavior.** GORM does not dispatch soft deletion by exact type equality; as the comment itself notes, it calls the clause interfaces on `reflect.New(IndirectFieldType)`. A wrapper such as `type WrappedDeleted struct { gorm.DeletedAt }` inherits those methods. I verified that GORM parses `RemovedAt WrappedDeleted` with `SoftDeleteQueryClause`/`SoftDeleteDeleteClause`, SQLite `Delete` performs a soft delete, and the default scoped query hides the row. This comparison still sets `SoftDelete=false`, so `ResolvedFromModel` exposes `removed_at` in both request and response as ordinary content. The signature describes one lifecycle while GORM executes another. Derive this fact from the parsed field’s actual clause behavior (or fail closed for unrecognized lifecycle-clause types), rather than comparing one concrete type. Add a runtime test using a wrapper that embeds `gorm.DeletedAt`; merely adding more exact types will recreate this bug for the next implementation.

leo-aa88 (commented):
# REQUEST CHANGES

The `BindNames` fix resolves the previous anonymous-scalar false positive. The adapter still does not provide the lossless GORM lifecycle projection it claims: soft-delete behavior is classified by exact concrete type while GORM dispatches it behaviorally.

## 1. Exact-type soft-delete detection disagrees with GORM

**BLOCKING.**

`FactsFromSchema` sets `SoftDelete` only when `IndirectFieldType` is exactly `gorm.DeletedAt`. GORM invokes field clause interfaces on a new value of that type, so a wrapper such as:

```go
type WrappedDeleted struct { gorm.DeletedAt }
type Row struct {
    ID        uint
    RemovedAt WrappedDeleted
}
```

inherits the soft-delete methods and is a real soft-delete column at runtime. I verified the parsed schema contains GORM's soft-delete query/delete/update clauses; against SQLite, `Delete` updates the marker and the default scope hides the row.

The adapter nevertheless returns `SoftDelete=false`, and untagged policy resolves `removed_at` as request-writable and response-readable ordinary content. The code's lifecycle model and GORM's runtime lifecycle are different. Derive this fact from actual parsed clause behavior, or reject lifecycle-clause types the adapter cannot classify. Do not grow another concrete-type allowlist.

CI is green. This is not a failing-test problem. The tests cover only exact `gorm.DeletedAt` value/pointer types and therefore encode the implementation's assumption instead of GORM's dispatch contract.

# VERDICT

This is a semantic/runtime contract defect in the schema adapter. A field that GORM owns as the deletion marker is handed to future DTO/mapping generation as ordinary user-controlled data—the precise kind of representation drift this package exists to prevent.

Do not merge until soft-delete classification follows GORM's actual field behavior and the wrapper case is covered end to end.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,175 @@
...
+// is the same behavioral dispatch GORM uses — matching gorm.DeletedAt and any type
+// embedding it — rather than an exact concrete-type comparison.
+func isSoftDeleteField(f *schema.Field) bool {
+	if f.IndirectFieldType == nil {
+		return false
+	}
+	v := reflect.New(f.IndirectFieldType).Interface()
+	_, query := v.(schema.QueryClausesInterface)
+	_, del := v.(schema.DeleteClausesInterface)
+	return query && del
```
leo-aa88: **BLOCKING — these are generic clause-provider interfaces, not soft-delete marker interfaces.** Implementing both says only that a field contributes clauses to query and delete callbacks; it does not say those clauses scope deleted rows or rewrite `DELETE` into `UPDATE`. I reproduced a `type Clauseful string` whose two methods each return an ordinary `clause.Where{1 = 1}`. GORM persists it as normal content and hard-deletes the row (`Unscoped` count is 0), while this predicate sets `SoftDelete=true` and `ResolvedFromModel` hides the column from both request and response. The previous exact-type check had a false negative; this replacement introduces a false positive for every non-soft-delete clause provider. Inspect the actual returned clause semantics (for example GORM’s concrete soft-delete query/delete clauses), or establish an explicit soft-delete classifier; do not assign lifecycle meaning from generic interface membership. Add the opposite-direction runtime test, because the current wrapper test proves only that all soft-delete fields match, not that all matches are soft-delete.

leo-aa88 (commented):
# REQUEST CHANGES

The wrapper regression is fixed, but the replacement predicate still does not match GORM's runtime semantics. It equates two generic extension interfaces with soft deletion and therefore hides unrelated persisted fields.

## 1. Generic clause providers are misclassified as soft-delete markers

**BLOCKING.**

`isSoftDeleteField` returns true for every type implementing both `schema.QueryClausesInterface` and `schema.DeleteClausesInterface`. Those interfaces describe callback clause contribution; neither promises soft-delete semantics.

I reproduced a scalar string type whose two methods return ordinary `clause.Where{1 = 1}` clauses. GORM persists the field as normal content and performs a hard delete—the row is absent even under `Unscoped`. The adapter nevertheless sets `SoftDelete=true`, so `ResolvedFromModel` removes the field from both request and response.

The previous implementation missed wrapper soft-delete types. This one silently hides every non-soft-delete type that happens to participate in both callbacks. Inspect the actual returned clause semantics or introduce an explicit classifier. Generic interface membership is not a lifecycle contract.

CI is green for completed jobs, and focused package tests, race tests, vet, and diff hygiene pass locally. This is not a failing-test problem. The current test proves the positive wrapper case but never attacks the opposite direction: whether every match is genuinely soft-delete.

# VERDICT

This remains a semantic/runtime abstraction defect. The adapter labels a hard-deleted, ordinary persisted column as framework-managed deletion state, so the model-derived API contract is wrong before generation even begins.

Do not merge until soft-delete detection distinguishes actual soft-delete clauses from arbitrary GORM clause providers and includes the negative runtime case.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,185 @@
...
+func isSoftDeleteField(f *schema.Field) bool {
+	if f.IndirectFieldType == nil {
+		return false
+	}
+	dc, ok := reflect.New(f.IndirectFieldType).Interface().(schema.DeleteClausesInterface)
+	if !ok {
+		return false
+	}
+	for _, c := range dc.DeleteClauses(f) {
+		if _, ok := c.(gorm.SoftDeleteDeleteClause); ok {
```
leo-aa88: **BLOCKING — this still matches one Go representation, not the parsed soft-delete behavior.** `gorm.SoftDeleteDeleteClause` has value-receiver methods, so `*gorm.SoftDeleteDeleteClause` is also a valid `clause.Interface` and performs the identical DELETE→UPDATE rewrite. I reproduced a wrapper whose `DeleteClauses` returns `&gorm.SoftDeleteDeleteClause{Field: f}`: GORM soft-deletes it (`visible=0`, `Unscoped=1`), while this value-only assertion returns false and the adapter exposes the marker as ordinary content again. Worse, this helper re-executes arbitrary user `DeleteClauses` code even though `schema.Parse` already executed it and stored the authoritative clauses; a stateful provider can make the second call disagree with the schema runtime will use. Read the clauses already stored on the parsed schema, associate the soft-delete clause with this field, and handle both value and pointer forms. Do not call the extension hook twice.

leo-aa88 (commented):
# REQUEST CHANGES

The generic-clause false positive is fixed. The replacement still does not consume GORM's parsed result: it re-runs user clause code and recognizes only the value representation of an exported clause whose pointer has identical runtime semantics.

## 1. Soft-delete detection still disagrees with the parsed schema

**BLOCKING.**

`isSoftDeleteField` invokes `DeleteClauses` again and type-asserts only `gorm.SoftDeleteDeleteClause`. Because that clause's methods have value receivers, `*gorm.SoftDeleteDeleteClause` is equally valid and performs the same DELETE-to-UPDATE rewrite.

I reproduced a wrapper returning `&gorm.SoftDeleteDeleteClause{Field: f}`. GORM soft-deletes the row: scoped count is zero and unscoped count is one. The adapter returns `SoftDelete=false` and exposes the lifecycle marker as ordinary content.

Re-executing `DeleteClauses` is also the wrong ownership boundary. `schema.Parse` already called this arbitrary extension hook and stored the clauses runtime will use; a stateful provider can return a different answer on this second call. Read the parsed schema's stored clauses, match the clause to the field, and support both pointer and value representations.

All CI checks are green. Focused package tests and diff hygiene pass; the pointer-clause runtime case is absent. This is not a failing-test problem. The current tests cover only the exact representation emitted by built-in `gorm.DeletedAt`.

# VERDICT

This remains a semantic/runtime adapter defect. The code claims to project GORM's authoritative parsed behavior but instead replays user code and recognizes one representation, allowing a real soft-delete marker to become user-controlled API data.

Do not merge until classification reads the parsed clauses GORM will actually execute and handles the clause's valid pointer form.

resourcepolicy/schema.go
```diff
@@ -0,0 +1,200 @@
...
+
+// FactsFromSchema is the lossless projection of a parsed GORM field into
+// FieldFacts. Soft-delete is read from the clauses schema.Parse already stored on
+// the schema (see softDeleteFields) — GORM's authoritative parsed result — so
+// gorm.DeletedAt, *gorm.DeletedAt, and any wrapper whose DeleteClauses rewrite
+// DELETE into an UPDATE are recognized, in either the value or pointer clause
+// form, while a generic clause provider that hard-deletes is not. Create/read
+// capability comes from GORM's own Creatable/Readable (the `->`/`<-` permission
+// tags); nothing is re-derived by hand.
+func FactsFromSchema(sch *schema.Schema, f *schema.Field) FieldFacts {
```
leo-aa88: **MAJOR — the fix creates an unenforced two-object invariant in the public API.** `FactsFromSchema` now accepts `(sch, f)`, but nothing proves that `f` belongs to `sch`. Pass a soft-delete field parsed from schema A with schema B and this silently returns `SoftDelete=false`, exposing the marker as ordinary content. The function still claims to be a lossless projection of *the field*, and GORM already stores the owning parsed schema on `f.Schema`; the PR description also still advertises the original one-argument API. Keep `FactsFromSchema(f)` and derive the clause set from `f.Schema` (while `FromSchema` can retain its one-scan private fast path), or reject mismatched inputs explicitly. Do not make callers maintain a relationship the type already represents authoritatively.

leo-aa88 (commented):
# REQUEST CHANGES

The previous soft-delete blocker is fixed at `6b093454`: the adapter now consumes GORM’s stored clauses, handles value and pointer clause representations, and no longer replays arbitrary extension hooks. The replacement public API still encodes the schema/field ownership relationship incorrectly.

## 1. `FactsFromSchema` accepts a schema/field pair that may disagree

**MAJOR.**

`FactsFromSchema(sch, f)` computes ordinary field facts from `f` but soft-delete state from `sch`. There is no check that `f` belongs to `sch`. Pair a real soft-delete field from schema A with schema B and the function silently reports `SoftDelete=false`, exactly the kind of contract drift this package exists to prevent.

GORM already stores the owning parsed schema as `f.Schema`. Preserve the one-argument `FactsFromSchema(f)` API advertised in the PR description and derive its clauses from that authoritative owner; keep the precomputed private `factsFromSchema` path inside `FromSchema` to avoid repeated scans. If the two-argument shape is retained, it must at least reject a mismatch instead of returning plausible-but-wrong policy facts.

`go test ./resourcepolicy/...` and `go test -race ./resourcepolicy/...` pass on this HEAD. CI’s completed checks are green. This is not a failing-test problem; the tests always pass matching pairs and therefore cannot expose the invalid public state.

# VERDICT

The runtime/semantic mismatch from the prior review is gone. The remaining defect is the abstraction boundary: a public projection of one parsed field should not require callers to supply a second, redundant owner and silently trust that relationship.

Do not merge this API shape as-is.

leo-aa88 (commented):
LGTM

PR #356, opened 2026-09-14
perf(framework): merge request_timeout into request_context (#268)

### LINKED ISSUE #268: [PERF-10] request_context is 11 allocs/op: drop gin.Context.Set, single ID string, direct header slices, merge timeout wrap

leo-aa88:
## Summary

`request_context` was the single largest allocator in the runtime middleware stack (11 allocs/op), and `request_timeout` then performed a second `Request.WithContext` on the same request. This PR collapses both into one middleware and one context wrap, per the #268 plan:

- **Drop both `c.Set` calls** (the Gin `Keys` map + two boxed strings). `GetRequestID`/`GetTraceID` now read the same context value the `*FromContext` accessors do — public API unchanged. (−3)
- **One `string()` for both generated IDs**: they are hex-encoded into a single `[68]byte` stack buffer and two sub-slices of the resulting string become the IDs. (−1)
- **Direct response-header slices**: one per-request `*requestMeta` holds both IDs and a `[2]string` backing array; `X-Request-Id`/`X-Trace-Id` alias `hdr[0:1:1]` / `hdr[1:2:2]` instead of `Header.Set`. Cap 1 keeps `Header.Add` safe and stops a write-through into the sibling header. (−2)
- **Merge the timeout wrap in**: `context.WithValue` → `applyTimeout` (keeps the #242 skip-when-tighter-deadline guard) → a single `Request.WithContext`. The standalone `request_timeout` layer is removed from the stack. (−1)
- Early-return in `traceIDFromTraceparent` on the empty header (CPU only).

Ablation ladder (production config, timeout enabled), allocs/op valid POST / plaintext:

| row | valid POST | plaintext |
| --- | ---: | ---: |
| baseline (Huma+Gin) | 34 | 14 |
| + request_context (**incl. timeout**) | 42 | 22 |
| + metrics | 42 | 22 |
| + security_headers | 42 | 22 |
| + xss | 46 | 22 |
| full `framework.App` | 46 | 22 |

`request_context` is now **+8** on both scenarios, down from +16 for the old `request_context` (+11) and `request_timeout` (+5) layers combined. The full app drops from 60/35 to **46/22** allocs/op.

Closes #268

## Acceptance criteria

- [x] Ablation: request_context (now including the timeout wrap) ≤ 8 allocs/op with the timeout enabled (measured +8/+8). ≤ 4 with it disabled: `applyTimeout`'s disabled path adds no context/timer allocation; locked by `TestApplyTimeoutDisabledIsNoop` / `TestRequestContextMiddlewareDisabledTimeoutImposesNoDeadline`.
- [x] `GetRequestID`, `GetTraceID`, `GetRequestIDFromContext`, `GetTraceIDFromContext` behave identically; `correlation_id_test.go` passes unchanged; the timeout tests were retargeted (see scope note) with every assertion preserved.
- [x] Inbound `X-Request-Id` and W3C `traceparent` still honored; `X-Request-Id` / `X-Trace-Id` response headers unchanged (`TestDefaultRouterPreservesRequestID`, `TestDefaultRouterCarriesTraceContext`).
- [x] Test that `Header.Add(RequestIDHeader, ...)` after the middleware does not write through the backing array — `TestRequestCorrelationHeaderSharedArrayContract`, mirroring `TestSecurityHeaderSharedValueContract`.
- [x] Middleware name list updated for the merged layer (`TestDefaultRuntimeMiddlewareOrder` no longer lists `request_timeout`).

## Scope notes

- The per-handler timeout logic moved from the standalone `requestTimeoutMiddleware` into the shared `applyTimeout` helper (the #242 guard's owner) plus the merged `requestContextMiddleware`. `request_timeout_test.go` was therefore retargeted to `applyTimeout` and the merged middleware rather than left literally unchanged — the merge the issue mandates makes the old "same context identity through the middleware" assertion impossible (request_context always wraps for the context value), so the guard is now tested at the unit that owns it. Every original assertion is preserved.
- The **decision** on whether the per-handler timeout should be on by default is #270 (PERF-12); this PR keeps today's default (`RequestTimeout = 60s`).

## Validation

- [x] `go test ./framework ./config ./contract` — pass
- [x] `go test ./framework -run '^$' -bench '^BenchmarkAblation$' -benchmem` — table above
- [ ] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run` — running

- [x] New behavior has tests; no DB-touching change (middleware only)
- [ ] Stable features ship docs and appear in an example app — N/A: no public API or behavior change; the example router (`examples/router`) already exercises `GetRequestID`/`GetTraceID` and is unaffected
- [x] Extraction preserves contracts — refactor and move, not rewrite
- [ ] Generators — N/A, no generator touched
- [ ] API changes regenerate OpenAPI + TS client — N/A, no API change
- [x] Scope stays inside milestone — post-v0.1 runtime-only, no M6 battery


leo-aa88 (commented):
LGTM.

The `[2]string` aliasing is the part that had to be right, and it is. `hdr[0:1:1]` / `hdr[1:2:2]` are three-index slices with cap==1, so `http.Header.Add` is forced to reallocate rather than write through into the sibling header's backing slot — and `TestRequestCorrelationHeaderSharedArrayContract` exercises exactly that write-through it prevents, not just the happy path. The header keys are already canonical MIME form (`X-Request-Id` / `X-Trace-Id`), so bypassing `c.Header()`'s canonicalization with a direct `c.Writer.Header()[...]` map write is safe.

The #242 deadline-skip guard survives the move into `applyTimeout` intact: `context.WithValue` adds no deadline, so `ctx.Deadline()` there is equivalent to the old `c.Request.Context().Deadline()` check, and `defer cancel()` still fires after `c.Next()`. `GetRequestID` / `GetTraceID` delegating to the `*FromContext` accessors drops the two `c.Set` boxes without changing the public contract — I checked the callers (`handleReadyz`, `contract.Install`, `examples/router`) and they all read through the exported getters, not the removed unexported gin keys. The one-alloc combined-ID encoding (single `[68]byte` -> one `string` -> two substrings) is sound.

Ship it.

PR #357, opened 2026-09-14
feat(config)!: make the per-handler request timeout opt-in (#270)

### LINKED ISSUE #270: [PERF-12] Decide: make the per-handler request timeout opt-in (4–5 allocs/op on every request by default)

leo-aa88:
## Summary

`request_timeout` ran on **every** request with a default `HTTP.RequestTimeout = 60s`. `context.WithTimeout` is ~4–5 allocs/op on the common path (no inbound deadline). This makes the per-handler deadline **opt-in** — option A of #270, recorded in **ADR-017**.

- **Default `HTTP.RequestTimeout = 0`** (disabled).
- **Omit the layer, don't pass through**: `runtimeMiddlewareStack` appends `request_timeout` only when `RequestTimeout > 0`. A disabled deadline costs nothing on the request path and produces no ablation row.
- **Decouple the connection-level safety net**: `HTTP.RequestTimeout` used to drive *both* the per-handler context deadline *and* the `http.Server` `Read`/`Write`/`IdleTimeout`. Naively defaulting it to 0 would have zeroed the server timeouts too — a DoS regression. They now fall back to a fixed `defaultHTTPServerTimeout` (60s) when `RequestTimeout <= 0`; a positive value still drives all three, unchanged. `ReadHeaderTimeout` stays 5s.
- **Scaffolded apps keep today's behavior**: `gombit new` writes `GOMBIT_HTTP_REQUEST_TIMEOUT=60s` into `.env.example` (golden trees regenerated).

Closes #270

## Acceptance criteria

- [x] Decision recorded (ADR-017) with the measured cost and the trade-off; docs updated (`docs/config.md`, `docs/router.md`).
- [x] Default changed to 0; `config` tests + docs updated; scaffold template sets an explicit `60s` value; the layer is absent from the stack when 0 (`TestRuntimeMiddlewareStackTimeoutOptIn`, `TestDefaultRuntimeMiddlewareOrder`); the ablation shows no `request_timeout` row (production default is now 0).
- [x] Timeout behavior when enabled is unchanged: `TestDefaultRouterRequestTimeoutSetsDeadline` and `TestRequestTimeoutMiddlewareImposesDeadlineWhenNonePresent` pass; `TestRunContextConfiguresServerTimeouts` (RequestTimeout > 0) unchanged.

## Scope notes

- **Server-timeout decoupling** (the fallback constant) is required for option A's stated invariant — "the `http.Server` timeouts remain the safety net" — to hold once `RequestTimeout` defaults to 0. `GOMBIT_HTTP_REQUEST_TIMEOUT=0` no longer zeroes the server timeouts (that is now unreachable via this var); a dedicated `HTTP.ServerTimeout` knob is deferred (ADR-017 "Alternatives considered").
- **Breaking**: apps relying on the implicit 60s per-handler deadline must set `GOMBIT_HTTP_REQUEST_TIMEOUT`. CHANGELOG marks it under Changed with a `BREAKING CHANGE`.
- Interacts with #268 (which merges the timeout wrap into `request_context`); the two conflict textually in `app.go`/`http_middleware.go`. Post-#268, "omit the layer when 0" becomes "`applyTimeout` no-ops when 0", which #268 already does — the merge resolution is mechanical.

## Validation

- [x] `go test ./...` — pass (goldentest regenerated via `-update`)
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run ./framework/... ./config/...` — 0 issues

- [x] New behavior has tests (opt-in stack test + server-timeout fallback test); no DB-touching change
- [x] Stable features ship docs and appear in an example app — docs updated; the scaffold (`gombit new`) sets the value and its golden trees are updated
- [x] Extraction preserves contracts — behavior for `RequestTimeout > 0` is unchanged
- [ ] Generators — N/A (template value + golden snapshots updated, no generator logic changed)
- [ ] API changes regenerate OpenAPI + TS client — N/A, no HTTP API change
- [x] Scope stays inside milestone — post-v0.1 runtime/config, no M6 battery


framework/app.go
```diff
@@ -452,12 +461,21 @@ func RunContext(ctx context.Context, app *App) error {
 		return fmt.Errorf("framework: listen: %w", err)
 	}
 
+	// The per-handler context deadline (HTTP.RequestTimeout) is opt-in and off by
+	// default (issue #270), but the connection-level timeouts are a safety net
+	// that must stay on. When RequestTimeout is set it still drives all three;
+	// when it is disabled they fall back to defaultHTTPServerTimeout rather than 0
+	// (unbounded).
+	serverTimeout := app.Config().HTTP.RequestTimeout
+	if serverTimeout <= 0 {
```
leo-aa88, line 470: **MAJOR / QUESTION — the narrowing removes the only escape hatch for an unbounded `WriteTimeout`.**

Before this PR, `GOMBIT_HTTP_REQUEST_TIMEOUT=0` zeroed `ReadTimeout`/`WriteTimeout`/`IdleTimeout` — that was the one way to run the server with no write deadline. After it, `0` is the default *and* clamps all three to a hard-coded 60s, and there is no longer **any** value of this variable that yields an unbounded `WriteTimeout`.

For the framework itself this is harmless — I checked, there's no first-party SSE / websocket / chunked-stream / long-upload surface (the only `Flush()` calls in the tree are CLI `tabwriter`). But a *user app* that adds an SSE endpoint, a WS upgrade, or a multi-minute upload/download previously had exactly one lever — set this var to 0 — and this PR silently takes it away while the documented replacement (`HTTP.ServerTimeout`) is deferred to "later." Post-merge such an app's long-lived responses are severed at 60s with no config knob to fix it short of not using `RunContext`'s server.

That lands the regression squarely on the apps least able to absorb it, gated behind a feature that isn't in this PR. The `ServerTimeout` field is the same ~4-line pattern as `RequestTimeout` right here — is there a reason it can't ship in the PR that creates the need for it? If it genuinely must wait, the CHANGELOG and ADR should say outright that streaming/upload apps have no way to lift `WriteTimeout` until it lands, rather than filing it under "Alternatives considered."

leo-aa88 (commented):
# COMMENT

The core change is right and the part that could have caused a DoS regression — naively letting `RequestTimeout=0` zero the `http.Server` read/write/idle timeouts — is correctly headed off with the `defaultHTTPServerTimeout` fallback, `ReadHeaderTimeout` kept at 5s, and a test that actually asserts the fallback value. `Validate` already rejects negatives, so the `serverTimeout <= 0` branch is reachable only via the intended `0`. Docs/ADR/golden/scaffold are all consistent.

One design consequence of the decoupling is under-weighted — the narrowing removes the *only* config escape hatch for an unbounded server write timeout, and the replacement is deferred. Detail inline. Not blocking (no first-party surface breaks, and it's disclosed in the ADR), but I'd push to ship the knob with the change that creates the need for it.

Note for whoever merges the PERF-10/12/13 trio: this PR keeps `requestTimeoutMiddleware` as a standalone conditional layer, while #268/PR 356 *deletes* that function and folds the deadline into `requestContextMiddleware`. If 356 lands first, this PR's `requestTimeoutMiddleware(...)` reference and `TestRuntimeMiddlewareStackTimeoutOptIn`'s search for a `request_timeout` layer no longer compile/pass — the resolution is a real rework (thread `RequestTimeout` into `requestContextMiddleware`, let `applyTimeout` no-op at 0), not the "mechanical" merge the description claims.

config/config.go
```diff
@@ -94,6 +94,13 @@ type Config struct {
 type HTTPConfig struct {
 	Addr           string
 	TrustedProxies []string
+	// RequestTimeout is the per-handler deadline propagated into the request
+	// context (and thus into DB/cache calls that honor it). It is opt-in: the
+	// default is 0, which disables it and omits the layer from the middleware
```
leo-aa88: **MAJOR — the #268 reconciliation is incomplete.** `requestContextMiddleware(cfg.HTTP.RequestTimeout)` is now installed unconditionally; `0` only makes `applyTimeout` a no-op. This public config contract still says the middleware layer is omitted. The same stale claim remains in `DefaultFor`, `docs/config.md`, and `CHANGELOG.md`, while the updated ADR and router docs correctly say there is no separate timeout layer anymore. Make every copy describe the actual invariant: the request-context layer always exists, and only deadline setup is skipped when the duration is zero.

leo-aa88 (commented):
# REQUEST CHANGES

The timeout behavior itself is coherent after rebasing onto #356: zero skips deadline creation while `request_context` remains the single always-installed middleware, and positive values still propagate a deadline. The public contract was only partially reconciled, however.

## 1. Public configuration documentation describes a layer that no longer exists

**MAJOR.**

`HTTPConfig.RequestTimeout`, `DefaultFor`, `docs/config.md`, and `CHANGELOG.md` all say that a zero timeout omits a timeout middleware layer. After #356, the runtime always installs `request_context`; zero only makes its internal `applyTimeout` step return without allocating a deadline. The ADR and router documentation were updated to this new architecture, but these four copies were not.

That is not harmless historical wording. An engineer reasoning from the exported field comment is being promised a middleware-stack change that the runtime does not perform. Remove the obsolete layer language everywhere and state the real invariant once: request context remains installed; per-handler deadline creation is opt-in.

CI is green for the completed checks. This is not a failing-test problem; tests do not validate the accuracy of these public descriptions.

# VERDICT

The implementation is acceptable, but the documented API contract contradicts it in multiple user-facing locations. This is a documentation/abstraction reconciliation defect introduced by the rebase, and the PR does not meet the repository's stable-feature documentation requirement until those claims are fixed.

Do not merge as-is.

leo-aa88 (commented):
LGTM

PR #358, opened 2026-09-14
feat(framework)!: move request-input sanitization out of the default pipeline (#271)

### LINKED ISSUE #271: [PERF-13] Decide: move request-input XSS sanitization out of the default pipeline (5 allocs/op per POST, rewrites API input)

leo-aa88:
## Summary

`xssMiddleware` ran on **every** request in the default pipeline: it buffered and re-encoded every JSON `POST`/`PUT`/`PATCH` body (+5 allocs/op) and rewrote API input. XSS is an **output-encoding** concern, and stripping markup on ingress corrupts faithful values — a comment of `x < y`, a stored `<b>bold</b>`, a code snippet — the class of bug behind #118/#201/#232. This moves sanitization out of the default and offers it as explicit, narrower opt-ins (decision recorded in **ADR-018**).

- **Default `framework.App` has no `xss` layer.** It does not read, decode, or rewrite request input. Nothing on the output side changes (React escapes text; admin templates escape untrusted data).
- **Per-app opt-in** `Security.SanitizeInput` (`GOMBIT_SECURITY_SANITIZE_INPUT`, default `false`) installs today's middleware unchanged — the `password`-key exemption, the 8 MiB cap, the #201 unclosed-element fix, and `WithRawBodyPaths` exemption all preserved.
- **Per-field opt-in** `framework.SanitizeHTML(s string) string` exports the existing `stripHTML` behavior for a handler to strip one untrusted field. (The issue wrote `gombit.SanitizeHTML`; the root `gombit` package only holds `ModulePath`, so the real symbol is `framework.SanitizeHTML`, the package generated apps import.)
- Docs: `docs/security.md` gains an "Input sanitization (opt-in)" section on the posture; `docs/router.md` and `docs/config.md` updated; scaffold `.env.example` documents the knob (commented, off).

Closes #271

## Acceptance criteria

- [x] Default `framework.App` stack has no xss layer; the ablation ladder shows no xss row (production default is `SanitizeInput=false`). `TestSanitizeInputOptInInstallsXSSLayer`, `TestDefaultRuntimeMiddlewareOrder`.
- [x] `{"description":"x < y"}` and `{"note":"<b>bold</b>"}` round-trip unchanged through a default app — `TestDefaultRouterDoesNotSanitizeInput` (body + query).
- [x] Opt-in flag restores today's behavior; existing `xss_test.go` unit cases pass unchanged, and the opt-in path is covered end-to-end by `TestSanitizeInputOptInSanitizesJSONBody` / `TestSanitizeInputOptInSanitizesQuery` (incl. the `password` exemption). `auth` `TestRawBodyPathThroughNew` now enables the opt-in so it still exercises the raw-body bypass.
- [x] `SanitizeHTML` exported and documented; scaffolded apps + `docs/` updated; CHANGELOG marks the default change as breaking for apps relying on ingress stripping.
- [x] #201 (data loss after an unclosed skip element) — already CLOSED/fixed; its `stripHTMLUnclosed` fix is retained unchanged on the opt-in path (it can no longer lose data on the default path, which does not sanitize).

## Scope notes

- **Breaking.** Apps relying on ingress stripping must set `GOMBIT_SECURITY_SANITIZE_INPUT=true` or escape on output (frontend/admin already do). Under Changed with a `BREAKING CHANGE` in the CHANGELOG.
- **Field-level resourcegen `sanitize` tag** is deferred to a follow-up (ADR-018 "Alternatives considered"); `SanitizeHTML` covers the handler-level need now.
- Interacts with #268/#270 (same `runtimeMiddlewareStack`); textual conflicts in `app.go` resolve mechanically (each PR guards its own layer independently).

## Validation

- [x] `go test ./...` — pass (goldentest regenerated via `-update`)
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run ./framework/... ./config/...` — 0 issues

- [x] New behavior has tests (default round-trip, opt-in sanitize, exported helper, opt-in stack); no DB-touching change
- [x] Stable features ship docs and appear in an example app — docs updated; scaffold documents the option and its golden trees are regenerated
- [x] Extraction preserves contracts — the sanitizer behavior is unchanged on the opt-in path
- [ ] Generators — N/A (template comment + golden snapshots updated; no generator logic changed)
- [ ] API changes regenerate OpenAPI + TS client — N/A, no HTTP API change
- [x] Scope stays inside milestone — post-v0.1 runtime/config, no M6 battery


docs/security.md
```diff
@@ -84,3 +84,46 @@ own handler is supported through the standard `http.Header` mutation APIs — us
...
+default app stores what the client sent, byte-for-byte: `{"description":"x < y"}`
+and `{"note":"<b>bold</b>"}` round-trip unchanged.
+
+This is deliberate. **XSS is an output-encoding problem, not an input problem.**
+The defense is escaping data at the point it is rendered:
+
+- The generated React frontend escapes text by default (JSX interpolation), and
+  the framework admin SPA renders values as text, not HTML.
+- Server-rendered admin templates escape untrusted data with Go's
+  `html/template` contextual escaping.
```
leo-aa88: **MAJOR (misleading security rationale).**

This PR turns off an on-by-default XSS protection and justifies it with "XSS is handled on output." Two output controls are named: the React SPA, and *server-rendered admin templates escaping untrusted data via `html/template` contextual escaping*. The first is real. The second describes nothing that exists.

`grep -rn 'html/template'` across the whole module returns **zero** hits. The only template import anywhere is `scaffold/generate.go`, and it's `text/template` — the code generator, not a runtime request path. There is no server-rendered HTML surface at runtime: the admin UI is a static React SPA served as embedded bytes through `c.Data` (`framework/adminui.go`), the API is JSON, `/metrics` is `text/plain`. The one runtime `html.EscapeString` call escapes the injected API prefix, not user data.

So the load-bearing defense after this change is React/JSX escaping in the SPA plus JSON responses not being an executable HTML context (backed by the #267 CSP). That actually holds — which is why this is MAJOR and not blocking. But as written, the next engineer who adds a server-rendered admin page will read this line, assume the framework contextually escapes for them, and ship stored XSS. It doesn't escape anything for them; there is no such path.

Fix: state the controls that exist (React/JSX escaping; JSON is not an HTML sink; CSP from #267) and delete the `html/template` sentence — or actually wire a server-rendered template path that uses it. The same claim is repeated in `config/config.go` and `docs/adr/018-input-sanitization-opt-in.md`; correct all three.

config/config.go
```diff
@@ -88,6 +89,23 @@ type Config struct {
...
 	Auth        AuthConfig
+	Security    SecurityConfig
+}
+
+// SecurityConfig contains opt-in application security toggles.
+type SecurityConfig struct {
+	// SanitizeInput installs the request-input HTML sanitizer (issue #271 /
+	// PERF-13). It is off by default: the framework does not rewrite request
+	// input. XSS is an output-encoding concern — the React frontend escapes
+	// text and admin templates escape untrusted data — and stripping markup on
```
leo-aa88: Same non-existent-control claim as flagged on `docs/security.md:101`: "admin templates escape untrusted data." There are no runtime `html/template` admin templates in this module (only `text/template` in the codegen). The real control is the React SPA + JSON output. Correct the rationale here too so the config doc doesn't promise escaping the framework doesn't perform.

docs/adr/018-input-sanitization-opt-in.md
```diff
@@ -0,0 +1,86 @@
...
+`io.ReadAll`, `bytes.NewReader`, `io.NopCloser`, and `mime.ParseMediaType`'s
+param map), and Huma already decodes the body twice (once into `any` for schema
+validation, once into the typed struct) — the sanitizer adds a third full read
+on the fast path and a fourth decode on the slow path. Removing it from the
+stack after the #241 content-type fast path took a valid POST from 46 to 42
+allocs/op.
+
+**It rewrites API input, which is the wrong layer for XSS defense.** XSS is an
+*output-encoding* concern. The generated React frontend escapes text by default
+and the admin surfaces render values as text or through `html/template`
```
leo-aa88: Same as `docs/security.md:101`: "the admin surfaces render values as text or through `html/template` contextual escaping." The `html/template` half documents a control that doesn't exist at runtime (no `html/template` anywhere in the module). The decision itself is fine — the React SPA escaping and JSON non-execution carry it — but an ADR that removes a default protection should cite the defenses it actually has. Drop the `html/template` clause.

leo-aa88 (commented):
# COMMENT

The mechanical move is clean: `xssMiddleware` becomes a conditional layer, the `password` exemption / 8 MiB cap / #201 fix / `WithRawBodyPaths` bypass all ride along untouched on the opt-in path, and the tests were genuinely re-pointed (the default-path test asserts `<script>...</script>` round-trips *verbatim*, not just that `x < y` survives — good, that's the actual contract). Config plumbing and golden regen are correct. The `auth` raw-body test flipping the opt-in on so it doesn't pass vacuously is exactly the right instinct.

One real problem, and it's in the justification rather than the code. This PR removes an on-by-default XSS defense; the entire case for that rests on "output is always escaped." One of the two named output controls does not exist in this repo. It doesn't sink the PR — the *other* control (React/JSX + JSON-is-not-an-HTML-sink) genuinely holds — but a security-posture change must not document a control it doesn't have. Details inline. Fix the three copies of the claim and this is good to merge.

framework/app.go
```diff
@@ -726,9 +726,18 @@ func runtimeMiddlewareStack(cfg config.Config, metrics *httpMetrics, csrfExemptP
 			name:    "security_headers",
 			handler: securityHeadersMiddleware(cfg.Environment == config.EnvironmentProduction, cfg.API.DocsEnabled),
 		},
-		// Raw-body paths (webhooks) skip input sanitization so their body reaches
-		// the handler unmodified for signature verification (WithRawBodyPaths).
-		{name: "xss", handler: xssMiddleware(rawBodyPaths...)},
+	}
+	// Input sanitization is opt-in (issue #271 / PERF-13). The default pipeline
```
leo-aa88, line 761: **BLOCKING — this removes the only request-size bound from raw Gin JSON routes.** I reproduced this against current `main` and this exact HEAD with an `app.Router().POST` handler using `ShouldBindJSON`: a JSON body just over 8 MiB is rejected on `main` with 413 before handler execution; on this branch it is fully decoded, the handler runs, and the response is 204. Huma’s own 1 MiB operation limit does not cover the first-class raw `app.Router()` escape hatch, and `ReadTimeout` limits time, not memory. The old cap may have existed to bound sanitizer buffering, but removing the sanitizer silently removes the observable protection too. Do not make applications enable input mutation to regain a basic memory-exhaustion bound. Split body-size enforcement from sanitization (for example, a non-decoding `MaxBytesReader` layer), and add adversarial tests for both raw Gin and Huma routes.

docs/adr/018-input-sanitization-opt-in.md
```diff
@@ -0,0 +1,86 @@
...
+- gives false confidence (it does not encode for any specific output context);
+- forces webhook paths to opt out via `WithRawBodyPaths` because re-encoding the
+  body breaks signature verification.
+
+## Decision
+
+**Move request-input sanitization out of the default pipeline; offer it as
+explicit, narrower opt-ins.**
+
+1. **Default pipeline has no `xss` layer.** A default `framework.App` does not
```
leo-aa88, line 47: **BLOCKING — the PR changes the architecture but leaves the repository’s authoritative contract saying the opposite.** `docs/GOMBIT_BUILD_PLAN.md` §4 still requires M1-8 sanitization in the *runtime default stack* and calls it a fundamental security default; `docs/lifecycle.md` repeats that invariant; `examples/router/main.go` still tells users its raw Gin echo is sanitized by default. This repository explicitly declares the build plan authoritative, and the working agreement requires stable changes to update docs and an example app. Right now ADR-018 says opt-in while the source of truth and shipped example say on-by-default. Update the build plan/lifecycle record and the example in this PR so there is one contract, not two.

docs/security.md
```diff
@@ -84,3 +84,46 @@ own handler is supported through the standard `http.Header` mutation APIs — us
 through the slice `Header.Values(key)` returns: those backing slices are shared
 process-globals, and an in-place write corrupts the value for every other
 in-flight request. `TestSecurityHeaderSharedValueContract` locks this contract.
+
+## Input sanitization (opt-in)
+
+Gombit does **not** sanitize request input by default (issue
+[#271](https://github.com/gombit-dev/gombit/issues/271) / PERF-13; ADR-018). A
+default app stores what the client sent, byte-for-byte: `{"description":"x < y"}`
```
leo-aa88: **MAJOR — this does not establish byte-for-byte round-tripping.** A default app does not store anything, and the added test decodes the request with `ShouldBindJSON`, re-encodes it with `c.JSON`, then compares decoded field values. Whitespace, key order, number spelling, and duplicate-key representation are not preserved. Byte fidelity is a separate contract provided only when a handler reads the raw body (and is why `WithRawBodyPaths` exists). Say that default middleware no longer mutates JSON/query *values*, or test and implement actual byte preservation; do not advertise the stronger contract based on a semantic-value test.

leo-aa88 (commented):
# REQUEST CHANGES

The branch now rebases cleanly onto current `main`, and the opt-in switch behaves as implemented. It still does not deserve merge: removing the sanitizer also removes a real request-size safety bound from raw Gin routes, while the repository’s authoritative architecture, examples, and security rationale contradict the new behavior.

## 1. Default raw Gin JSON routes lose their only body-size bound

**BLOCKING.**

The old default `xss` layer read at most 8 MiB + 1 and returned a D10 413 before handlers ran. Making that layer conditional removes the bound with it. I reproduced the difference with the same raw `app.Router().POST` handler calling `ShouldBindJSON`: current `main` returns 413 without invoking the handler; this HEAD accepts and decodes a body over 8 MiB, invokes the handler, and returns 204.

Huma’s operation-level limit does not protect raw Gin, and raw Gin is a first-class documented escape hatch. `ReadTimeout` is not a memory bound. Separate size enforcement from data mutation and test both dispatch paths. Applications should not have to opt into corrupting input to avoid unbounded body allocation.

## 2. The authoritative build plan and shipped example still require the opposite default

**BLOCKING.**

ADR-018 declares sanitization opt-in, but `docs/GOMBIT_BUILD_PLAN.md` §4 still declares M1-8 an on-by-default, fundamental security invariant. `docs/lifecycle.md` repeats it. `examples/router/main.go` still tells users that its raw Gin echo receives sanitized values by default.

The repository explicitly names the build plan as the source of truth, and issue #271 requires documentation plus example coverage. An ADR cannot silently fork the contract. Update the authoritative plan, lifecycle documentation, and executable example as part of this breaking change.

## 3. The security rationale cites a nonexistent output control

**MAJOR.**

The earlier inline threads remain unresolved: the PR repeatedly says server-rendered admin templates use `html/template` contextual escaping. There is no runtime `html/template` import or server-rendered admin path; admin is a React SPA. The actual controls are JSX text escaping, JSON not being an HTML sink, and the response CSP. A security decision that removes a default protection must name controls that actually exist.

## 4. The tests prove value preservation, not byte-for-byte round-tripping

**MAJOR.**

`TestDefaultRouterDoesNotSanitizeInput` decodes with `ShouldBindJSON`, re-encodes with `c.JSON`, and compares decoded strings. That proves the middleware no longer strips those values. It does not preserve JSON bytes: whitespace, key order, number spelling, and duplicate keys may all change. Rewrite the public claim as value preservation unless raw-byte identity is actually implemented and tested.

The changed package tests pass. This is not a green-CI question: the suite does not test an oversized default raw-Gin request, and it cannot validate contradictory prose.

# VERDICT

The central implementation change is incomplete at the architecture boundary. Sanitization, request-size enforcement, and output escaping are three different concerns; this PR removes the first and accidentally removes the second while documenting an imaginary instance of the third. It also leaves the source of truth and example app advertising the old contract.

Do not merge until the body-size regression is fixed and the repository has one accurate security contract.

framework/http_middleware.go
```diff
@@ -28,6 +28,65 @@ const traceIDLen = 32
...
+		// requests.
+		if c.Request.ContentLength > maxRequestBodyBytes {
+			abortWithErrorEnvelope(c, contract.PayloadTooLarge("JSON body exceeds the 8MiB limit"))
+			return
+		}
+		// A chunked or misdeclared body (ContentLength < 0, or a lying length) is
+		// bounded during read instead: MaxBytesReader errors past the cap, so a
+		// handler's ShouldBindJSON / Huma's decode fails rather than allocating
+		// without limit.
+		c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, maxRequestBodyBytes)
```
leo-aa88: **BLOCKING — the streamed-body path does not provide the claimed pre-dispatch D10 rejection.** The `ContentLength > cap` fast path works, but for `ContentLength == -1` (ordinary chunked transfer) this wraps the body and immediately calls `c.Next()`. I reproduced an oversized chunked `application/json` request on this exact HEAD: the raw Gin handler ran, `io.ReadAll` received `http: request body too large`, the handler ignored it, and the response was 204. The same test against current `main` returns 413 without entering the handler because the old sanitizer reads the bounded body before dispatch. So the code says “413 before any handler runs,” while runtime says “the handler may notice a read error if it bothers checking” — and no D10 envelope is guaranteed. Pre-read at most cap+1 on the unknown-length path before `c.Next()` (the normal known-length fast path can stay allocation-cheap), restore the body when within bounds, and add a chunked/`ContentLength=-1` regression test asserting both D10 413 and zero handler execution.

framework/http_middleware.go
```diff
@@ -28,6 +28,65 @@ const traceIDLen = 32
...
+// signature-verifying webhook keeps its exact bytes, exactly as before.
+func requestBodyLimitMiddleware(exemptPaths ...string) gin.HandlerFunc {
+	exempt := make(map[string]struct{}, len(exemptPaths))
+	for _, path := range exemptPaths {
+		if path = strings.TrimSpace(path); path != "" {
+			exempt[path] = struct{}{}
+		}
+	}
+	return func(c *gin.Context) {
+		if _, ok := exempt[c.Request.URL.Path]; ok {
```
leo-aa88: **BLOCKING — raw-byte fidelity is not a reason to disable the size bound on public webhooks.** `http.MaxBytesReader` does not decode, re-encode, or otherwise alter any accepted bytes; it only refuses data beyond the limit. `WithRawBodyPaths` exists because the sanitizer mutates JSON and because cookie CSRF cannot cover server callbacks. Reusing that list here couples unrelated policies and deliberately leaves signature-verifying endpoints—the attacker-facing routes that must read the body before rejecting a bad HMAC—open to unbounded reads. The public `WithRawBodyPaths` comment promises sanitizer and CSRF exemptions, not “disable request-size protection.” Apply the fixed limit to raw-body paths too; if oversized webhooks genuinely need a different policy later, add an explicit body-limit configuration rather than smuggling an unlimited-body opt-out through an input-fidelity option.

leo-aa88 (commented):
# REQUEST CHANGES

The previous documentation, source-of-truth, value-fidelity, and known-length raw-Gin findings are fixed at `be04331b`. The replacement body-limit abstraction is still not correct: its strongest guarantees disappear for streamed requests, and it deliberately disables protection on the webhooks most exposed to attacker-controlled bodies.

## 1. Chunked bodies reach handlers and do not get the promised D10 413

**BLOCKING.**

The middleware rejects only when a declared `Content-Length` is already over 8 MiB. For `ContentLength == -1`—normal chunked transfer—it installs `MaxBytesReader` and calls `c.Next()`. The handler therefore runs before size is known, and the framework never owns the resulting error or response.

I reproduced this on the exact HEAD with an oversized chunked `application/json` request to a raw Gin handler. The handler ran, `io.ReadAll` returned `http: request body too large`, the handler ignored that error, and the response was 204. The same test against current `main` returns D10 413 without handler execution because the old sanitizer reads the bounded body before dispatch.

That directly contradicts the new middleware comment, ADR, router documentation, and CHANGELOG claim that oversized JSON is rejected with D10 413 before any handler runs. Pre-read at most cap+1 on the unknown-length path before `c.Next()` and restore acceptable bodies; keep the cheap declared-length path. Add a `ContentLength=-1` regression test. The current tests all use `strings.Reader`, which supplies a known length, so they prove only the easy branch.

## 2. `WithRawBodyPaths` silently becomes an unlimited-body opt-out

**BLOCKING.**

`WithRawBodyPaths` exists to prevent sanitizer re-encoding and to exempt server callbacks from cookie CSRF. `MaxBytesReader` does not mutate any accepted bytes, so raw-byte signature verification does not require bypassing it. Reusing `rawBodyPaths` in the new limiter couples unrelated policies and leaves public webhook handlers free to read arbitrarily large bodies before rejecting an invalid HMAC.

Worse, the exported `WithRawBodyPaths` contract still mentions only byte preservation, sanitizer bypass, and CSRF exemption. It does not tell callers that declaring a webhook disables request-size protection. Apply the fixed limit to raw paths too. If a future application genuinely needs larger bodies, give body limits their own explicit configuration instead of hiding an unlimited-body switch inside an input-fidelity option.

The changed-package tests, focused race tests, vet, and all GitHub CI checks pass. CI is green. This is not a failing-test problem. The current tests omit chunked transfer and explicitly encode the unsafe raw-path exemption as desired behavior.

# VERDICT

Separating sanitization from size enforcement is the correct architecture. This implementation stops one layer short: it wraps reads instead of enforcing the advertised response/dispatch contract, and it carries a sanitizer-specific exemption into an independent security boundary.

Do not merge until oversized streamed bodies are rejected deterministically before dispatch and raw-byte paths retain the body-size bound.

framework/http_middleware.go
```diff
@@ -28,6 +30,80 @@ const traceIDLen = 32
...
+		// oversized one is rejected with a D10 413 before c.Next(), so the
+		// framework owns the response instead of leaving the handler to notice a
+		// read error it may ignore.
+		buf, err := io.ReadAll(io.LimitReader(c.Request.Body, maxRequestBodyBytes+1))
+		_ = c.Request.Body.Close()
+		if err != nil {
+			// A read failure on the stream: hand the handler/Huma an empty body so
+			// it emits its own validation error rather than a broken read.
+			c.Request.Body = io.NopCloser(bytes.NewReader(nil))
+			c.Next()
```
leo-aa88: **BLOCKING — a transport read failure is erased and dispatched as a valid empty request.** On this path the middleware has taken ownership of reading the stream. Replacing a partial/broken body with `""` and calling `c.Next()` does not make Huma/Gin reliably emit validation: raw handlers may accept empty input or ignore binding errors, and operations with optional bodies can execute. I reproduced a body that yields partial JSON and then `io.ErrUnexpectedEOF` on this exact HEAD; the handler ran with an empty body and returned 204. The framework has converted malformed transport state into different application input. Abort here with a stable D10 client error and never dispatch, then add a failing-reader regression test asserting the handler does not run. Do not copy the sanitizer's old empty-body workaround into the independent size boundary.

framework/http_middleware.go
```diff
@@ -28,6 +30,80 @@ const traceIDLen = 32
...
+			return
+		}
+		if int64(len(buf)) > maxRequestBodyBytes {
+			c.Request.Body = io.NopCloser(bytes.NewReader(nil))
+			abortWithErrorEnvelope(c, contract.PayloadTooLarge("JSON body exceeds the 8MiB limit"))
+			return
+		}
+		c.Request.Body = io.NopCloser(bytes.NewReader(buf))
+		c.Request.ContentLength = int64(len(buf))
```
leo-aa88, line 131: **MAJOR — this constructs a request state that the HTTP server never produces.** A real chunked server request arrives with `ContentLength == -1` and `TransferEncoding == ["chunked"]`. After buffering, this line changes only `ContentLength`, so the handler receives both a positive content length and chunked transfer encoding. I reproduced `ContentLength=11 TransferEncoding=[chunked]` on this exact HEAD. That contradictory metadata is observable to raw Gin handlers and can leak into code that forwards or signs the request. Restoring the bytes does not require rewriting framing metadata: preserve `ContentLength == -1`, or deliberately normalize the whole representation by clearing `TransferEncoding` as well. Add a test using an actual chunked-shaped request; the current test sets only `ContentLength = -1`, so it cannot catch this invariant violation.

leo-aa88 (commented):
# REQUEST CHANGES

The previous oversized-chunked-body and raw-webhook exemptions are fixed. This HEAD still gets the unknown-length error path wrong: after taking ownership of the stream, the middleware invents a different request and dispatches it. It also leaves actual chunked requests in a framing state that `net/http` would never parse from the wire.

CI is green. This is not a failing-test problem. The current tests cover readable under-cap and oversized streams, but they do not exercise read failure or real chunked framing metadata.

## 1. Broken streams become successful empty requests

**BLOCKING.**

`io.ReadAll` can return partial bytes and an error. The middleware discards both, replaces the body with an empty reader, and invokes the handler. On this exact HEAD, a reader returning partial JSON followed by `io.ErrUnexpectedEOF` reached a raw Gin handler as `""`; the handler returned 204. "Huma/Gin will validate it" is not an invariant—raw handlers and optional-body operations can accept the fabricated empty input.

Once this middleware owns the read, it must own the failure: abort before dispatch with a stable D10 client error and test that no handler executes.

## 2. Body restoration creates contradictory HTTP framing

**MAJOR.**

An incoming chunked request has `ContentLength == -1` and `TransferEncoding == ["chunked"]`. The middleware changes only `ContentLength`, producing a request with both a positive length and chunked transfer encoding. I reproduced `ContentLength=11 TransferEncoding=[chunked]` in the handler. Preserve the original unknown length, or normalize both fields deliberately; do not leave downstream raw handlers and forwarding/signing code an impossible representation.

# VERDICT

The sanitization split is now mostly coherent, but the independent body-limit abstraction still has broken error ownership and request-representation invariants. These are implementation and test defects in an always-on middleware boundary, not cosmetic cleanup.

Do not merge `fcc8430`.

framework/http_middleware.go
leo-aa88: Still **BLOCKING** on `e55585d`. I reran the failing-stream probe against this exact HEAD: partial JSON followed by `io.ErrUnexpectedEOF` still invoked the raw handler with an empty body and returned 204. The new commit does not touch this branch or add the missing read-error test.

framework/http_middleware.go
leo-aa88: Still **MAJOR** on `e55585d`. I reran the chunked-shaped request against this exact HEAD and the handler still observed `ContentLength=11 TransferEncoding=[chunked]`. The added known-length hardening does not normalize or preserve the unknown-length request metadata, and the submitted chunked test still omits `TransferEncoding`.

framework/http_middleware.go
```diff
@@ -28,6 +30,99 @@ const traceIDLen = 32
 
 var traceparentPattern = regexp.MustCompile(`^[0-9a-f]{2}-([0-9a-f]{32})-[0-9a-f]{16}-[0-9a-f]{2}$`)
 
+// maxRequestBodyBytes bounds a JSON request body (8 MiB). It caps the memory a
+// single request can force the framework — or a raw app.Router() handler that
+// calls ShouldBindJSON — to allocate. This size bound used to live incidentally
+// inside the input sanitizer's buffering; issue #271 / PERF-13 made
+// sanitization opt-in, so the bound now stands on its own in
+// requestBodyLimitMiddleware and protects every app whether or not sanitization
```
leo-aa88: **MAJOR — `protects every app` is false across the first-class `WithRouter` escape hatch.** `New` installs `runtimeMiddlewareStack` only when it constructs the default router; for a supplied router it calls only `configureTrustedProxies`. I reproduced an oversized JSON request through `New(WithRouter(gin.New()))` on this exact HEAD: the handler read it, ran, and returned 204. The custom-router architecture is intentionally application-owned, so do not silently change that boundary. Qualify the new `always-on` / `every app` / `every route` claims as applying to the default router, and add `request_body_limit` to `docs/router.md`'s list of middleware the application owns under `WithRouter`. A boundary test should lock this so the security documentation cannot drift again.

leo-aa88 (commented):
# REQUEST CHANGES

The `e55585d` self-review hardens a synthetic lying-`Content-Length` path and documents the JSON-only scope, but it leaves both active runtime defects untouched and adds another false boundary claim. The default-path performance result remains credible: the new body-limit row costs one allocation on valid/invalid POST, so removing the sanitizer still nets roughly four allocations. Performance is not the blocker. Correctness is.

CI is green. This is not a failing-test problem. The current tests do not exercise stream read failure, real chunked framing metadata, or the first-class `WithRouter` boundary.

## 1. Broken streams still become successful empty requests

**BLOCKING.**

On `io.ReadAll` failure, the middleware discards both the partial bytes and the error, substitutes an empty body, and dispatches. Reproduced on this exact HEAD: partial JSON followed by `io.ErrUnexpectedEOF` reached the raw handler as `""`, and the handler returned 204. Once the middleware owns the read, it must own the failure and abort with a stable D10 client error before application code executes.

## 2. Restored chunked requests still have impossible framing

**MAJOR.**

The middleware changes `ContentLength` from `-1` to a positive value without clearing `TransferEncoding: chunked`. Reproduced on this exact HEAD: the handler observed `ContentLength=11 TransferEncoding=[chunked]`. Preserve the original unknown length or normalize both fields; the current test sets only `ContentLength=-1`, so it does not model a real chunked request.

## 3. The supposedly universal limit disappears under `WithRouter`

**MAJOR.**

`New` installs `runtimeMiddlewareStack` only for the default router. With a supplied router it configures trusted proxies and nothing else. Reproduced on this exact HEAD: an oversized JSON body through `New(WithRouter(gin.New()))` reached the handler and returned 204. The escape-hatch architecture is intentional, but the new `always-on`, `every app`, and `every route` language is false. Qualify those claims and document `request_body_limit` among the layers a custom-router application owns.

# VERDICT

The deepest problem remains ownership at the middleware boundary: the implementation consumes request state, then either fabricates different input or exposes contradictory metadata. Separately, the documentation erases the established default-router/custom-router distinction.

Do not merge `e55585d`.

leo-aa88:
LGTM

PR #359, opened 2026-09-15
feat(resourcegen): model-first DTO + mapper engine — RESGEN-1 slice 3 (#352)

leo-aa88:
## Summary

Slice 3 of the ADR-016 model-first, generator-owned resource epic. It adds the engine that **derives the request DTO, the response DTO, and the model↔DTO mappers from the GORM model + its resolved `resourcepolicy`** — not from the CLI `name:type` grammar. This is the model-derived replacement for the hand-frozen `bookData` / create-input / `toBookData` that drift from the model and cause #218.

For a `Book{ gorm.Model; Title; TenantID gombit:"read,server"; Password gombit:"write"; Internal gombit:"-" }` it emits:

- `bookData` (response): `ID, CreatedAt, UpdatedAt, Title, TenantID` — write-only `Password`, hidden `Internal`, and soft-delete `DeletedAt` are absent.
- `bookCreateBody` (request): `Title, Password` — the server-managed `TenantID`, the read-only key/timestamps, and hidden columns are absent (the #218 split, model-derived).
- `toBookData` (composite literal) and `bookFromCreateBody` (assignments, so an embed path like `row.Audit.Note` is valid Go), setting only request columns.

New (all in package `resourcegen`, unexported):
- `buildModelResource` — `schema.Parse` + `resourcepolicy.FromSchema`/`ResolveAll`; persistence facts (column, Go type, embed bind path) from the parsed schema, read/write/server policy from `resourcepolicy`. Fails closed on contradictory/unsatisfiable policy and on a same-named leaf collision.
- `renderGoType` — `reflect.Type` → Go type expression + import paths (scalars, `time.Time`, `database/sql` null types, framework decimal, pointers, `[]byte`); fails closed on undecidable types or an ambiguous package base name.
- `renderModelDTOs` — emits one generator-owned file (DO-NOT-EDIT banner + grouped imports).

Related #352

## Acceptance criteria

This slice is one increment of epic #352; it advances, and does not yet fully close, the epic's two hard criteria:

- [x] Foundation for **"changing a model field + policy requires no manual sync"**: DTOs and mappers are now *derived* from the model, so a field/policy change flows into them by regeneration. (Enforced end to end once wired into `make resource` in slice 4 and gated by `generate --check` in slice 5.)
- [x] Generated output carries the DO-NOT-EDIT banner and is byte-deterministic (golden + determinism tests) — the precondition for the regenerate-and-compare drift gate.

## Scope notes

Deferred, by the epic's slice plan:
- **Slice 4** — the CRUD handler + hooks that consume these DTOs (incl. Huma input/output wrappers and `BeforeCreate` for server columns), and wiring this engine into `gombit make resource` to emit `*.gen.go`.
- **Slice 5** — `gombit generate --check` drift gate.
- **Slice 6** — remove `resourcecheck` + the migration guide.

Not attempted here (kept minimal): richer Huma validation tags (maxLength/enum from schema facts), list/filter/sort/aggregate surfaces, and relationship DTOs. The engine is not yet called by any command, so no generated app output changes (goldentest is untouched).

## Validation

- [x] `go test ./resourcegen/ ./resourcepolicy/ ./goldentest/` — pass (incl. a hermetic `go build` of the generated DTOs against a matching model; goldentest unchanged)
- [x] `go build ./...`, `go vet ./resourcegen/ ./resourcepolicy/`
- [x] `golangci-lint run ./resourcegen/` — 0 issues

- [x] New behavior has tests — pure engine; no DB-touching runtime code (schema parsing is in-process, exercised on SQLite-independent models)
- [ ] Stable features ship docs and appear in an example app — N/A until the engine is wired into `make resource` (slice 4); docs + migration guide land in slice 6
- [x] Extraction preserves contracts — additive; the legacy CLI generator path is untouched
- [x] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files — engine renders bytes only; no file writes/AST edits in this slice
- [ ] API changes regenerate the OpenAPI doc + TS client — N/A, no API/handler wired yet
- [x] No secrets in generated frontend source — N/A (no frontend)
- [x] Scope stays inside this issue's milestone — post-v0.1 epic #352, no M6 battery creep


resourcegen/modelgen.go
```diff
@@ -0,0 +1,282 @@
...
+		name := t.Name() // a type with a package path is named
+		base := pkgBase(pkgPath)
+		if !isGoIdent(base) {
+			// reflect exposes the import path, not the package's declared name; when
+			// the path's last segment is not a usable identifier (e.g. gopkg.in/x.v2)
+			// we cannot emit a correct qualifier. Fail closed rather than emit code
+			// that will not compile.
+			return "", nil, fmt.Errorf("resourcegen: cannot derive a package qualifier for %q (path segment %q is not a Go identifier); this column type is not supported yet", t.String(), base)
+		}
+		return base + "." + name, []string{pkgPath}, nil
```
leo-aa88: **BLOCKING — this is not a Go package-name resolver.** `PkgPath`'s last path segment is not the declared package identifier, and it is also the current package for model-local named types. I reproduced both failures on this exact HEAD: a local `type ReviewStatus string` makes package `resourcegen` import `github.com/gombit-dev/gombit/resourcegen` and emit `resourcegen.ReviewStatus` (a self-import/import cycle); a type from path `.../reviewtypesv2` whose declaration is `package renamed` is emitted as `reviewtypesv2.Column`, which does not exist. The current named-composite test actually blesses the self-import and never compiles that generated case. Carry destination/model package identity and explicit import aliases in the representation: render local named types unqualified, and qualify external types with their actual or assigned alias. Rejecting only path basenames that are not identifiers does not fail closed.

resourcegen/modelgen.go
```diff
@@ -0,0 +1,282 @@
...
+			base := pkgBase(imp)
+			if other, ok := pkgBaseByPath[base]; ok && other != imp {
+				return modelResource{}, fmt.Errorf("resourcegen: imports %q and %q share package name %q; DTO generation cannot disambiguate them yet", other, imp, base)
+			}
+			pkgBaseByPath[base] = imp
+			imports[imp] = struct{}{}
+		}
+		res.Fields = append(res.Fields, modelField{
+			GoName:     r.GoName,
+			AccessPath: strings.Join(f.BindNames, "."),
```
leo-aa88, line 159: **BLOCKING — flattening the bind path to a string loses the pointer/nullability information the mapper needs.** GORM accepts pointer-embedded models. For an embedded `*ReviewAudit` field tagged `gorm:"embedded"`, this records `ReviewAudit.Note`; the create mapper zero-initializes the row and executes `row.ReviewAudit.Note = ...`, which I compiled and ran on this exact HEAD: it panics with a nil-pointer dereference. `to...Data` can likewise panic when the embedded pointer is nil. 'Valid Go' is not runtime correctness. Preserve structured path segments with pointer metadata and initialize or nil-handle them, or reject pointer embeds during build until supported. Add an execution test for the generated mapper; substring and parser tests cannot catch this.

resourcegen/modelgen.go
```diff
@@ -0,0 +1,282 @@
...
+	policyFields, err := resourcepolicy.FromSchema(sch)
+	if err != nil {
+		return modelResource{}, err
+	}
+	resolved, err := resourcepolicy.ResolveAll(policyFields)
+	if err != nil {
+		return modelResource{}, err
+	}
+
+	res := modelResource{Package: pkg, TypeName: typeName}
```
leo-aa88, line 129: **BLOCKING — the alleged model-derived representation still trusts an independent model name.** `schema.Parse` already knows the model identity, but this line copies arbitrary caller input. I passed `&ReviewNamedModel{}` with `typeName="DifferentType"`; the build succeeded and returned a resource whose fields came from `ReviewNamedModel` while every emitted DTO and mapper targets `DifferentType`. That is exactly the duplicate-representation drift ADR-016 is supposed to eliminate. Derive the model name and package identity from `sch.ModelType` (and validate any intentionally separate output destination) so a `modelResource` cannot describe one model's fields under another model's name.

leo-aa88 (commented):
# REQUEST CHANGES

This slice claims a model-derived, fail-closed DTO/mapper engine, but its representation omits package identity and pointer lifetime. It emits source that cannot compile for ordinary named field types and mappers that panic for valid GORM pointer embeds.

CI is green. This is not a failing-test problem. The current tests do not exercise local named types, package-name/path mismatch, caller/model identity disagreement, or pointer-embedded mapper execution.

## 1. Named field types are qualified with a guessed package name

**BLOCKING.**

`renderGoType` assumes the import path basename is the package identifier. Go does not guarantee that, and a model-local named type must not be imported or qualified at all. This produces uncompilable generated code for both ordinary local enum types and external packages whose declared name differs from their path basename.

The representation needs destination/model package identity plus explicit deterministic aliases. The existing test merely asserts the incorrect string; it never compiles the result.

## 2. Pointer-embedded models produce mappers that panic

**BLOCKING.**

`buildModelResource` collapses GORM's bind path to `ReviewAudit.Note`, discarding that `ReviewAudit` is a pointer. The generated create mapper dereferences that nil pointer immediately after zero-initializing the model; the response mapper has the same unsafe access when the embedded pointer is nil.

Carry structured access segments with pointer metadata and generate initialization/nil behavior, or reject pointer embeds until they are supported. The mapper must be compiled and executed in a regression test.

## 3. Model identity is still duplicated caller input

**BLOCKING.**

Fields come from `schema.Parse(model)`, but the emitted model name comes from an unrelated `typeName` argument without validation. The engine can therefore describe one model's fields while generating mappers for another type. That violates this issue's one-authoritative-representation contract before slice 4 even wires it into the CLI.

Derive identity from `sch.ModelType`; validate any genuinely separate output-package choice explicitly.

# VERDICT

The deepest defect is architectural: real model facts are reduced to strings while identity and lifetime are supplied or discarded elsewhere. The resulting `modelResource` can represent contradictory and runtime-unsafe states, so the next slice cannot safely treat it as an authoritative intermediate representation.

Do not merge PR 359. Fix the representation and add compile-and-execute adversarial tests first.

Brunotlps (commented):
Superseded — reposted below with the correct review state: https://github.com/gombit-dev/gombit/pull/359#pullrequestreview-5217465817

Brunotlps (changes requested):
# REQUEST CHANGES

One defect, three attempts against it: the intermediate representation stores strings where the pipeline needs facts. `58b3886` fixed that for the type expression and left it intact for package identity and policy provenance. The engine emits correct-looking Go on the happy path and is structurally unable to hand slice 4 the facts slice 4 needs.

The current tests do not exercise package identity, import emission, unexported type names, or policy provenance.

---

## 1. `modelResource` is a lossy projection, and slice 4 is the one that pays

**BLOCKING.**

`resourcepolicy.Resolved` (`resourcepolicy/policy.go:143-148`) carries `FieldFacts` + `InRequest` + `InResponse` + `CreateSource`. `buildModelResource` projects that into six fields (`resourcegen/modelgen.go:131-138`):

```go
res.Fields = append(res.Fields, modelField{
    GoName: r.GoName, AccessPath: ..., Column: r.Column,
    GoType: goType, InRequest: r.InRequest, InResponse: r.InResponse,
})
```

`CreateSource`, `NotNull`, `HasDefault`, `PrimaryKey`, `Creatable`, `Readable` — all discarded.

Take the golden model. `TenantID` is `gombit:"read,server"` → `InRequest=false, InResponse=true, CreateSource=Server`. `CreatedAt` is a GORM auto-timestamp → `InRequest=false, InResponse=true, CreateSource=None`.

In the emitted `modelResource` those two fields are byte-for-byte identical. Slice 4 has to generate the `BeforeCreate` signature listing the server-managed columns, and it cannot tell `TenantID` (the hook must set it) from `CreatedAt` (GORM sets it) from this representation. It will have to re-parse the model and re-resolve the policy.

That is exactly the duplicated derivation ADR-016 exists to abolish. This type's doc comment (`modelgen.go:27-31`) invokes *"ADR-016's one-authoritative-representation rule"* while the constructor, 100 lines down, throws the authoritative facts away.

**Suggestion:** keep `resourcepolicy.Resolved` (or at minimum `CreateSource` plus the nullability facts) on `modelField`. The projection should *add* generator-specific facts — `AccessPath`, `GoType`, `imports` — not *subtract* policy ones.

## 2. The package clause is derived from the exact premise this commit declares invalid

**BLOCKING.**

`modelgen.go:54-59`, in this diff, in this file:

> *"reflection exposes an import path, not the package's declared name — the two can differ (e.g. a package at `.../foo.v2` declared `package foo`), and a bare import would bind the wrong identifier."*

`modelgen.go:88`:

```go
pkg := pkgBase(sch.ModelType.PkgPath())
```

`modelgen.go:103` → `modelgen.go:309`:

```go
b.WriteString("package " + r.Package + "\n\n")
```

A model in `internal/models/` declared `package book`: `pkgBase` → `"models"`, `isGoIdent("models")` passes, and the generated file opens with `package models` beside a `model.go` that says `package book`. It does not compile. Go does not require the directory name to match the package name, and nothing here checks.

`reflect` does not expose the declared package name. This derivation is not merely wrong — it is unimplementable. The original finding asked for the right thing:

> *"Derive the model name and package identity from `sch.ModelType` (and validate any intentionally separate output destination)."*

The commit did the first half and did not implement the second.

**Suggestion:** the generator knows where it is writing the file — take the destination package as an explicit parameter and validate it (an identifier, not a keyword, and — since slice 4 writes beside `model.go` — cross-checked against the package clause already on disk). `TypeName` from `sch.ModelType.Name()` is correct and should stay.

## 3. `render` qualifies unexported names, and the test asserts the uncompilable result

**BLOCKING.**

`modelgen.go:223-231`:

```go
name := t.Name()
if t.PkgPath() == tr.modelPkgPath { return name, nil }
alias, err := tr.aliasFor(t.PkgPath())
return alias + "." + name, nil
```

No check that `name` is exported. And `modelgen_test.go:85-89` enshrines it:

```go
extExpr, extImports := renderExternal(t, nss)
if extExpr != "resourcegen.namedStringSlice" { ... }
```

`namedStringSlice` is lowercase. `resourcegen.namedStringSlice` does not compile from any other package. Nothing in the suite compiles that output, so the assertion stands unchallenged.

The self-import half was fixed (`:224-225`, local → unqualified, correct). The unexported half was left in place, and the test still asserts it as the expected output. A test that freezes source that cannot compile is worse than no test: it will defend the defect against the next person who tries to fix it.

Reachable through a type alias to an unexported type in a dependency (`type Exported = unexported` — `reflect` reports `unexported`). Rare, but the emitted string is unconditionally wrong when it happens.

**Suggestion:** fail closed in `render` when `name` is unexported and the package is not the model's. Change the test to an exported external type (`types.Decimal` already does this correctly at `modelgen_test.go:48`) and add a rejection case for the unexported one.

## 4. The alias/import mechanism is not compiled by any test

**MAJOR.**

This is the machinery `58b3886` exists to add. Its coverage:

- `TestRenderModelDTOsGolden` (`modelgen_test.go:267-327`) — a single import, `import "time"`, which takes the *bare* branch of `importSpec.line()` (`modelgen.go:66-68`).
- `TestGeneratedDTOsCompileAndRun` (`:384-422`) — the comment at `:376-379` states it: *"All field types are predeclared, so the temp module needs no external requires."* Zero imports in the only compiled output.
- `TestTypeRendererAliasFallback` (`:108-118`) — calls `aliasFor` directly and checks `isGoIdent` and distinctness. It never renders a file.

So `aliasFor → importSpec → line() → renderImports` for third-party paths produces no source that any test compiles. Which incorrect implementations still pass? Emitting `"path" alias` instead of `alias "path"`. Omitting the alias in the third-party group. Placing third-party specs in the std group. All green.

**Suggestion:** extend `TestGeneratedDTOsCompileAndRun` with a model carrying `time.Time` (std, bare), `types.Decimal` (third-party, aliased), and two distinct packages sharing a basename (uniquified alias). Compile and execute it. A `require` on the framework's own module is not a hermeticity problem — it is already in `go.mod`.

## 5. Validation constraints regress against the generator this replaces

**MAJOR.**

`modelgen.go:320` and `:328` emit exactly two tags:

```go
"`json:\"" + f.Column + "\" doc:\"" + f.GoName + "\"`"
```

The legacy path this engine is meant to supersede emits more (`resourcegen/fields.go:604-624`): `minLength:"1"` for a required string, `maxLength:"255"`, `minimum:"0"` for uint, `enum:"..."`.

The PR lists this as deferred. Acceptable in general — but look at which constraint was dropped. `minLength:"1"` on a NOT NULL string is exactly what rejects `""` on create. Without it, the client sends `{"title": ""}`, `resourcepolicy` is satisfied (a create source exists), GORM inserts the empty string, NOT NULL passes, and the row is wrong.

Silent zero-fill of a NOT NULL column on create is symptom #3 in issue #218 and in ADR-016's Context. The engine built to eliminate #218 currently emits a request DTO that reproduces it, while the generator it replaces did not.

Note also that `NotNull` and `HasDefault` are available on `resourcepolicy.FieldFacts` (`policy.go:54-55`) and are discarded by finding 1. The facts exist; the representation throws them away; the DTO cannot express the constraint. The three findings are the same finding.

**Suggestion:** either carry the facts and emit the constraint in this slice, or state explicitly in the PR that slice 3 knowingly ships a request DTO weaker than the legacy path on the exact symptom the epic exists to fix, and point to the slice that closes it.

## 6. `isGoIdent` is not a package-qualifier validator

**MAJOR.**

`modelgen.go:243-246`:

```go
base := pkgBase(path)
if !isGoIdent(base) { base = "pkg" }
```

`isGoIdent` (`resourcegen/names.go:185-203`) checks letters, digits and underscore only. It does not reject reserved words. A path ending in `/map`, `/type`, `/go`, `/func` or `/select` yields an alias that is a keyword, and `renderImports` emits it into the import block:

```go
import map "example.com/x/map"   // does not parse
```

The same hole guards `pkg` at `modelgen.go:89`. Low frequency, granted. But `aliasFor`'s doc comment (`:234-238`) promises *"the path's last segment when that is a Go identifier, else a synthetic name"* — and a keyword is not a usable identifier in that position. The guard does not meet the contract that justifies its existence.

**Suggestion:** add a keyword check (`go/token.IsKeyword`) to `isGoIdent` or to a dedicated `isUsableIdent`, and cover the rejection path — nothing currently tests the `base = "pkg"` branch reaching a rendered file either.

## 7. `ensureValueOnlyPath` uses a promotion-aware lookup where it needs a same-level one

**MAJOR.**

`modelgen.go:152-168`:

```go
for _, seg := range bindNames[:len(bindNames)-1] {
    for t.Kind() == reflect.Pointer { t = t.Elem() }   // dead
    sf, ok := t.FieldByName(seg)
    if sf.Type.Kind() == reflect.Pointer { return err }
    t = sf.Type
}
```

Two problems.

**Dead code.** `sch.ModelType` is already dereferenced by `schema.Parse`, and from the second iteration `t = sf.Type`, which was just proven non-pointer. The inner loop never executes. It defends a state the function itself made impossible, and obscures which check is load-bearing.

**Wrong lookup.** `reflect.Type.FieldByName` does a breadth-first search through embedded fields, including embedded pointers, and returns the promoted field. `sf.Type` is then the leaf's type, not the pointer traversed to reach it — precisely the fact this function exists to detect. The function needs "the field named `seg` declared at this level"; it asks "any field reachable as `seg` from here."

I cannot prove a live hole from this diff: if GORM's `BindNames` is always the exact level-by-level chain, promotion never fires and this is only the wrong tool. That invariant is written down nowhere, and the only test (`modelgen_test.go:216-229`) covers a single-level embed. Please state the invariant in the comment and add a nested-embed case (`A → B → leaf`) plus a pointer at an intermediate level.

**Suggestion:** scan `t.NumField()` for an exact name match at each level instead of `FieldByName`, drop the dead deref loop, and test nesting.

## 8. The wire contract now inherits database column naming

**MAJOR.**

`modelgen.go:320` uses `f.Column` — the database column name — as the JSON field name. The legacy path derives it from the field (`resourcegen/fields.go:135-140`). For a plain field the two coincide; for `gorm:"column:note_b"` they do not, and the API now exposes `note_b`.

ADR-016 says *"HTTP contracts are preserved where possible."* A `column:` override is a persistence decision, and this silently promotes it to a public API decision. ADR-016's own ownership table puts "column name" under **GORM schema owns** and the API surface under **policy owns** — routing the column name into the JSON tag crosses that line.

**Suggestion:** if this is intentional, document it in the ADR or the field-policy grammar and give users a way to name the wire field. If it is incidental, derive the JSON name from `GoName`.

## 9. Minor

- `aliasFor` returns `(string, error)` and never returns an error (`modelgen.go:239-254`), leaving a dead `if err != nil` in `render` (`:227-230`). Residue of the previous fail-closed contract. Drop the error, or make the checks from findings 3 and 6 use it.
- The PR description is stale. It still lists `renderGoType` as one of the three new functions; `git grep renderGoType` at `58b3886` returns nothing — it became `typeRenderer.render`. Both acceptance-criteria boxes are ticked `[x]` under text stating the slice *"advances, and does not yet fully close"* them. Please fix the body; a reviewer should not have to diff the commits to learn what the PR contains.
- The three review threads were resolved by their own author with no reply explaining what changed. `58b3886` lands 12 minutes after the review and addresses all three, but "resolved" without rationale forces the next reviewer to re-derive the mapping. Leave the reasoning in the thread.
- There is no §4 backlog entry for RESGEN-1. `AGENTS.md` makes `docs/GOMBIT_BUILD_PLAN.md` §4 authoritative and ties issues to it 1:1; BENCH-1 and HOST-0 both have entries with AC / deps / size / labels. #352 has none, so there is no in-repo AC to review this PR against — only the issue body. ADR-016 is properly filed and Accepted, so the decision is recorded; the backlog entry is not. Worth reconciling before slice 4, and it is the owner's call, not this PR's job.

---

# VERDICT

The root problem is **architecture**, not implementation. `modelResource` is presented as ADR-016's single authoritative representation and is in fact a lossy, partly guessed projection: it discards the policy facts slice 4 and the request DTO both need (findings 1 and 5), and it derives half its identity from `sch.ModelType` and the other half from string manipulation over an import path that the same file proves cannot yield a package name (finding 2). Every individual defect here has that shape — a real fact reduced to a string, or dropped, before it reaches the code that needs it.

The secondary problem is **tests**. The only compiled artifact in the suite is deliberately import-free (finding 4), and one test actively asserts an uncompilable identifier as correct (finding 3). For a source generator, "it compiles" is the cheapest oracle available, and every shape this engine can emit should be going through it.

Otherwise the working agreement holds: no locked decision is reopened, ADR-016 is implemented as written rather than re-litigated, no M6 battery leaks in, `resourcepolicy` is consumed rather than duplicated, and the Huma/OpenAPI pipeline is untouched (nothing is wired yet, so §5.4 is correctly N/A). Scope discipline is good. The slice is real work.

Do not merge yet. Fix the representation before slice 4 builds a CRUD handler on top of it — carry `CreateSource` and the nullability facts, take the destination package as an explicit validated input, reject unexported and keyword identifiers, and compile every emitted shape. Findings 6-8 can land in the same pass; finding 9 is cleanup.

Findings 1, 2 and 3 are what block the merge; the rest can be argued. If you agree with the direction, I am happy to help with the fixes — particularly carrying `Resolved` through the projection and extending the compile-and-run test to cover the import paths, since those two close most of the list between them.

resourcegen/modelgen.go
```diff
@@ -0,0 +1,476 @@
...
+// says a create source exists, so resourcepolicy is satisfied, but the empty
+// string is not the data the column was declared to require: the same silent
+// zero-fill issue #218 exists to eliminate, reproduced by the DTO built to
+// prevent it (issue #352 review). The legacy field-grammar path emits this
+// constraint (fields.go, humaTags); this is the model-first path catching up,
+// not new scope — see modelField's doc comment for why NotNull survives the
+// projection to make this possible at all.
+func (f modelField) requestTag() string {
+	tag := `json:"` + f.jsonName() + `"`
+	if f.GoType == "string" && f.NotNull {
```
leo-aa88: **BLOCKING.** `f.GoType` is not a type classification — it's the literal Go source text `render()` produced (see `modelField.GoType`'s own doc: "rendered Go type expression, e.g. `string`, `*time.Time`, `types.Decimal`"). Comparing it against the string literal `"string"` only matches the bare predeclared kind.

But this PR's own earlier commit (386bf89 / 58b3886) exists specifically to make defined types render **by identity, not decomposed to their underlying kind**: `type Tags []string` stays `Tags`. The same logic applies uniformly to `type Slug string`, `type Email string`, `type Username string` — ordinary patterns for "a string column that means something." For any of those, `f.GoType` is `"Slug"`, not `"string"`, so this condition is false and no `minLength` is emitted.

Concrete failure:
```go
type Slug string
type Article struct {
    ID   uint  `gorm:"primaryKey"`
    Name Slug  `gorm:"not null"`
}
```
`Name`'s `GoType` renders as `"Slug"`. The generated create body accepts `{"name": ""}`: it satisfies `resourcepolicy` (a create source exists) and satisfies NOT NULL at the DB layer, but it's exactly the silent empty-string zero-fill this PR's commit message says finding 5 "clos[es]... the exact silent zero-fill issue #218 exists to eliminate." It doesn't close it — only for the one spelling of string least likely to appear on a real domain model, since real models are the ones that *bother* to define `Slug`/`Email`/`Username` as distinct types.

Compare with the legacy path this claims to match (`fields.go:604-611`), which keys off `f.Type == FieldString` — the CLI grammar's own *kind* enum, immune to spelling. The model-first path's textual proxy is not actually equivalent, so "matching the legacy field-grammar path" isn't true for named types.

**Fix:** carry the underlying `reflect.Kind` (or a simple `IsStringKind bool`) alongside `GoType` in `modelField`, set from `f.FieldType.Kind() == reflect.String` at `buildModelResource` time, and test that in `requestTag` instead of string-comparing the rendered expression.

resourcegen/modelgen.go
```diff
@@ -0,0 +1,476 @@
...
+	if len(third) > 0 {
+		groups = append(groups, strings.Join(third, "\n"))
+	}
+	return "import (\n" + strings.Join(groups, "\n\n") + "\n)\n\n"
+}
+
+// isStdImport reports whether an import path is a standard-library package: its
+// first path segment carries no dot (a domain), so "database/sql" is std and
+// "github.com/gombit-dev/gombit/types" is not.
+func isStdImport(importPath string) bool {
```
leo-aa88: **MAJOR** (inert today — this engine isn't wired into any command yet — but live the moment slice 4 calls it).

This heuristic is used (via `importSpec.line()`, line 79) to decide when it's safe to emit a bare, unaliased import — which is only safe when the target package's **declared** name is guaranteed to equal the import path's trailing segment. For real stdlib, Go guarantees that. But "first path segment has no dot" is not equivalent to "is stdlib" — it's equivalent to "the module this package lives in has a non-domain-qualified module path," and gombit itself explicitly supports that: `scaffold/options.go`'s `validateModule` rejects only an empty string, backslashes/spaces, and empty/`.`/`..` segments — `gombit new --module myapp` is accepted today, no dot required anywhere.

That's the exact scenario every other branch of this file was hardened against in the prior review round: a package whose declared `package` clause name doesn't match its import path's basename (the whole reason `render()` now qualifies named types "by identity" instead of guessing from `PkgPath`, and the whole reason third-party imports are *always* explicitly aliased regardless of whether the alias would coincidentally equal the basename). `isStdImport` reopens that same hole for a non-dotted module path with an internal package whose folder name doesn't match its `package` clause — legal, unremarkable Go, and unreachable from any test in this PR (every fixture — `time`, `database/sql`, `github.com/gombit-dev/gombit/types`, the two `collidepkg*` fixtures — is either genuine stdlib or dotted).

Concrete failure: app scaffolded with `gombit new --module myapp`, a shared package at `myapp/types` that declares `package apitypes` (folder name ≠ package clause — legal, happens). `isStdImport("myapp/types")` returns `true` (no dot in `"myapp"`); `pkgBase` yields `"types"`; alias assigned is `"types"` (matches basename) → a bare import `import "myapp/types"` is emitted → Go binds that import to the package's own declared name, `apitypes`, not `types` → every reference in the generated file reads `types.Foo`, which does not exist. `go build` fails on a file this engine's own doc comments promise cannot be emitted broken.

**Fix:** there's no existing stdlib-detection helper in this repo to reuse (checked) — either add one backed by Go's actual standard library set, or, simpler and consistent with how third-party imports are already handled here: drop the bare-import special case entirely and always emit an explicit alias for every non-local import. The cosmetic win (`import "time"` vs `import time "time"`) isn't worth reintroducing the one trust assumption this PR spent three commits eliminating everywhere else.

leo-aa88 (commented):
# REQUEST CHANGES

This is a genuinely careful piece of engineering — three follow-up commits in this same PR already closed eight real defects from a prior review pass (identity derivation, alias correctness, pointer-embed rejection at nested levels, policy-fact preservation). The test suite goes as far as compiling and *executing* the generated mappers in a throwaway module, which is more than most codegen PRs bother with. But the fail-closed contract this engine repeatedly claims for itself — "never emit code that will not compile," "matching the legacy field-grammar path" — has two places where it's checked against the wrong thing, and both are demonstrable from the diff, not hypothetical. See inline comments.

# VERDICT

Both findings are the same shape: a piece of this engine trusts a syntactic proxy (a rendered type string, an import-path's dot-count) for a semantic fact (is this really a plain string kind, is this package's declared name really its path basename) that the rest of the PR goes out of its way *not* to trust. The architecture and the bulk of the implementation are sound — the policy/persistence ownership split, the pointer-embed rejection, the identity-preserving type renderer are all correctly reasoned and well tested. These are two narrow gaps in an otherwise-closed contract, not a design problem, and both have small, mechanical fixes (carry a kind flag; drop or correctly source the std/alias special-case). Fix them — finding 1 especially, since it's live in this slice's own golden/tests today, just untested for the exact case it should cover.

leo-aa88 (commented):
LGTM

PR #362, opened 2026-09-16
feat(resourcegen): model-first CRUD handler + hooks engine — RESGEN-1 slice 4 (#352)

leo-aa88:
## Summary

Slice 4 of the ADR-016 model-first epic: the CRUD **handler + hooks** emitter, consuming the `modelResource` slice 3 derives. It renders two files per resource:

- **Generator-owned handler** (`.gen.go`, DO-NOT-EDIT): the Huma input/output wrappers, a `Handler` over GORM, `list`/`get`/`create`, and `Register`. The create path owns the invariant sequence — **build model from request DTO → `BeforeCreate` hook → persist** — so a server-managed column (`gombit:"read,server"`, absent from the request body) is set by the hook, never zero-filled and never requiring an edit to generated plumbing (the #218 fix).
- **Human-owned hooks file** (no banner): a default no-op implementation of a per-resource, **typed** hooks interface `BeforeCreate(ctx, *Model, <t>CreateBody) error`, generated once and then owned by the developer.

`Register` wires the human-owned `Hooks{}`. Operation IDs match the legacy scheme (`list-books`, `get-book`, `create-book`) so swapping handlers in slice 5 preserves the OpenAPI/client contract. Create + read only (no update/delete), matching the legacy generated surface.

Related #352

## Acceptance criteria

Advances epic #352; this slice specifically delivers the hooks mechanism the ADR requires:

- [x] Server-managed columns move off an opt-out list onto an explicit `BeforeCreate` hook (ADR-016 Consequences); the generated handler owns the invariant sequence and calls it.
- [x] Editing generated plumbing is not the customization path — the handler is DO-NOT-EDIT; customization is the human-owned hooks file.
- [x] Output is byte-deterministic (determinism test), the precondition for slice 5's `generate --check`.

## Scope notes

Deferred by the slice plan:
- **Slice 5** — wire this + the DTO engine into `gombit make resource` (replacing the human-owned handler.go, updating goldentest/frontend/routes) and add the `gombit generate --check` drift gate. The hooks file's write-once/never-overwrite behavior lands there.
- **Slice 6** — remove `resourcecheck` + migration guide.

Not in this slice: update/delete operations, and the list query surface (filters/search/sort/aggregate) — this slice's `list` is plain pagination, matching the #218-relevant read+create loop. Pure emitter: nothing calls it yet, so no generated-app output changes (goldentest untouched).

## Validation

- [x] `go test ./resourcegen/` — pass, incl. a compile-AND-execute test that builds the generated DTO+handler+hooks against the real framework (replace directive + `go mod tidy`) and runs create/get/list on SQLite, asserting the `BeforeCreate` hook sets the NOT NULL server column and it round-trips into the response
- [x] `go build ./...`, `go vet ./resourcegen/`, `golangci-lint run ./resourcegen/` — clean

- [x] New behavior has tests — pure emitter + a real compile-and-execute integration test; the executed DB path runs on SQLite (the generated SQL is GORM's, already covered on the PG/MySQL matrix elsewhere)
- [ ] Stable features ship docs and appear in an example app — N/A until wired into `make resource` (slice 5); docs + migration guide land in slice 6
- [x] Extraction preserves contracts — additive; legacy handler generation untouched; operation IDs preserved
- [x] Generators are idempotent/additive, AST-safe, never overwrite user-owned files — this slice only renders bytes; the write-once hooks behavior is slice 5
- [ ] API changes regenerate OpenAPI + TS client — N/A, not wired to any command yet
- [x] No secrets in generated frontend source — N/A (no frontend)
- [x] Scope stays inside its milestone — post-v0.1 epic #352, no M6 battery creep


resourcegen/modelhandler.go
```diff
@@ -0,0 +1,200 @@
...
+	b.WriteString("\t\treturn nil, database.MapLoadError(ctx, err, \"" + singular + " not found\", \"load " + singular + "\")\n")
+	b.WriteString("\t}\n")
+	b.WriteString("\treturn &get" + typ + "Output{Body: contract.Data[" + data + "]{Data: to" + typ + "Data(row)}}, nil\n}\n\n")
+
+	// create: the invariant sequence — build from the request, run the hook, then
+	// persist. bookFromCreateBody sets only request columns; server-managed columns
+	// are the hook's responsibility, so they are never zero-filled silently.
+	b.WriteString("func (h *Handler) create(ctx context.Context, input *create" + typ + "Input) (*create" + typ + "Output, error) {\n")
+	b.WriteString("\trow := " + unexported(typ) + "FromCreateBody(input.Body)\n")
+	b.WriteString("\tif h.Hooks != nil {\n")
```
leo-aa88: **BLOCKING.** This renders as, in the generated `create`:
```go
row := bookFromCreateBody(input.Body)
if h.Hooks != nil {
    if err := h.Hooks.BeforeCreate(ctx, &row, input.Body); err != nil {
        return nil, err
    }
}
if err := h.DB.WithContext(ctx).Create(&row).Error; err != nil { ... }
```
`bookFromCreateBody` (slice 3) deliberately sets only request-source columns, leaving every server-managed column — by design — at its Go zero value, on the stated premise that `BeforeCreate` is what fills it in. This line is the one place that promise is conditional: when `h.Hooks` is nil, the hook is skipped and the zero-filled row is persisted anyway. For a NOT NULL `uint` server column (exactly the `TenantID` example this PR's own tests use), the DB happily accepts `0` — there's no constraint violation to catch it, no error, no panic. That's the precise silent zero-fill issue #218 (and this slice) exists to eliminate, reopened for any caller that constructs `Handler` directly instead of going through `Register`.

`Register` (the only generated wiring path) always sets `Hooks: Hooks{}`, so this branch is dead code on that path — its only live purpose is to make direct construction (`&Handler{DB: db}`, no `Hooks`) degrade silently instead of failing loudly. That's backwards: direct construction is exactly the case that most needs the invariant enforced, since nothing else checks it. This PR's own compile-and-run test constructs `Handler` by hand too (`&Handler{DB: db, Hooks: serverHooks{}}`) but always sets `Hooks`, so nothing here is exercised — CI is green because no test tries to break this, not because it's safe.

**Fix:** fail closed instead of skipping — e.g.
```go
if h.Hooks == nil {
    return nil, contract.WithContext(ctx, contract.Internal("create "+singular+": Handler.Hooks is not set"))
}
if err := h.Hooks.BeforeCreate(ctx, &row, input.Body); err != nil {
    return nil, err
}
```
(or drop the nil-check entirely and let a nil-interface call panic — worse ergonomics, but at least loud). Either way, add a test that builds `&Handler{DB: db}` with no `Hooks` and asserts `create` does not silently succeed with a zero-valued server column.

leo-aa88 (commented):
# REQUEST CHANGES

The architecture is right: a per-resource, compiler-checked hooks interface instead of a callback map, a generator-owned handler that owns the invariant build→hook→persist sequence, and a real compile-and-execute test (`TestGeneratedHandlerCompilesAndRuns`) that runs create/get/list against SQLite through the real framework. That's a good bar. But the one invariant this slice exists to guarantee — "a server-managed column ... is set by the hook, never zero-filled" — has a silent bypass in the generated `create`, and nothing in this PR's test suite exercises the path that hits it. See inline comment.

# VERDICT

The emitted `create` treats `Hooks == nil` as "skip the hook" instead of "this handler is misconfigured." `Register` (the only path this PR generates) always supplies a non-nil `Hooks{}`, so the hole doesn't fire through the sanctioned path — but `Handler`'s fields are exported specifically so it can be constructed directly (this PR's own tests do it: `&Handler{DB: db, Hooks: serverHooks{}}`), and nothing stops `&Handler{DB: db}` with `Hooks` left at its zero value. When that happens, the exact column this slice's running example uses to demonstrate the fix (`TenantID gombit:"read,server"`, a NOT NULL uint) silently inserts as `0` — no error, no panic, a full HTTP 201 with a wrong tenant id in the response. This is the #218 zero-fill bug the whole slice is built to close, reopened by omission rather than by a typo, in the one runtime path (`create`) this PR most wants to get right. CI is green; this isn't a failing-test problem, it's a contract the test suite never tries to break. Fix the guard to fail closed (return an internal error, or drop the nil-check and let a nil-interface call panic loudly) and add a test that constructs `&Handler{DB: db}` with no `Hooks` and asserts `create` does NOT silently succeed with a zero-filled server column.

leo-aa88 (commented):
# APPROVE

Follow-up on the nil-Hooks finding from the previous round (pullrequestreview-5220956799). Fixed correctly, and better than my suggested placement: the generated `create` now checks `h.Hooks == nil` and returns `contract.Internal(...)` *before* even building the row from the request body, not after — no wasted work on the failure path, and no window where a partially-built row exists without a hook having run.

Coverage is at both levels this codebase already uses for this file:
- `TestRenderModelHandlerStructure` (fast, always runs) now asserts the guard exists and precedes the row-build (`iGuard < iBuild`), and that the emitted error text is exactly right.
- `TestCreateFailsClosedWithoutHooks` asserts the real end-to-end behavior against a live SQLite DB: `&Handler{DB: db}` with `Hooks` left nil, `create` returns an error, and — the part that actually matters — zero rows exist afterward. I initially misread where this test lives (inside the `handlerRunTest` string constant, compiled into the temp module by `TestGeneratedHandlerCompilesAndRuns`) as a placement mistake, since `go test -run TestCreateFailsClosedWithoutHooks ./resourcegen/...` reports "no tests to run." It isn't a mistake: `Handler`, `createBookInput`, and `bookCreateBody` aren't real symbols in the `resourcegen` package, only in the generated output, so this test *has* to live where the pre-existing `TestGeneratedCRUD` already lives, for the same reason — I checked, and `TestGeneratedCRUD` isn't independently `-run`-able either. Consistent with the file's established pattern, not a regression in it.

**Verified independently**: checked out `9f2e2e38f7e4df241fbbe0f8a34499014d544121` in a worktree, ran `go build ./...`, `go vet ./resourcegen/...`, and `go test ./resourcegen/... ./resourcepolicy/...` with no `-short` (so `TestGeneratedHandlerCompilesAndRuns` and the nested nil-Hooks test actually execute, not just the fast structural ones) — all clean, all pass. Confirmed CI's own `Test` job (`ci.yml`) also runs `go test ./...` without `-short`, so this protection is live there too, not just locally.

# VERDICT

The finding is closed with a real fix and a real end-to-end regression test, not a surface patch. Good to merge.

PR #367, opened 2026-09-16
ci(bench): warn when the microbench snapshot predates the code it measures (#293)

### LINKED ISSUE #293: [PERF-14] Warn in CI when the committed microbench snapshot predates the code it measures
## Summary

Nothing in CI notices when the committed benchmark snapshot falls behind the code it measures. `benchmark-report-drift` verifies `README ≡ f(snapshot)` — it regenerates the Performance block and fails on drift — but nothing verifies `snapshot ≡ f(main)`. The gate does exactly what it was designed to do; the gap is in the design, not the implementation.

That gap is how #266 happened: the framework-tax table advertised numbers measured at `88442ae`, before PERF-1..4 (#239, #240, #241, #242) and PERF-11 (#269) merged, and it sat there un-flagged through five performance PRs because no check was looking.

This is the deferred **AC 3 of #266** ("(Optional) staleness warning in CI"), split out so #266 could stay on one axis.

## Why this is cheap now

Before per-unit provenance (#292), there was nothing precise to compare: `metadata.json` carried one `git_commit` for three measurement groups produced hours and days apart, so a staleness check could only have been a coarse guess about the snapshot as a whole.

#292 adds `metadata.json`'s `groups` key, where each measurement unit records the commit, host and Go toolchain **its own** rows were produced at. For the framework-tax table the units are the four published stacks, so a staleness check can now be exact:

```
groups.microbench.<stack>.git_commit   vs   the last commit touching the code that table measures
  for each published stack: nethttp, gin, huma, gombit
```

## Proposed fix

Add a step to the existing **`benchmark-report-drift`** job (`.github/workflows/ci.yml`). It already has the Go toolchain, needs no Docker, and runs on every PR.

Compare each published stack's `groups.microbench.<stack>.git_commit` against the last commit touching the paths whose changes move those numbers. The `gombit-ablation` unit is not published in the README and is out of scope:

- `framework/` — the measured runtime
- `benchmarks/micro/` — the harness itself
- `benchmarks/internal/microbench/` — the parser/schema

When the snapshot commit is an ancestor of that commit, emit a GitHub `::warning::` naming the stack, both commits and the one-line remedy (`make benchmark-micro benchmark-report`, a few minutes, no Docker).

### Hard constraint: warning, never a gate

`docs/plans/BENCH-1-benchmark-suite.md` §3 lists as a non-goal:

> | Add hard perf-regression gates to normal PR CI | Shared runners are too noisy; smoke-only per issue §11 |

This must **exit 0 always**. It is a staleness signal about *provenance*, not a comparison of measured values, so it does not re-open that non-goal — but the implementation must not drift into one. Do not compare numbers; compare commits.

### Implementation note (verified)

`benchmark-report-drift` currently checks out with `actions/checkout@v7` at the default depth of 1, so `git log -1 -- <paths>` finds nothing there. The step needs `fetch-depth: 0`. There is precedent in the same file — the `benchmark-scripts` job already does this (`ci.yml:235`).

## Acceptance criteria

- [ ] A PR where any published stack's `groups.microbench.<stack>.git_commit` predates the last commit touching `framework/`, `benchmarks/micro/`, or `benchmarks/internal/microbench/` surfaces a CI warning naming both commits and the remedy.
- [ ] The step exits 0 in every case, including when the snapshot is stale — it is a warning, never a gate (BENCH-1 §3 non-goal above).
- [ ] A snapshot at or after that commit produces no warning.
- [ ] A snapshot with no entry for a published stack (pre-`groups`, or a stack never run) degrades honestly rather than crashing or reporting a false "fresh".

## Scope notes

- **`crud` and `footprint` are deliberately out of scope.** Their staleness is a known, accepted cost: refreshing them is hours of Docker, not a few minutes of Go, so a warning on every PR would be noise nobody can act on. The check is worth having precisely where the fix is cheap. Extending it later is a separate decision.
- **Different axis from #291.** That issue is "the publishability guard never judges the **host**" — a snapshot taken on a developer laptop publishes banner-free. This one is "nothing notices the snapshot is older than the **code**". Neither subsumes the other; both are gaps in the same guard surface.
- Depends on #292 (per-unit provenance) being merged, since the check reads `groups.microbench.<stack>.git_commit`.

leo-aa88:
## Summary

`benchmark-report-drift` verifies `README ≡ f(snapshot)` — but nothing verifies `snapshot ≡ f(main)`. That gap is how #266 happened: the framework-tax table advertised numbers measured before PERF-1..4 and PERF-11 merged and sat un-flagged through five perf PRs. This adds a **CI warning** (never a gate) that fires when the committed microbench snapshot is older than the code it measures.

- New `benchmarks/scripts/microbench-staleness.sh`, run in the existing `benchmark-report-drift` job. When the commit that last refreshed `benchmarks/results/latest/` is a strict ancestor of the last commit touching `framework/`, `benchmarks/micro/`, or `benchmarks/internal/microbench/`, it emits a `::warning::` naming both commits and the remedy (`make benchmark-micro benchmark-report`, ~40s, no Docker).
- **Warning, never a gate** (BENCH-1 §3 non-goal): it exits `0` in every case, including when stale, and compares *commits*, never measured values.
- `benchmark-report-drift`'s checkout gains `fetch-depth: 0` (the `git log -1 -- <paths>` lookups find nothing under the default depth-1 clone — the implementation note in #293; mirrors the `benchmark-scripts` job).
- `crud`/`footprint` are out of scope: refreshing them is hours of Docker, so a per-PR warning would be noise nobody can act on. The microbench group refreshes in ~40s of Go.

Closes #293. Depends on #292 (per-group provenance) — merged.

## Divergence from the issue's literal mechanism (please sanity-check)

#293 specifies comparing `metadata.json`'s `groups.microbench.gombit.git_commit` (the commit the bench ran at, added by #292) against the last code commit. **This repo squash-merges**, which discards a PR's branch commits — so the recorded provenance commit is almost always *orphaned from `main`* (I verified the current snapshot's `e9a98f7` is unreachable). A literal provenance-commit ancestry check would therefore hit #293's own "not in fetched history → degrade honestly" path on essentially **every** PR, making the warning inert.

To make it actually fire, the script keeps the identical *ancestor* semantics but keys them on the commit that last refreshed the snapshot files (`git log -1 -- benchmarks/results/latest/`), which **is** in `main`'s history and is squash-robust. The recorded `groups.microbench.gombit.git_commit` is still read and surfaced in the warning ("measured at …") when present. If you'd rather stay literal to the provenance-commit approach (and accept it's inert under squash-merge), say so and I'll switch it.

Real-repo behavior today (the snapshot *is* stale — #268/#270/#271 landed after the last refresh):

```
::warning::microbench snapshot is stale: last refreshed at 57d58dd… (microbench data measured at e9a98f7…), but framework benchmarks/micro benchmarks/internal/microbench last changed at ab315b4…. Refresh with: make benchmark-micro benchmark-report (~40s, no Docker).
```

## Acceptance criteria

- [x] A PR whose snapshot predates the last commit touching `framework/` / `benchmarks/micro/` / `benchmarks/internal/microbench/` surfaces a CI warning naming both commits and the remedy — `microbench-staleness_test.sh` case 1 (stale) plus the real-repo run above.
- [x] The step exits 0 in every case, including when stale — every branch ends in `exit 0`; asserted in all five test cases.
- [x] A snapshot at or after that commit produces no warning — test cases 2 (refreshed after) and 3 (same commit).
- [x] A snapshot with no `groups.microbench` entry degrades honestly (no crash, no false "fresh") — the decision no longer depends on that field (it only enriches the message); case 4 covers a snapshot with no usable provenance, and a totally absent metadata file is handled too.

## Scope notes

- Different axis from #291 (host-provenance guard); neither subsumes the other.
- The warning will fire on the next PRs until the snapshot is refreshed on the reference box and the milestone (#264) is closed — that's intended, persistent visibility.

## Validation

- [x] `bash benchmarks/scripts/microbench-staleness_test.sh` — PASS (5 cases)
- [x] `bash benchmarks/scripts/microbench-staleness.sh` against the real repo — warns + exit 0 (output above)
- [x] `.github/workflows/ci.yml` parses as valid YAML
- [ ] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run` — N/A (no Go changed; shell + YAML + docs only)
- [ ] Project `code-review` skill — pending

- [x] New behavior has tests — `microbench-staleness_test.sh`, wired into the `benchmark-scripts` CI job; no DB-touching change
- [x] Stable features ship docs and appear in an example app — documented in `benchmarks/docs/methodology.md`; N/A for an example app (CI-only tooling)
- [x] Extraction preserves contracts — N/A (new tooling)
- [ ] Generators — N/A
- [ ] API changes regenerate OpenAPI + TS client — N/A (no API change)
- [x] Scope stays inside milestone — post-v0.1 bench tooling; no M6 battery


leo-aa88 (commented):
# APPROVE

Small, well-scoped, and honest about its own limits. A few things worth noting, none blocking:

- The hard constraint ("never gate, always exit 0") actually holds under adversarial reading, not just by the trailing `exit 0`s: `set -uo pipefail` without `-e` means no command failure aborts the script, and every variable that could be undefined is assigned via `"$(...)"` (always defined, possibly empty) before use, so there's no `-u` unbound-variable trap waiting to fire either. The `command -v python3` guard means the enrichment clause degrades cleanly when python3 is absent rather than erroring.
- The divergence from #293's literal spec (keying off the snapshot directory's last-touched commit instead of the recorded `groups.microbench.gombit.git_commit`, because squash-merge orphans that commit from `main`) is the right call and is argued honestly in the PR body, not smuggled in — and the recorded provenance commit is still surfaced in the message, so nothing is lost, just not used as the comparison key.
- Verified independently: checked out `2a9b29780ccef9c1783f53ae7b77135c827d14a3` in a worktree, ran `microbench-staleness_test.sh` (5 cases, PASS), ran the script against the real repo (correctly warns, exit 0, matches the PR's quoted output), and confirmed the new step lands in the intended `benchmark-report-drift` job (not misplaced) and the workflow still parses as valid YAML.

One thing to watch, not a blocker: the ancestor-only comparison (`snapshot_commit` strict ancestor of `code_commit` ⇒ stale, anything else ⇒ fresh) relies on the checked-out history being effectively linear from HEAD, which squash-merge-to-main gives you — worth keeping in mind if this repo ever grows a rebase-merge or long-lived-branch workflow where two commits on the same file set could be genuinely unordered rather than one always dominating the other. Not a real risk today; just the assumption underpinning the two-way branch (`is-ancestor` in one direction, "else fresh" in the other) instead of checking both directions explicitly.

Good to merge.

PR #369, opened 2026-09-16
feat(resourcegen): model-first list-query parity — RESGEN-1 slice 4.5 (#352)

leo-aa88:
## Summary

Slice **4.5** (inserted before 5b, per review direction): port the legacy generated handler's declared **list-query surface** — `filterable` / `sortable` / `searchable` / `aggregatable` (#260/#272) — to the model-first handler, so the 5b `make resource` switch is a pure ownership/derivation change, **not a feature downgrade**. Parity only: same wire semantics, same spelling, no new operators/ranges.

- **resourcepolicy**: the `gombit` tag gains the four query tokens (API-policy facts, like `read`/`write`); `Resolved` carries `Filterable`/`Sortable`/`Searchable`/`Aggregatable`. **New rule: a queryable field must be response-visible** — a hidden or write-only column with a query surface leaks through result membership, counts, and ordering, so `gombit:"-,searchable"` (or `write,filterable`) fails resolution.
- **resourcegen**: `buildModelResource` validates **type-appropriateness** where the Go type is known (filterable ⊂ string/integer/bool; searchable ⊂ string; aggregatable ⊂ numeric incl. `types.Decimal`; sortable unrestricted) — the split the legacy path made from its field grammar, kept type-correct. `renderModelHandler` **consumes the resolved facts** (never re-reads tags) to emit the `?<field>=` / `?search=` / `?ordering=` / `?aggregate=` surface + `ListMeta`, byte-identical to the legacy output for the same field config; with no query capabilities it's unchanged (`PageMeta`).

Example:
```go
type Book struct {
    gorm.Model
    Title string `gombit:"read,write,searchable,sortable"`
    Genre string `gombit:"read,write,filterable,sortable"`
    Price int64  `gombit:"read,write,filterable,aggregatable"`
}
```

Related #352

## Acceptance criteria

- [x] The model-first handler reaches parity with the legacy handler's declared query surface (filter/sort/search/aggregate), so 5b introduces no regression.
- [x] Query capability is one derivation: declared on the model, resolved by `resourcepolicy`, consumed by the handler — never re-inspected.
- [x] Hidden/write-only fields cannot expose a covert query surface (response-visible rule, fail closed).

## Scope notes

Parity, not extension: no operators, ranges, grouping, or arbitrary filtering. **Known gap:** a string-`enum` filter carries no `enum:"..."` constraint on its query param — the GORM schema stores an enum as a plain varchar, so the allowed values aren't recoverable from the model (bool filters still get `true/false` from the kind). Legacy path (files.go/fields.go) untouched.

## Validation

- [x] `go test ./resourcegen/ ./resourcepolicy/ ./generate/` — pass, incl. a compile-AND-execute test running filter/search/ordering/aggregate (`sum:price`=30) through the generated list handler on SQLite against the real framework
- [x] `go build ./...`, `go vet`, `golangci-lint run` on the three packages — clean

- [x] New behavior has tests — resolver + type rules + handler structure + a real query compile-and-execute
- [ ] Stable features ship docs and appear in an example app — N/A until make resource emits the layout (5b)
- [x] Extraction preserves contracts — the legacy handler/query codegen is untouched; the model-first path reproduces its wire semantics
- [x] Generators are idempotent/additive, AST-safe — pure emitter; deterministic
- [ ] API changes regenerate OpenAPI + TS client — N/A (not wired to a command yet)


resourcegen/modelgen.go
```diff
@@ -154,18 +155,72 @@ func buildModelResource(model any, pkg string) (modelResource, error) {
...
+	if f.Searchable && f.Kind != reflect.String {
+		return fmt.Errorf("resourcegen: column %q (%s) is not searchable; a searchable column must be a string", f.Column, f.GoType)
+	}
+	if f.Aggregatable && !isAggregatableType(f.Kind, ft) {
+		return fmt.Errorf("resourcegen: column %q (%s) is not aggregatable; an aggregatable column must be numeric (integer, float, or decimal)", f.Column, f.GoType)
+	}
+	return nil
+}
+
+func isFilterableKind(k reflect.Kind) bool {
```
leo-aa88, line 232: **BLOCKING — confirmed by reproduction.** `isFilterableKind` (and the parallel checks in `validateQueryTypes` and `isAggregatableType` below) switches on `f.Kind` directly. For a nullable column (`*string`, `*int64`, `sql.NullString`, …), `f.Kind` is `reflect.Pointer` (or `reflect.Struct` for the `sql.Null*` wrappers) — not in any of these `case` lists — so the column is rejected as "not filterable"/"not searchable"/"not aggregatable" outright, even though it's a perfectly normal string/int/bool column that merely happens to be nullable.

This is a genuine parity regression, not a pre-existing legacy limitation: the legacy field-grammar path keys these rules off its own `FieldType` enum (`fields.go` `typeAllowsFilter`/`typeAllowsSearch`/`typeAllowsAggregate`), which is orthogonal to nullability — a `price:int?` column there is just as filterable/aggregatable as `price:int`. This PR's own commit message says the goal is parity so the slice-5b switch is "not a feature downgrade," but for any nullable column it is one.

I reproduced it directly against `buildModelResource`:
```go
type m struct {
    ID   uint    `gorm:"primaryKey"`
    Name *string `gombit:"read,write,filterable"`
}
// -> resourcegen: column "name" (*string) is not filterable; a filterable column must be a string, integer, or bool

type m struct {
    ID    uint   `gorm:"primaryKey"`
    Price *int64 `gombit:"read,write,aggregatable"`
}
// -> resourcegen: column "price" (*int64) is not aggregatable; an aggregatable column must be numeric (integer, float, or decimal)
```
Both rejected. As a control, `*types.Decimal` with `aggregatable` **is** accepted — because `isAggregatableType`'s decimal branch (line 211 below) explicitly dereferences pointers before comparing to `decimalType`:
```go
for ft.Kind() == reflect.Pointer {
    ft = ft.Elem()
}
return ft == decimalType
```
That dereference only feeds the decimal comparison — the numeric-kind `case` above it (`reflect.Int`, `reflect.Float64`, …) runs on the original, still-wrapped `k` and never gets a second chance after unwrapping. So `*types.Decimal` works but `*int64`/`*float64` don't, which is itself an inconsistency independent of the filterable/searchable gap.

**Fix:** unwrap a pointer (and, if you want full nullable-type parity, `sql.Null*`) once, in one place, and have `isFilterableKind`, the numeric branch of `isAggregatableType`, and `filterKindExpr` (`modelhandler.go:295`) all consult the unwrapped kind — `filterKindExpr` has the identical blind spot (its `default` case would currently mis-map a nullable int to `database.FilterString` once the validation gate above stops blocking it, so it needs the same fix to avoid trading one bug for another).

leo-aa88 (commented):
# REQUEST CHANGES

The policy layer is careful — the response-visible rule for query capabilities (`gombit:"-,searchable"` fails resolution) is exactly right, and the decimal-is-aggregatable-despite-being-a-struct-kind case is handled correctly. But the type-appropriateness gate this slice adds checks `reflect.Kind` directly against the field's own (possibly pointer) type, never unwrapping a nullable column first — so it rejects filterable/searchable/aggregatable on any pointer-typed (nullable) column, even though the legacy grammar this slice explicitly claims parity with does not draw that distinction. Confirmed by direct reproduction, not just reading the code — see inline comment. That's a real regression against this PR's own stated acceptance criterion ("so the 5b make resource switch is ... not a feature downgrade"), for an unremarkable, everyday case (an optional/nullable string or numeric column).

# VERDICT

The fix is narrow and consistent with how the PR already treats `types.Decimal` — that comparison unwraps a pointer before checking against `decimalType`; `isFilterableKind`, the numeric-kind branch of `isAggregatableType`, and `filterKindExpr` (`resourcegen/modelhandler.go`) all need the same treatment, applied through one shared helper rather than three separately-drifting kind switches. Once that's in, add coverage for a nullable string/int/bool alongside the existing non-nullable cases — the current tests only ever construct non-pointer fields, which is exactly why this shipped clean through CI.

leo-aa88 (commented):
# APPROVE

Fixed correctly, and more thoroughly than I asked for. `effectiveKind(t reflect.Type)` unwraps pointers *and* maps every `database/sql` nullable wrapper (`sql.NullString`, `NullInt64/32/16`, `NullByte`, `NullBool`, `NullFloat64`) to the scalar kind it carries, and `modelField.Kind` is now always this effective kind — so `isFilterableKind`, `isAggregatableType`'s numeric branch, and `filterKindExpr` are all fixed together by construction instead of needing three separate patches. That's a better fix than the one I suggested (a single source of truth instead of three call sites each doing their own unwrapping).

I re-ran my exact original reproduction (`*string filterable`, `*int64 aggregatable`) against this commit — both now accepted — plus `sql.NullString` and a sanity check that a genuinely bad type (`string` marked `aggregatable`) is still rejected. All as expected. The new `TestBuildModelResourceAllowsNullableQueryTypes` covers the same ground and goes further: it specifically asserts `filterKindExpr` maps a nullable `*int64` to `database.FilterInt64` and a nullable `*bool` to `database.FilterBool`, not the wrong string fallback — exactly the secondary issue I flagged as latent in `filterKindExpr` once the upstream gate was fixed.

**Verified independently**: checked out `8aaefe75c3bcfb4ec38143a01d38eb51e7204402` in a worktree, ran `go build ./...`, `go vet`, and the full `resourcegen`/`resourcepolicy` suites — clean — then re-ran my original repro plus a `sql.NullString` case and a negative control, all passing as expected.

# VERDICT

Closed with a real fix, and the shared-helper approach is more maintainable than a patch at each call site would have been. Good to merge.

Issue #370, opened 2026-09-16
[bug] run-crud raw k6 summaries omit the benchmark, so a second workload overwrites crud-list's raw evidence


Brunotlps:
## Problem

`run-crud` writes each trial's raw k6 summary to `raw/<framework>_c<n>_t<n>.json` (`benchmarks/scripts/run-crud/main.go`, `summaryPath`). The name carries no benchmark.

Since #361, rows and provenance are keyed on `(framework, benchmark)`, so a second workload for an app keeps that app's `crud-list` rows. Its raw summaries do not survive: running `gombit` under any other workload overwrites `raw/gombit_c100_t1.json` and the rest with that workload's output.

`benchmarks/results/README.md` states why `raw/` is committed:

> `latest/` **is committed** (including `raw/`, the per-trial k6 summaries, so a later parser fix can reconstruct the numbers)

After such a run, the published `crud-list` rows no longer have the raw evidence that promise refers to, and nothing warns.

## Why it is latent today

`crud-list` is the only workload, so framework and `(framework, benchmark)` still select the same files. It becomes reachable with the second workload BENCH-1 Phase 5 plans (`make benchmark-auth`, `make benchmark-techempower`), the same trigger as #361.

## Expected

A run writes and replaces exactly its own raw summaries: the file name includes the benchmark, e.g. `raw/<framework>_<benchmark>_c<n>_t<n>.json`.

## Note for whoever fixes it

The 150 committed files under `results/latest/raw/` use the old name, so the fix either renames them in the same change or reads both names. Nothing in the repo parses `raw/` today (only `run-crud` writes it; `benchmarks.yml` uploads `results/dispatch/raw/` as an artifact).

## Acceptance criteria

- [ ] Raw summary file names include the benchmark; a test runs two workloads for one framework and asserts both sets of raw files survive.
- [ ] The committed `results/latest/raw/` snapshot matches the new naming (or the reader handles both), and `benchmarks/results/README.md` still describes it accurately.

Found while fixing #361. Pre-existing; not introduced there.

PR #371, opened 2026-09-16
fix(bench): key run-crud rows and provenance on (framework, benchmark) (#361)

### LINKED ISSUE #361: [bug] run-crud mergeRows drops a framework's rows for every other workload
## Problem

`mergeRows` in `benchmarks/scripts/run-crud/main.go` replaces a framework's previous rows by dropping **every** existing row for that framework, whatever its workload:

```go
// mergeRows drops any existing rows for framework (a re-run replaces them) and
// appends the new ones; rows for other frameworks are kept.
func mergeRows(existing, newRows []result.Result, framework string) []result.Result {
	merged := make([]result.Result, 0, len(existing)+len(newRows))
	for _, r := range existing {
		if r.Framework != framework {
			merged = append(merged, r)
		}
	}
	return append(merged, newRows...)
}
```

`result.Result` carries a `Benchmark` field, but the merge ignores it. A run of one workload for an app therefore silently deletes that app's rows for **every other workload** in the same `results.json`. Nothing warns; the rows, each potentially hours of containerized measurement, are simply gone.

## Reproduction

Seeded a `results.json` with three rows, then ran `run-crud`'s real `run()` (k6 injected through the existing `k6Runner` seam, as `TestRunWritesAndAccumulatesAcrossFrameworks` does) for `framework: gombit`:

| before | after |
| --- | --- |
| `gombit` / `crud-list` | `gombit` / `crud-list` (replaced — correct) |
| `gombit` / `auth-jwt` | **deleted** |
| `gin-gorm` / `techempower-json` | kept (different framework) |

## Impact, in both directions

- **Re-running `crud-list` deletes other workloads.** As reproduced above: a later `make benchmark-crud-all` (or `APPS=gombit …`) erases the affected apps' rows for any other workload. The README CRUD table is unaffected in this direction (its `crud-list` rows are the ones rewritten); the loss lands in `results.json` / `results.csv` and in `summary.md`, which renders one table per benchmark.
- **Recording a second workload through this merge deletes `crud-list`.** If another workload's rows are written through `mergeRows` for an app that already has `crud-list` rows, those are dropped, and `make benchmark-report` then renders the README CRUD table without that app.

## Why it is latent today

- `run-crud` is the only writer of `results.json`, and it hard-codes `Benchmark: "crud-list"` in `run`, so today framework and (framework, benchmark) select the same rows.
- It becomes reachable as soon as a second workload writes to the same file, which the benchmark plan requires: `docs/plans/BENCH-1-benchmark-suite.md` Phase 5 plans a Gombit-only auth-overhead benchmark (`make benchmark-auth`) and a TechEmpower-inspired subset across all six apps (`make benchmark-techempower`), with the AC *"each `make benchmark-*` target runs independently and writes rows to `results.json`"*. `Makefile:3` notes *"-auth and -techempower land with their phases"*.
- The rest of the pipeline already handles multiple workloads: `internal/summary` groups by `(framework, benchmark, concurrency)`, and PR #185 fixed `summary.md` to render one table per benchmark with two workloads. The merge is the one piece still keyed as if only one workload existed.
- No open issue currently schedules those workloads (#141 is closed), so there is no fixed date for the trigger. It could equally be fixed as part of whichever change adds the second workload; this issue exists so that change does not inherit the bug unknowingly.

## Related: `-workload` is accepted but not reflected in the rows

`run-crud` accepts an undocumented `-workload` flag (not used by the `Makefile`, `run-crud-all.sh`, the docs or `benchmarks.yml`). Whatever script it names is mounted as `/workload/crud-list.js` and its rows are labelled `crud-list`. Measuring a different workload that way would therefore replace the README CRUD table's rows with another workload's numbers. Same spot, same underlying cause: the row's `benchmark` should come from the workload that ran, not a constant.

## Expected

A run replaces exactly the rows it produced: the merge key is `(framework, benchmark)`, and `benchmark` identifies the workload that actually ran.

## Note for whoever fixes it

Since #292, `metadata.json` records provenance per **unit**, and each group's unit must equal its data file's merge key (`groups.crud.<framework>` today, matching `mergeRows`). Changing the merge key to `(framework, benchmark)` must change the CRUD provenance unit with it, in `scripts/run-crud` (the `StampUnit` call) and in `internal/report` (the units `writeCRUDTable` publishes); otherwise recording one workload would relabel another workload's rows. #292 hit exactly this mismatch once, with footprint's variant.

## Acceptance criteria

- [ ] `mergeRows` keys on `(framework, benchmark)`; a test seeds two workloads for one framework, re-runs one, and asserts the other survives.
- [ ] A row's `benchmark` reflects the workload that ran (or `-workload` is removed), with a test.
- [ ] The CRUD provenance unit equals the new merge key, with a test that recording one workload does not change another workload's recorded provenance.

Found while auditing #292 (issue #266). Pre-existing; not introduced there.

Brunotlps:
## Summary

`run-crud` replaced a framework's rows by dropping **every** existing row for that framework, whatever its workload, and labelled every row `crud-list` whatever `-workload` named. A run of one workload therefore silently deleted the app's rows for every other workload in `results.json`. It is latent while `crud-list` is the only workload; BENCH-1 Phase 5 plans more that write to the same file.

The provenance #292 (#266) introduced carried the same single key: `run-crud` stamped `groups.crud.<framework>`, and the report captioned the CRUD table from bare frameworks. Recording a second workload would also have relabelled `crud-list`'s commit and host. It is the defect ad595a0 fixed for footprint's variant, one group over.

- **Merge key is `(framework, benchmark)`.** `mergeRows` keeps a row unless both match. The benchmark comes from the workload script that ran (`workloads/<benchmark>.js`) instead of a constant. An empty benchmark fails `run()` before anything is written, and `:` is rejected because it separates the two halves of the unit.
- **One unit, one derivation.** `result.ProvenanceUnit` derives `framework:benchmark` for both the writer (`scripts/run-crud`) and the reader (`internal/report`), mirroring `footprint.ProvenanceUnit`.
- **Snapshot migration.** The six committed CRUD units move from `<framework>` to `<framework>:crud-list` with byte-identical provenance, so the README `## Performance` block is unchanged.
- **`-workload` says what it now controls.** Its help states that the file name becomes the recorded benchmark, and the script mounts as `/workload/script.js` rather than a name that claimed every workload was `crud-list`.
- **Shared protocol is guarded.** `metadata.json` records `concurrency`, `trials`, `duration_seconds`, `warmup_seconds` and `benchmark_tool` once per snapshot, so `run-crud` refuses a run whose values differ from the recorded ones while rows it does not replace remain, and modifies nothing. Addresses the review's blocking finding (its second option). Incoming parameters are validated before measuring and compared as values (an incoming zero is never "unstated"). Producers do not lock `OUT_DIR`, and concurrent writers to one `OUT_DIR` are unsupported.

Closes #361

## Acceptance criteria

- [x] **`mergeRows` keys on `(framework, benchmark)`; a test seeds two workloads for one framework, re-runs one, and asserts the other survives.** `TestMergeRowsReplacesOnlyTheSameFrameworkAndBenchmark` covers same app/other workload, other app/same workload, and both. `TestRunReplacesOnlyItsOwnWorkload` is the issue's reproduction through the real `run()`, with k6 injected.
- [x] **A row's `benchmark` reflects the workload that ran, with a test.** Kept `-workload` and derived the benchmark from it (`TestBenchmarkNameComesFromTheWorkloadScript`). `TestRunReplacesOnlyItsOwnWorkload` then runs a non-default workload and asserts its rows carry its own benchmark and add to the snapshot. `TestRunWithoutBenchmarkFailsAndWritesNothing` covers the guard.
- [x] **The CRUD provenance unit equals the new merge key, with a test that recording one workload does not change another workload's recorded provenance.** Writer: `TestRunReplacesOnlyItsOwnWorkload` seeds `gombit:auth-jwt` provenance, records `crud-list`, and asserts `auth-jwt`'s entry is unchanged and nothing is filed under the bare framework. Reader: `TestSecondWorkloadCannotRelabelTheCRUDTable` mirrors `TestEmbeddedVariantCannotRelabelTheContainerRow`.

Each new test was checked against the fix with one half reverted at a time. Merge keyed on the framework alone, the constant `crud-list` label, stamping the bare framework, the report captioning by framework, and dropping the empty-benchmark guard each fail at least one test. Reverting the snapshot migration fails no Go test but fails `make benchmark-report-check` (`README.md is out of date`), which CI's `benchmark-report-drift` job runs.

## Scope notes

- **Raw k6 summaries still use the old key.** `raw/<framework>_c<n>_t<n>.json` has no benchmark, so a second workload would overwrite `crud-list`'s per-trial summaries even though its rows now survive, and `results/README.md` keeps `raw/` so a parser fix can reconstruct the numbers. Fixing it renames the 150 committed raw files, so it is left out of this surgical fix and tracked in #370.
- **Pre-existing local snapshots.** An `OUT_DIR` written after #292 and before this PR records `groups.crud.<framework>`. Its CRUD rows caption as unrecorded until re-run, and the old keys stay as unpublished orphans. That fails honestly (no row is attributed to a commit it did not run at), and the committed snapshot is migrated here.
- `scripts/k6load` (footprint's under-load run) still mounts its script as `/workload/crud-list.js`. It writes no rows, so it is untouched.
- **The protocol guard also covers the framework axis.** `APPS=gombit make benchmark-crud-all TRIALS=1 DURATION=1s` over a full snapshot already kept the other apps' rows and rewrote the protocol they are described with. That gap predates this PR and is closed by the same check.
- **UX cost of the guard.** A populated snapshot cannot move to a new protocol one app at a time, because each run is refused while the other apps' rows remain. Use a separate `OUT_DIR` or start the snapshot over. That includes the documented `make benchmark-crud-all CONCURRENCY=1,10,100 ...` example when run over a populated `results/latest`; the error says so.
- **`make benchmark-metadata` is not guarded.** It also writes the protocol fields and measures nothing. Recording the protocol per unit, so workloads with different pins can share a snapshot, and guarding `collect-host-info` are tracked in #377.
- **Concurrent writers are unsupported.** No producer locks `OUT_DIR` (`run-crud`, `StampUnitFile`, `collect-host-info`). `run-crud` reads the snapshot once and rechecks it before writing, which catches a producer that finished during the sweep but is not synchronization.
- **A corrupt `metadata.json` now fails before measuring.** The pre-flight check reads it even when `results.json` has no rows, instead of failing after the sweep.
- No workload is added. `docs/plans/BENCH-1-benchmark-suite.md`'s historical "merges by framework key" log entries are left as written.
- **CHANGELOG N/A:** the benchmark harness has no CHANGELOG entries (BENCH-1, #289).

## Validation

- [x] `gofmt -l benchmarks/`: clean
- [x] `go build ./...`: OK
- [x] `go test ./...`: pass
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run`: 0 issues
- [x] `make benchmark-report-check`: README matches the data
- [x] Mutation-checked: guard removed, too wide, too narrow, write-time check removed, symmetric zero wildcard restored, each single input validation dropped, and `Recorded()` ignored each fail at least one test
- [x] `LC_ALL=C bash benchmarks/scripts/{run-crud-all,run-suite,footprint-all,benchmark-target}_test.sh`: OK
- [x] Project `code-review` skill run against this diff. It returned COMMENT with no blocking finding: the raw-summary key (in Scope notes), the undocumented `-workload` contract and its misleading mount path (fixed), orphaned units in older local snapshots (in Scope notes), and a README line over 80 columns (fixed).

- [x] New behavior has tests; DB-touching changes pass the SQLite + PostgreSQL + MySQL matrix: 4 new Go tests for the merge key and 11 for the protocol guard and input validation (in `run-crud` and `metadata`), the framework-only merge test rewritten for the new key, and updated unit literals in `run-crud`, `report` and `collect-host-info`. DB matrix N/A: no persistence code touched.
- [x] Stable features ship docs and appear in an example app: `methodology.md` documents the CRUD unit; `benchmarks/README.md` describes the merge. Example app N/A: benchmark harness, not a framework feature.
- [x] Extraction preserves contracts — refactor and move, don't rewrite code that already passes its tests: the merge, stamping and caption code keep their shape; only the key changes.
- [ ] Generators are idempotent/additive, AST-safe, and never overwrite user-owned files (if this PR touches generators): N/A, no generator touched.
- [ ] API changes regenerate the OpenAPI doc + TS client in this PR: N/A, no API change.
- [ ] No secrets in generated frontend source; `VITE_*` is treated as public: N/A, no frontend file touched.
- [x] Scope stays inside this issue's milestone — no M6 "battery" creep (jobs, events, scheduler, mail, storage, gRPC, multi-tenancy, i18n): only `benchmarks/` changes; no workflow, ADR or locked decision touched.


benchmarks/scripts/run-crud/main.go
```diff
@@ -175,16 +183,18 @@ func run(cfg runConfig, k6run k6Runner) error {
...
-	// supported run, so stamping the whole crud group here would caption every
-	// other app's rows with a commit they never ran at (issue #266, round 2).
-	meta = metadata.StampUnit(meta, metadata.GroupCRUD, cfg.framework, meta.Provenance())
+	// This run's own provenance, filed under this (app, workload) alone. run-crud
+	// replaces exactly those rows and preserves the others, and APPS= subsetting
+	// is a supported run, so stamping the whole crud group here would caption
+	// every other app's rows with a commit they never ran at (issue #266, round
+	// 2), and stamping the app alone would do the same to its other workloads
+	// (#361).
+	meta = metadata.StampUnit(meta, metadata.GroupCRUD, result.ProvenanceUnit(cfg.framework, cfg.benchmark), meta.Provenance())
```
leo-aa88, line 192: **BLOCKING.** This only makes provenance unit-scoped. `metadata.Collect` above still puts `Concurrency`, `Trials`, `DurationSeconds`, `WarmupSeconds`, and `BenchmarkTool` in snapshot-global fields, and `metadata.Merge` keeps the incoming values wholesale. Concrete failure: run `crud-list` with `100 VUs / 5 x 30s`, then run `auth-jwt` with `1 VU / 1 x 1s`. Both row sets and both provenance entries now survive, but `metadata.json` says `1 VU / 1 x 1s`; `report.Render` uses those fields for the status banner and **How these were measured** while publishing the old `crud-list` table. The report therefore attributes one workload's protocol to another. The new test runs both workloads with the same config, so the tests prove only the non-adversarial case. Make the protocol/tool metadata follow the same unit boundary and derive the report from the rendered units, or explicitly reject merging incompatible parameters before writing anything.

leo-aa88 (changes requested):
# REQUEST CHANGES

The `(framework, benchmark)` row key and per-unit provenance fix the deletion/relabeling bug the issue names. The snapshot contract is still incomplete, however: the second workload now survives while its run rewrites shared measurement metadata used to describe the first workload.

## 1. Multi-workload rows still share last-writer-wins protocol metadata

**BLOCKING.**

See the inline comment on `benchmarks/scripts/run-crud/main.go`. The patch makes workload identity per unit for rows and provenance but leaves protocol/tool metadata snapshot-global. A second workload with different flags causes the generated README to describe `crud-list` with that second workload's parameters. The new end-to-end test uses identical flags for both workloads, so it cannot expose the mismatch.

# VERDICT

This is an architecture/contract mismatch, not a merge-key implementation error. The new unit boundary is `(framework, benchmark)`, but not all metadata that gives that unit meaning follows the boundary. Either make measurement metadata unit-scoped (and have the report consume the rendered units' metadata) or reject a merge whose shared parameters differ, with an adversarial test. Do not merge while one workload can silently rewrite how another workload is reported.

benchmarks/scripts/run-crud/main.go
Brunotlps: ## Response to the blocking finding

Confirmed, and the scenario reproduces: `metadata.Merge` keeps `concurrency`, `trials`, `duration_seconds`, `warmup_seconds` and `benchmark_tool` from the incoming run, and `report.writeMethodology` / `reducedFrom` publish them for every row.

I took the second option you offered: `run-crud` now refuses the merge when those parameters differ from the recorded ones and rows it does not replace remain, and modifies nothing (5e37994).

## What changes

- `metadata.RunParams` / `ConflictsWith`: two records conflict only when both state a value and the values differ. A zero is "not recorded".
- `runConfig.runParams()` feeds both `metadata.Collect` and the guard, so what is compared is what is written.
- The check runs at the start of `run()`, so a doomed run fails before the hours of measurement, and again in `writeOutputs` before anything is written, because the snapshot can change during the sweep.
- A fresh `OUT_DIR`, a snapshot that records no parameters, and a run that replaces every row in the file are unaffected.

## Tests

- Your scenario: `crud-list` at 100 VUs / 5 x 30s, then `auth-jwt` at 1 VU / 1 x 1s. It is refused and `results.json` / `metadata.json` stay byte-identical.
- The positive control: the same second workload under the recorded protocol succeeds.
- The framework axis, the run-replaces-everything case, and a snapshot that changes mid-sweep.
- Each test was checked against the guard removed, too wide, too narrow, and with the write-time check removed.

## Two things I want to be explicit about

1. The framework-axis gap predates this PR: `APPS=gombit make benchmark-crud-all TRIALS=1 DURATION=1s` over a full snapshot already kept the other apps' rows and rewrote the protocol they are described with. The same check closes it.
2. There is a UX cost. A populated snapshot cannot move to a new protocol one app at a time, because each run is refused while the other apps' rows remain. Use a separate `OUT_DIR` or start the snapshot over. That includes the documented `make benchmark-crud-all CONCURRENCY=1,10,100 ...` example when run over a populated `results/latest`. The error message says so.

## Out of scope here

- `make benchmark-metadata` (`collect-host-info`) also writes these fields and is not guarded. The `RunParams` comment and `methodology.md` say so.
- Recording the protocol per unit, so workloads with different pins can share a snapshot, changes the metadata schema, the report and the README, so I did not fold it into this PR. It is tracked in #377, including how the per-unit record should be shaped.

If you would rather have the per-unit option in this PR instead of the guard, say so and I will rework it.

benchmarks/internal/metadata/metadata.go
```diff
@@ -453,6 +454,105 @@ func Merge(existing, incoming Metadata) Metadata {
...
+// than claiming zero. Only a disagreement between two stated values means the
+// merged snapshot would misdescribe rows.
+func (recorded RunParams) ConflictsWith(incoming RunParams) []string {
+	var diffs []string
+	if len(recorded.Concurrency) > 0 && len(incoming.Concurrency) > 0 &&
+		!equalInts(recorded.Concurrency, incoming.Concurrency) {
+		diffs = append(diffs, fmt.Sprintf("concurrency %s -> %s",
+			intList(recorded.Concurrency), intList(incoming.Concurrency)))
+	}
+	if recorded.Trials != 0 && incoming.Trials != 0 && recorded.Trials != incoming.Trials {
```
leo-aa88: **BLOCKING.** `0` cannot mean "unstated" on the incoming side while `Merge` is last-writer-wins. Concrete CLI path: seed another unit's rows under `trials: 5`, then run this command for a different `(framework, benchmark)` with `-trials 0` and otherwise identical parameters. `ConflictsWith` ignores `5 -> 0`; `run()` performs the warm-ups, produces no measured rows, and `metadata.Merge` overwrites the snapshot-wide `Trials` with `0` while preserving the seeded rows. The report now says those five-trial rows used zero trials—the exact relabeling this guard claims to make impossible. The same representation error applies to every incoming zero/empty field, including a deliberately zero warm-up: absence and a real zero value are collapsed, yet the incoming zero is still written. Make missingness directional or presence-aware and reject incomplete incoming run parameters, or preserve the recorded value when incoming is genuinely unspecified. Add an end-to-end test that proves a zero-valued incoming run cannot rewrite the protocol of preserved rows.

benchmarks/scripts/run-crud/main.go
```diff
@@ -234,17 +264,20 @@ func parseSummaryFile(path string) (k6.Summary, error) {
...
-func writeOutputs(outDir, framework string, newRows []result.Result, meta metadata.Metadata) error {
-	rows, err := mergedResults(filepath.Join(outDir, "results.json"), newRows, framework)
+// truncating it: re-running one framework's workload replaces those rows,
+// while running each framework (or workload) in turn accumulates them all.
+// metadata's version maps are unioned the same way so a multi-framework
+// snapshot records every implementation that contributed.
+func writeOutputs(outDir, framework, benchmark string, newRows []result.Result, meta metadata.Metadata) error {
+	// Read before writing anything, so an unreadable snapshot, or a merge the
+	// guard refuses, fails the run with the snapshot untouched.
+	if err := mergeable(outDir, framework, benchmark, meta.RunParams()); err != nil {
```
leo-aa88: **BLOCKING.** This is a TOCTOU check, not concurrent-writer protection. After it returns, `writeOutputs` separately rereads results and metadata and then truncates three files with `os.Create`; no lock spans check → read → merge → write. Starting from an empty `OUT_DIR`, two runs with different protocols can both pass this check and both read empty state, then interleave so `results.json` comes from one run while `metadata.json` comes from the other (and `results.csv` can disagree with both); a less unlucky schedule still silently loses one run. `TestWriteIsRefusedWhenTheSnapshotChangesDuringTheSweep` mutates the snapshot *before* this check, so it proves only a serial recheck, not the concurrent `make` case claimed at lines 317–320. Hold a snapshot lock across the second check, all merge reads, and a staged commit—or explicitly declare concurrent writers unsupported and remove the guarantee.

leo-aa88 (changes requested):
# REQUEST CHANGES

The `(framework, benchmark)` merge key and provenance migration satisfy the issue's three acceptance criteria. The protocol guard added after the previous review still does not uphold its own invariant, though: it has a representation hole for zero-valued incoming parameters and a synchronization hole for the concurrent-writer case it explicitly claims to handle.

## 1. Zero-value wildcard lets an incoming run rewrite shared protocol metadata

**BLOCKING.**

See the inline comment on `metadata.RunParams.ConflictsWith`. Missingness is treated symmetrically, but `metadata.Merge` is not symmetric: incoming top-level values always win. A real `-trials 0` run therefore preserves another unit's rows while replacing their recorded trial count with zero.

## 2. The write-time recheck is TOCTOU, not concurrency control

**BLOCKING.**

See the inline comment in `writeOutputs`. The patch promises protection from another producer in the same `OUT_DIR`, but the check, merge reads, and three truncating writes have no shared lock or transaction. The new test only changes the files before the second check; it never exercises a writer after the check has passed.

# VERDICT

The original key/provenance implementation is sound, but the remedial guard still has an invariant/architecture defect and missing adversarial tests. CI is green. This is not a failing-test problem; the current tests encode the zero-value hole and serialize the alleged concurrent writer before the unprotected commit window.

Do not merge until incoming parameter presence is represented consistently with merge behavior and the concurrent-write claim is either implemented with synchronization or removed from the contract.

benchmarks/internal/metadata/metadata.go
Brunotlps: Agreed: missingness was symmetric while `Merge` is not. Fixed in be9b62a, in two layers.

**Incoming parameters are never incomplete.** `run()` now rejects, before measuring anything, `-trials` below 1, an empty or non-positive concurrency level, a duration that is not positive, a negative warm-up, and an empty `-k6-image` (`validateRunParams`). `durationSeconds` is only reached with input that passed that check, so an unparseable string can no longer become 0.

**The comparison is directional.** Every incoming field is a value, because `Merge` writes it: `ConflictsWith` now compares all five exactly, zero included. Only the recorded side can be absent, and only as a whole (`RunParams.Recorded()`): a snapshot that states none of the five has no protocol to misdescribe. Once it states any of them, all five are compared. A partial record, such as `make benchmark-metadata` given only `-trials`, conflicts on its zeros, which fails closed.

A deliberate zero-second warm-up is accepted and compared as a value: it matches a snapshot recorded with no warm-up and conflicts with one recorded with a warm-up.

Tests:

- `TestZeroTrialRunCannotRewriteThePreservedRowsProtocol`: your case end to end through `run()`. Another unit's rows at `trials: 5`, then `-trials 0` with otherwise identical parameters. Refused before k6 starts, and `results.json` / `results.csv` / `metadata.json` stay byte-identical. It also covers the side effect where the old path deleted the run's own unit rows.
- `TestRunRejectsIncompleteOrInvalidRunParameters`: one case per rejected input, each asserting nothing is written.
- `TestZeroWarmupIsComparedAsAValue`, plus `TestRunParamsConflictsWith` / `TestRunParamsRecorded` for the predicate.

Each layer is checked on its own: restoring the symmetric wildcard, dropping any single validation, or ignoring `Recorded()` each fails at least one test. The zero-trial end-to-end test fails only when both layers are removed, since either one alone refuses that run.

benchmarks/scripts/run-crud/main.go
Brunotlps: Agreed: that was a serial recheck described as concurrency control. I took the second option and removed the guarantee from the contract, in be9b62a.

- `writeOutputs` now reads the snapshot once, and the check and the merge both use that same in-memory data (`readSnapshot` → `refuseIncompatible` → `mergeRows` / `metadata.Merge`). The re-reads between check and merge are gone.
- The comments on `run()`, `refuseIncompatible` and `writeOutputs`, `methodology.md` and `benchmarks/README.md` now say what the write-time check is: it catches a snapshot another producer rewrote while this run was measuring. It is not concurrency control. No producer locks `OUT_DIR`, a write that overlaps this run's own read-merge-write can still interleave with it, and running two producers against one `OUT_DIR` at the same time is documented as unsupported.
- The test is renamed to what it proves, `TestWriteRecheckCatchesASnapshotRewrittenDuringTheSweep`, and no longer calls the other producer concurrent.

I did not add a lock because it would only mean something if every producer honored it. `metadata.StampUnitFile` (footprint, microbench) and `collect-host-info` also read-modify-write `metadata.json` with `os.Create` and no lock, so a lock in `run-crud` alone would still not synchronize the snapshot. That gap predates this PR. I can track a cross-producer lock and atomic writes separately if you want it.

Reading the snapshot once is a property of the code structure. I did not find a way to observe it in a test without adding a lock or an injection hook, so the test covers the sequential rewrite and not an interleaving.

leo-aa88 (approved):
LGTM

PR #372, opened 2026-09-17
feat(resourcegen)!: make resource emits the model-first layout — RESGEN-1 slice 5b (#352)

leo-aa88:
## Summary

`gombit make resource` now emits the **model-first, generator-owned** shape (ADR-016) instead of a human-owned handler, and does so as a **preflighted, atomic** operation:

1. **Plan** (`resourcegen.Plan`): compute the human-owned scaffold — `model.go` (with `gombit:"..."` policy translated from the CLI query modifiers), the `.gombit-resource` marker, the AST-wired `AutoMigrate`/`Register`, and the frontend — **without writing**. All static checks (app layout, name, HTTP-path collision, the enum gap, the legacy-layout guard) run here.
2. **Phase-2 preflight** (`generate.PlanResource`): render the generator-owned `dto.gen.go`/`handler.gen.go` + seed `hooks.go` from the *pending* model, compiled through a **Go build overlay** (`go run -overlay`) run from the real module root — so the model is actually compiled and validated while the real tree is untouched and local `replace` directives keep their meaning. Nothing is written.
3. **Commit**: apply the merged plan (scaffold + generated files) as **one transaction** (`fsTx` rolls back every write on any filesystem failure), only after both phases validate. The migration is generated as an explicit **post-commit** step. `--dry-run` prints the exact merged plan and writes nothing.

`renderModel` emits `gombit:"..."` tags from the CLI modifiers (`filterable`/`sortable`/`searchable`/`aggregatable` → `read,write,<caps>`, since a query capability must be response-visible); a plain field stays untagged (read+write default); a `belongs_to` FK keeps its filterable-by-default.

Related #352

## Feature downgrade: enum

This is **not** a pure derivation change — `enum(...)` fields are **dropped**. Their values are not recoverable from the GORM schema (an enum is stored as a plain varchar), so the model-first generator cannot yet derive the DTO/OpenAPI constraint. `make resource` **rejects** enum fields (fail loud, never silently degrade to string) and the help/`docs/cli.md`/`docs/contract.md` no longer advertise enum as supported. Use a `string` for now; a model-first enum policy is a planned follow-up. Every other contract — scalar, relation, and the full list-query filter/sort/search/aggregate surface (4.5) — is preserved.

## Create-DTO validation (schema-derivable only)

Restored from facts the model unambiguously states — never re-parsed from the CLI grammar: `maxLength` from a column's `Size` (only when `> 0`), `minimum:"0"` for unsigned ints, alongside the existing `minLength:"1"` for NOT NULL strings. Enum value constraints are not emitted (the class-B gap above).

## Legacy handler removal (necessary here)

Dropping `handler.go`/`routes.go` from `make resource` orphans the legacy `renderHandler`/`renderRoutes` + their query-only `Field` helpers — an unused-lint failure and dead code — so they're removed here. Their unit tests are superseded by the model-first handler tests (`modelhandler_test.go`). The **separate `resourcecheck` package + `*_drift_test.go` removal is slice 6**.

## Migration guide (shipped here)

Because this is a breaking change to the generated layout, `docs/migration-model-first-resources.md` documents converting a legacy hand-customized resource to hooks, in the **correct, executable order** (extract customization → mark + remove the legacy plumbing → `gombit generate` → move logic into the *seeded* `hooks.go`, once the create-body type exists). `--force` is **not** offered as a migration shortcut (it does not remove the legacy handler). `goldentest.TestMigrateLegacyResourceToModelFirst` walks a legacy resource through the documented steps and compiles the result.

## Acceptance criteria (ADR-016 / #352)

- [x] `gombit make resource` changes shape: generator-owned `*.gen.go` + a human-owned hooks file, instead of a human-owned handler.
- [x] One authoritative derivation: both `make resource` and `gombit generate` produce the `*.gen.go` through the same `RenderResource` (Program Mode).
- [x] Full list-query parity (4.5) + schema-derivable create validation; enum is an explicit, documented downgrade.
- [x] The operation is preflighted and atomic — no half-migration; `--dry-run` reflects the real plan.

## Scope notes

- **Deferred to a small follow-up (5b-ii):** the CI `gombit generate --check` drift-gate job (now there's an in-repo app layout to gate — the goldentest demo).
- **Slice 6:** remove the `resourcecheck` package + generated drift tests.

## Validation

- [x] `go test ./...` — pass (0 failures). The goldentest compile step runs `gombit generate` in a temp copy (local `replace`) then builds, proving make resource + generate + build works end to end across scalar / relation / list-query / aggregate resources; new tests cover the legacy-layout refusal, the transactional rollback, the overlay-backed dry-run, and the end-to-end migration path.
- [x] `go build ./...`, `go vet`, `golangci-lint run` on the changed packages — clean (the 2 `createsuperuser.go` gosec G115 are pre-existing and unflagged by CI).

- [x] New behavior has tests; the model-first output compiles + runs (goldentest + modelhandler compile-and-run)
- [x] Stable features ship docs and appear in an example app — CLI/contract docs updated; migration guide shipped; the goldentest demo is the in-repo example
- [x] Extraction preserves contracts — HTTP contracts preserved (operation IDs, query surface, DTO validation); this is a `!` breaking change to the generated **source layout/customization model** (and a documented enum downgrade), per ADR-016
- [x] Generators are idempotent/additive, AST-safe, never overwrite user-owned files — the marker + model are idempotent; `.gen.go` are generator-owned; `hooks.go` is seed-once
- [ ] API changes regenerate OpenAPI + TS client — N/A (the generated HTTP surface is unchanged; OpenAPI is emitted by Huma at runtime as before)
- [x] No secrets in generated frontend source — unchanged frontend


resourcegen/files.go
```diff
@@ -52,10 +52,14 @@ func unexported(name string) string {
 }
 
 func renderFeatureFiles(ctx renderContext) ([]fileSpec, error) {
+	// Model-first (ADR-016): make resource scaffolds the human-owned model and marks
```
leo-aa88, line 59: **BLOCKING.** The comment says `model` is human-owned, but `renderModel` still calls `goBanner()`, whose value is `Code generated by gombit make resource. DO NOT EDIT.` `checkOverwrite` also uses that banner to classify the file as generated, and the updated tests literally call the model generator-owned. ADR-016 requires the model + policy to be the human-edited source of truth. What the fuck is the ownership boundary supposed to mean if the authoritative file tells the developer not to edit it? Emit the model with a human-owned/scaffold marker (not the generated banner), preserve it on regeneration, and add a test that edits the model then regenerates DTO/handler output without `--force`.

resourcegen/modelgen.go
```diff
@@ -454,32 +460,50 @@ func (f modelField) responseTag() string {
...
+//	minLength:"1"    a NOT NULL string — else "" satisfies resourcepolicy (a create
+//	                 source exists) but is not the data the column requires (#218).
+//	maxLength:"<N>"  a string whose column has a real size (Size > 0); an
+//	                 unset/driver-dependent size asserts nothing.
+//	minimum:"0"      an unsigned integer column.
+//
+// The string checks key off f.Kind (reflect.String), not the GoType text: a
+// defined `type Slug string` renders as "Slug" yet is Kind reflect.String and
+// still zero-fills to "", so it needs the same constraint. Enum value constraints
+// are NOT emitted — the GORM schema stores an enum as a plain varchar, so the
```
leo-aa88, line 476: **BLOCKING.** Calling this a known gap does not make it compatible with the PR's `no feature downgrade` contract. The legacy path emitted `enum:"draft,published"`; this patch renders `enum(draft,published)` into a plain string model field with only `gorm:"size:32"`, so Program Mode can recover only `maxLength`. The generated create DTO accepts and persists any short string, and OpenAPI loses the enum, while the generated React form still hard-codes the two values. That is three representations disagreeing again—the exact architecture ADR-016 exists to remove. Enum values are API policy, so preserve them in authoritative model policy and derive Huma/OpenAPI/frontend from there, or reject enum fields until that representation exists. Restore an adversarial API/OpenAPI test; compilation cannot catch this.

cli/make.go
```diff
@@ -121,6 +122,8 @@ internal/platform (not only the new resource). Otherwise the GORM model is
 still loader-ready for gombit db makemigrations.`,
```
leo-aa88, line 124: **BLOCKING.** This line switches a stable public command to a breaking source layout, but the same PR leaves `docs/cli.md` promising `handler.go` + `routes.go`, `docs/tutorial.md` printing and editing those files, and `examples/tutorial/README.md` documenting them. ADR-016 explicitly requires a migration path and upgrade/CHANGELOG entry because existing handler customizations must move to hooks. The working agreement says stable features ship docs and an example; the PR checklist itself marks that item incomplete. Documentation for behavior that no longer exists cannot be deferred after the command changes. Ship the migration and update the CLI/tutorial/example with this switch, or keep the old command behavior until that slice is ready.

cli/make.go
```diff
@@ -135,6 +138,22 @@ still loader-ready for gombit db makemigrations.`,
 			if err != nil {
 				return fmt.Errorf("gombit make resource: %w", err)
 			}
+			if dryRun {
+				// Nothing was written, so there is no compilable model to generate
+				// from; tell the developer the second phase that a real run performs.
+				_, _ = fmt.Fprintln(stdout, "dry-run: skipping gombit generate (would produce the *.gen.go DTOs/handler and seed hooks.go)")
```
leo-aa88: **MAJOR.** `--dry-run` is documented as `print files that would be written without writing`, but this skips the entire second phase and prints only a wildcard sentence. It does not list the exact `dto.gen.go`, `handler.gen.go`, and hooks paths, and it can report success even when the real Program Mode loader would fail. The current test asserts only phase-one files, so it blesses the mismatch. Stage phase one in a temporary app tree and run the same `generate.Generate(..., DryRun: true)` path there, or otherwise build one complete plan before mutating anything; then assert dry-run lists every real artifact and surfaces generation errors.

leo-aa88 (commented):
# REQUEST CHANGES

This patch does not satisfy the ownership boundary or the no-feature-downgrade claim it is merging. The generated source of truth still tells users not to edit it, enum constraints disappear between the CLI grammar and Huma, dry-run skips half the command, and the stable documentation still describes the deleted layout.

## 1. The human-owned source of truth is emitted as DO NOT EDIT

**BLOCKING.**

See the inline comment on `resourcegen/files.go`. ADR-016 says the model is human-owned and authoritative. The emitted model still carries `Code generated by gombit make resource. DO NOT EDIT.`, and the updated tests explicitly call it generator-owned. That is the opposite ownership contract.

## 2. Enum values vanish before Huma/OpenAPI

**BLOCKING.**

See the inline comment on `resourcegen/modelgen.go`. `status:enum(draft,published)` now becomes only `string` plus `gorm:"size:32"`; the derived request DTO has `maxLength:"32"` but no `enum` tag. A direct API client can submit any short string and GORM persists it, while `docs/cli.md` still promises API-layer validation. CI is green because the old enum-tag assertion was deleted and the replacement end-to-end test only compiles generated code.

## 3. The breaking CLI ships against documentation for the deleted layout

**BLOCKING.**

See the inline comment on `cli/make.go`. The working agreement requires stable behavior to ship with docs and an example. `docs/cli.md`, `docs/tutorial.md`, and `examples/tutorial/README.md` still tell users that `make resource` creates human-owned `handler.go`/`routes.go` and show customization there. ADR-016 also requires a real migration path and upgrade/CHANGELOG entry for this break. Deferring all of that leaves the released command and its documentation mutually exclusive.

## 4. `--dry-run` skips the operation it claims to preview

**MAJOR.**

See the second inline comment on `cli/make.go`. Help promises to print the files that would be written, but the implementation omits `dto.gen.go`, `handler.gen.go`, and the exact hooks path and never verifies that Program Mode generation succeeds. The dry-run test only checks phase-one files, so an implementation that cannot complete a real run still passes.

# VERDICT

The deepest problem is an incomplete architecture transition: the patch switches the public command to model-first output before the authoritative model, API constraints, dry-run behavior, tests, and documentation have crossed the same boundary. These are contract and specification defects, not cleanup for a later slice. Do not merge until the emitted ownership is truthful, enum semantics are represented model-first (or enum resources are rejected), the complete command is previewed, and the breaking migration/docs ship with the behavior.

cli/make.go
```diff
@@ -135,6 +138,22 @@ still loader-ready for gombit db makemigrations.`,
...
+			if dryRun {
+				// resourcegen.Generate already previewed the scaffold plus the exact
+				// files gombit generate would produce; nothing was written, so there is
+				// no compilable model to actually run generation against here.
+				return nil
+			}
+			// Phase 2: derive the generator-owned *.gen.go (DTOs, mappers, CRUD
+			// handler) and seed the human-owned hooks file from the model just
+			// written — the same Program-Mode path gombit generate uses.
+			if err := generate.Generate(cmd.Context(), generate.Options{
```
leo-aa88: **BLOCKING:** This validation runs only after phase one has committed the marker/model/AST/frontend writes. Reproduced with the base branch's legacy `Book`: phase one creates `.gombit-resource`, this call rejects `handler.go`, and the marker remains, causing every later `gombit generate` to fail on that package. `--dry-run` skips this call and falsely reports the phase-two artifacts. Preflight both phases on a scratch/overlay tree and apply only after the complete plan succeeds; use that same plan for dry-run.

resourcegen/generate.go
```diff
@@ -41,6 +41,16 @@ func Generate(ctx context.Context, opts Options) error {
...
 		return err
 	}
+	// Enum values are API policy the GORM schema cannot carry (it stores an enum as
+	// a plain varchar), so the model-first generator cannot yet derive the enum
+	// constraint for the DTO/OpenAPI. Rather than silently drop it — three
+	// representations disagreeing, the exact thing ADR-016 removes — reject enum
+	// fields until a model-first enum policy exists (a later slice).
+	for _, f := range fields {
+		if f.Type == FieldEnum {
+			return fmt.Errorf("resourcegen: field %q is an enum, which the model-first generator does not support yet (enum values are not recoverable from the GORM schema); use a string field for now — a model-first enum policy is planned", f.JSONName)
```
leo-aa88: **BLOCKING:** Rejecting enum is safer than silently dropping its values, but the public command contract still advertises it. `gombit make resource --help` lists `enum`, includes enum examples, and promises validation; `docs/cli.md` later calls it supported/API-validated; `docs/contract.md` still documents enum resource commands; and the PR body says there is no feature downgrade. A documented invocation now dies here. Either carry enum policy through the model-first representation or remove/update every contradictory help, reference, example, and PR claim.

docs/migration-model-first-resources.md
```diff
@@ -0,0 +1,95 @@
...
+
+   ```sh
+   gombit generate
+   ```
+
+   This writes `dto.gen.go` + `handler.gen.go` and seeds `hooks.go` (if absent).
+   Review the diff, wire your customization into `hooks.go`, and commit the
+   `*.gen.go` — a per-PR `gombit generate --check` keeps them fresh.
+
+Prefer to start clean? `gombit make resource Book <fields> --force` re-scaffolds
```
leo-aa88: **BLOCKING:** This fallback cannot migrate the legacy resource the guide is about. `--force` overwrites the model, but phase one does not remove `handler.go`/`routes.go`; phase two then rejects the legacy handler. The result is a clobbered model and no generated replacement. Remove this advice or implement a preflighted migration path that handles the legacy files safely, and test the documented sequence end to end.

leo-aa88 (commented):
# REQUEST CHANGES

HEAD fixes the original model-ownership bug and correctly refuses to silently erase enum constraints. It is still not mergeable. The two-phase command mutates authoritative state before phase two validates it, the public enum contract contradicts the runtime, and the advertised migration escape hatch is destructive and nonfunctional.

## 1. Phase one commits state before phase two can reject it

**BLOCKING.**

`resourcegen.Generate` writes the model/marker/AST/frontend plan before `generate.Generate` checks the legacy layout. I reproduced this against the base branch's generated `Book` resource: `make resource Book title:string:required` prints `create internal/book/.gombit-resource`, keeps the old model, and then exits with `uses the legacy handler layout`. The marker remains beside `handler.go` and `routes.go`; subsequent application-wide `gombit generate` discovers it and refuses to regenerate anything.

The dry-run follows a different path: it skips phase two and prints a hard-coded artifact list, so it reports files that the real command deterministically cannot produce. That is not fail-closed behavior. It is a half-migration plus a lying preview. Preflight both phases against a scratch/overlay tree, derive the preview from that same complete plan, and apply writes only after the whole operation validates.

## 2. The command rejects enums while its public contract advertises them

**BLOCKING.**

The runtime now rejects every `FieldEnum`, which is a defensible temporary policy. But `gombit make resource --help` still lists `enum` as supported, includes enum examples, and says values are validated; the same `docs/cli.md` section later lists enum as supported and API-validated, `docs/contract.md` still uses enum resource examples, and the PR body claims this is a pure derivation change with no feature downgrade. A user copying the documented command gets an immediate error.

Either represent enum policy in the model-first source of truth or remove enum from every command/help/reference/example and state the downgrade accurately in the PR description. One paragraph saying enums are rejected does not repair contradictory normative documentation ten lines later.

## 3. The documented legacy migration fallback cannot succeed

**BLOCKING.**

The guide tells legacy users that `make resource ... --force` re-scaffolds and regenerates everything. It does not remove the legacy `handler.go` or `routes.go`; it overwrites the human-edited model first, then phase two rejects the still-present legacy handler. The literal numbered procedure is also backwards: it shows a `hooks.go` using `bookCreateBody` before generation has created that type, but Program Mode must compile the package in order to generate it.

This breaking change requires an executable, tested migration sequence. Generate the DTO/hook types before asking users to reference them, preserve the old handler until customization is extracted, and do not advertise `--force` as a migration command unless it actually handles the legacy layout safely.

CI is green. This is not a failing-test problem; the tests do not exercise a legacy `make resource` invocation or execute the migration guide.

# VERDICT

The remaining defect is architectural error handling, reinforced by incorrect specification and migration documentation. Phase ownership is undefined: phase one changes discovery state, phase two can reject that state, and dry-run substitutes a second hand-maintained description instead of validating the operation.

Do not merge HEAD. Make the operation preflighted/atomic, align every enum contract with runtime behavior, and prove the migration path end to end.

resourcegen/fstx.go
```diff
@@ -0,0 +1,96 @@
...
+	prior, err := os.ReadFile(path) // #nosec G304 -- generator output path under the app work dir
+	existed := err == nil
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	created, err := mkdirAllTracked(filepath.Dir(path))
+	if err != nil {
+		return err
+	}
+	if err := os.WriteFile(path, content, 0o644); err != nil { //nolint:gosec // generated source is a non-secret artifact
```
leo-aa88, line 38: **BLOCKING:** `os.WriteFile` truncates an existing file before writing, but this transaction records the step only after the call succeeds. A disk-full/short-write error can leave this target truncated (or leave a partial new file), and the caller's rollback has no journal entry with which to restore/remove it. The test fails in `MkdirAll`, before touching the target, so it misses this invariant. Write to a sibling temp file and rename only after a complete close/sync, with the prior file/mode journaled for later rollback.

resourcegen/generate.go
```diff
@@ -105,34 +186,168 @@ func Generate(ctx context.Context, opts Options) error {
...
 		}
-		if opts.DryRun || !item.write {
+		p := filepath.Join(dir, e.Name())
+		if p == modelAbsPath {
 			continue
 		}
-		full := filepath.Join(opts.WorkDir, filepath.FromSlash(item.relPath))
-		if err := os.MkdirAll(filepath.Dir(full), 0o750); err != nil {
-			return fmt.Errorf("resourcegen: mkdir %s: %w", filepath.Dir(full), err)
+		overlay[p] = stub
```
leo-aa88, line 255: **BLOCKING:** This stubs *every* non-model Go file, not merely stale `*.gen.go`. That hides `hooks.go`, services, repositories, and custom files during preflight even though `Apply` preserves them. Example: a hook references `row.Title`; `--force` changes the model to `Name`; preflight sees only this stub and succeeds, then commit keeps the real hook and leaves the package uncompilable (and returns success when Atlas is absent). Render the new artifacts, then compile an overlay of the complete final tree including preserved human code.

generate/generate.go
```diff
@@ -325,8 +346,17 @@ func loadArtifacts(ctx context.Context, opts Options, absWorkDir, module string,
...
+	// Stage any pending (not-yet-written) sources as a Go build overlay so the
+	// loader compiles the resource as it will be committed, without touching the
+	// real tree. Backing files live under the same temp dir and vanish with it.
+	overlayArg, err := writeOverlay(tmpDir, overlay)
+	if err != nil {
+		return nil, err
+	}
 	var stdout bytes.Buffer
-	goArgs := []string{"run", "-mod=mod", "./" + filepath.ToSlash(loaderRel)}
+	goArgs := []string{"run", "-mod=mod"}
```
leo-aa88, line 357: **BLOCKING:** `-mod=mod` is a mutating mode: `go run` may update the real app's `go.mod`/`go.sum`. This executes before `plan.Apply` and during `--dry-run`, so those changes are outside the transaction and violate the documented no-write contract. The test's pre-emptive `go mod tidy` just avoids observing the bug. It also breaks Go workspace mode, where explicit `-mod=mod` is rejected. Use readonly/vendor/workspace-compatible resolution or stage module metadata in scratch space.

docs/migration-model-first-resources.md
```diff
@@ -0,0 +1,103 @@
...
+   in step 3):
+
+   ```go
+   func (Hooks) BeforeCreate(ctx context.Context, row *Book, body bookCreateBody) error {
+       row.TenantID = tenantFromContext(ctx)
+       return nil
+   }
+   ```
+
+   If you customized more than create (custom list filtering, extra endpoints),
```
leo-aa88, line 92: **BLOCKING:** Keeping a file does not preserve this behavior. Generated `Register` always binds its own unexported list/get/create methods; there is no list hook/service injection point, and extra endpoints are not registered merely by living in the package. Custom list filtering cannot replace the generated route at all. Either say such resources must remain legacy until a supported opt-out/extension point exists, or document and test real registration/service wiring. The migration test's empty `Handler` does not exercise this claim.

leo-aa88 (commented):
# REQUEST CHANGES

The previous marker/enum/migration-order defects are fixed, and every CI job is green on `fa1888b`. The replacement architecture still does not satisfy its defining claims: the transaction can corrupt the file whose write fails, force-mode preflight deliberately ignores preserved human code, and dry-run can modify module files before the transaction exists. The migration guide also advertises a customization path the generated dispatcher cannot execute.

## 1. The rollback journal excludes the write that actually fails

**BLOCKING.**

`fsTx.write` reads the prior bytes, calls `os.WriteFile`, and appends the journal step only after `os.WriteFile` succeeds. But `os.WriteFile` opens an existing file with truncation before writing. A disk-full, quota, NFS, or short-write error can therefore leave an existing file truncated—or a new file partially created—while returning an error. Because the step was never appended, `applyWrites` rolls back earlier files but cannot restore/remove the failed one. The test only forces `MkdirAll` to fail before the target is touched, so it proves the easy path and misses the transaction's actual invariant.

This is not atomic commit. Write complete contents to a sibling temporary file, close/sync it, then rename; journal the rename/backup so both the current write and earlier writes are recoverable. Preserve the original mode as well.

## 2. Force-mode preflight hides the human-owned files commit preserves

**BLOCKING.**

When `--force` changes the model, `stageOverlay` replaces every other `.go` file in the package with `package <pkg>`. That includes `hooks.go`, services, repositories, and arbitrary human-owned customization. Phase two therefore validates a fake package, while `resolveArtifact` deliberately keeps the real seed-once hooks.

Concrete failure: customize `hooks.go` to reference `row.Title`, then run `make resource Book name:string --force`. Preflight stubs the hook and succeeds; commit replaces the model/DTOs but preserves the real hook, leaving `row.Title` undefined. With Atlas absent, the command even returns success. You built a compile preflight and then removed the exact code that needed preflighting. `ಠ_ಠ`

Render the new artifacts first, then perform a second compile against an overlay of the complete final tree: new model + new generated files + every preserved human-owned file. Do not classify every non-model Go file as stale generated plumbing.

## 3. Dry-run and preflight can mutate go.mod/go.sum

**BLOCKING.**

`loadArtifacts` runs `go run -mod=mod` from the real application module. Go's `mod` mode writes required changes to `go.mod` and `go.sum`; this call happens during `PlanResource`, before `plan.Apply`, and it also runs for `--dry-run`. The new test explicitly runs `go mod tidy` first so it does not observe this mutation. That does not make the command read-only; it merely arranges the happy path. The explicit `-mod=mod` also rejects workspace mode where only `readonly` or `vendor` is accepted.

Use a non-mutating module mode for preflight (`-mod=readonly`, respecting vendor/workspace behavior), or run against a scratch module/overlay that includes module metadata. A dry-run that edits `go.sum` is not a dry-run, and a preflight mutation is outside the rollback transaction.

## 4. The migration guide promises unsupported custom-handler preservation

**BLOCKING.**

The guide says custom list filtering and extra endpoints can be kept in another hand-written file. The generated `Register` always binds its own unexported `list/get/create` methods and exposes no list hook or service injection point; a neighboring Go file is not automatically registered. Custom list behavior cannot replace the generated GET route, and an extra endpoint needs explicit registration wiring the guide never supplies. The new migration test uses an empty `type Handler struct{}` and therefore proves only file replacement, not migration of customized behavior.

Either document that resources with customization beyond `BeforeCreate` must remain on the legacy layout until a supported opt-out/extension point exists, or provide and test the registration/service boundary required to preserve that behavior. Do not tell users their code survives merely because the file survives.

CI is green. This is not a failing-test problem. The current tests do not exercise partial writes, force with customized hooks, an untidy/workspace module during dry-run, or an actually customized legacy handler.

# VERDICT

The deepest defect remains architectural: the patch calls a staged approximation a validated transaction, but the approximation excludes real committed code and performs unjournaled mutations. The comments and PR description promise stronger invariants than the runtime provides.

Do not merge. Make preflight observe the complete final tree, make module resolution read-only, implement real atomic file replacement, and narrow or complete the migration contract.

leo-aa88 (commented):
LGTM

PR #375, opened 2026-09-20
Convert tutorial example to model-first + drift-gate its generated files (#352)

### LINKED ISSUE #352: [RESGEN-1] Model-derived, generator-owned resource contracts (supersedes contract-drift in #218)

leo-aa88:
## Summary

Converts the tutorial example to the model-first layout and adds an in-process freshness check over its committed generated files. **This does not close #352** — see "Scope" below.

`examples/tutorial/internal/task` moves off the legacy hand-written `handler.go`/`routes.go` onto model-first:

- `task.go` — the human-owned model (unchanged shape) is the source of truth.
- `dto.gen.go` + `handler.gen.go` — **generator-owned** (`DO NOT EDIT`), now committed.
- `hooks.go` — human-owned `BeforeCreate`, seeded once.
- `admin.go`, `main.go` — unchanged (`main` still calls `task.Register`, now from `handler.gen.go`).

### The freshness check

`TestGeneratedFilesAreFresh` re-derives the generator-owned files from the model with the generator's derivation core (`resourcegen.RenderResource`) and byte-compares the committed `*.gen.go`, failing CI (via the existing **Test** job) if the model changes without regenerating. Regenerate with `go test ./examples/tutorial/internal/task/ -update`.

It is **not** `gombit generate --check`: it exercises only the derivation + byte-compare, not the CLI's `.gombit-resource` discovery, Program-Mode compilation, app-layout/legacy-handler validation, or `checkArtifacts`. The tutorial is part of the framework module, not a standalone app, so it can't host that CLI — this is an in-process substitute for a resource that lives here.

### Contract preserved

The existing HTTP-level `handler_test.go` is unchanged and **passes against the generated handler** — confirming model-first preserves the wire contract (pagination + clamping, `not_found`, `conflict`, `page`/`per_page` in OpenAPI). The model-first response DTO additionally exposes `gorm.Model`'s `created_at`/`updated_at` (derived from the model — the old hand-written handler omitted them); the README and `docs/tutorial.md` are updated to match.

## Scope

- Does **not** close #352. The named acceptance criterion — a per-PR `gombit generate --check` **CLI** gate over a real app layout (the goldentest demo, deferred as 5b-ii in #372) — remains open; CI still does not run the CLI.
- The other slice-6 items are already satisfied: `resourcecheck` and the old `*_drift_test.go` were never merged to `main` (PR #298 closed as a dead end), and the migration guide shipped with #372/#374.

Related #352.

## Validation

- `go build ./...`, `go vet`, `golangci-lint run` on the changed packages — clean (committed `*.gen.go` carry the standard generated-file header and are skipped by lint).
- `go test ./...` — pass (0 failures). `TestGeneratedFilesAreFresh` passes fresh and fails on a model change without `-update` (verified locally); `handler_test.go` passes against the generated handler.


examples/tutorial/internal/task/gen_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/gombit-dev/gombit/examples/tutorial/internal/task"
+	"github.com/gombit-dev/gombit/resourcegen"
+)
+
+var updateGen = flag.Bool("update", false, "rewrite the committed task *.gen.go from the model")
+
+// TestGeneratedFilesAreFresh is this example's per-PR `gombit generate --check`
```
leo-aa88: **[suggestion]** This is not `gombit generate --check`.

The test calls `resourcegen.RenderResource` in-process and byte-compares. That is the derivation core, and it is a real freshness check for these committed files. It is not the command ADR-016 / #352 named as the per-PR CI gate.

`gombit generate --check` also discovers via `.gombit-resource`, compiles models through Program Mode (`go run` in the app module), validates app layout, refuses a legacy `handler.go`, and goes through `checkArtifacts`. None of that runs here. `.github/workflows/ci.yml` still has no `generate --check` job. The `generate` package tests `--check` with a fake loader, not against a committed app.

THE COMMENT SAYS: this is the per-PR `gombit generate --check`.
THE RUNTIME DOES: `RenderResource(&task.Task{}, "task")`.

Do not close #352 on this PR. Either add a CI job that actually runs `gombit generate --check` against a real app layout (the goldentest demo #372 deferred as 5b-ii), or drop the AC-complete claim and describe this as an in-process substitute that cannot host the CLI.

**Suggestion:** Keep the byte-compare. Change the comment, README, and PR body so they do not equate this test with `gombit generate --check`, and leave the named CI job open.

leo-aa88 (commented):
## Summary

The tutorial conversion is real: `Register` moved into generator-owned `handler.gen.go`, hooks are seeded, the existing HTTP tests go through `task.Register`, and `TestGeneratedFilesAreFresh` does byte-compare `RenderResource` output against the committed `*.gen.go`. I ran that test locally; it passes. Lint/Build are green; Test was still in progress at review time.

The PR still should not close #352. ADR-016 and the issue name a per-PR `gombit generate --check` gate. This is an in-process `RenderResource` call. Those are not the same command, and CI still never runs the CLI.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

leo-aa88 (commented):
LGTM

